PR #11, opened 2021-09-05
Optionally require a deposit to make a proposal

ben2x4:
This PR closes https://github.com/JakeHartnell/cw-dao/issues/1. 

The dao can now be configured to require a deposit to make a proposal. The deposit is returned if the proposal passes and becomes property of the dao if the proposal fails. 

This change will require future work to prevent the dao from accidentally spending proposal deposit funds. This will most likely be implemented by having a separate contract for managing the deposit funds to make the distinction more clear. 

This PR also persists the proposer's address to the proposal state object. This was necessary to return the proposal deposit and seems generally useful.

### DIFF
--- a/src/contract.rs
+++ b/src/contract.rs
@@ -5,15 +5,15 @@ use crate::query::{
     VoteListResponse, VoteResponse, VoterResponse,
 };
 use crate::state::{
-    next_id, parse_id, Ballot, Config, Proposal, Votes, BALLOTS, CONFIG, PROPOSALS,
+    next_id, parse_id, Ballot, Config, Proposal, ProposalDeposit, Votes, BALLOTS, CONFIG, PROPOSALS,
 };
 use cosmwasm_std::{
     entry_point, to_binary, Addr, Binary, BlockInfo, CosmosMsg, Deps, DepsMut, Empty, Env,
-    MessageInfo, Order, Response, StdResult, Uint128,
+    MessageInfo, Order, Response, StdResult, Uint128, WasmMsg,
 };
 use cw0::{maybe_addr, Duration, Expiration};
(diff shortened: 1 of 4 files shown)


src/contract.rs
```diff
@@ -205,14 +254,37 @@ pub fn execute_execute(
...
+        .add_messages(refund_msg)
         .add_messages(prop.msgs)
         .add_attribute("action", "execute")
         .add_attribute("sender", info.sender)
         .add_attribute("proposal_id", proposal_id.to_string()))
 }
 
+fn get_proposal_deposit_refund_message(
+    proposer: &Addr,
+    config: &ProposalDeposit,
```
JakeHartnell, line 270: Making a comment for later, but feel this variable could be named a bit better... like why not `deposit` or `proposal_deposit`? Not a big deal though... good enough for now.

src/contract.rs
```diff
@@ -1378,16 +1485,130 @@ mod tests {
...
+                        amount: Uint128::zero(),
+                        token_address: Cw20Contract(cw20_addr),
+                    }
                 },
             }
         )
     }
+
+    #[test]
+    fn test_proposal_deposit_works() {
```
JakeHartnell, line 1505: Can be a separate ticket, but we should add a test to check the when the proposal deposit is 0, it all works without having to set an allowance. Will make a ticket.

JakeHartnell (approved):

src/contract.rs
ben2x4: All the existing tests have the proposal deposit set to zero, so we should be good from a coverage perspective. Thought could make sense to have a explicit test to document the requirement.

PR #21, opened 2021-09-14
Use BalanceAtHeight

### LINKED ISSUE #17: Update votes when determing if a proposal has passed to reflect current CW20 balances
This is a fairly large issue in the current implementation. After a user votes for a proposal, they can then transfer the cw20 to a different wallet and vote again. Effectively this gives anyone unlimited votes. 

One fix is to update vote power when checking if a proposal has passed. We could go further and just record the voter address and only get the token balance when evaluating if the proposal has passed. 

Another fix is to require voters to deposit their token when voting to be returned after the proposal resolves.

IMO, I think the first solution gives more flexibility to the token holders.

JakeHartnell:
So this uses the new cw20-gov token and the new `BalanceAtHeight` query. Similary to compound, when someone votes it checks their balance at the start of the proposal.

Closes #17.

Also added some more CI stuff and updated the schema.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,4 @@
+[alias]
+wasm = "build --release --target wasm32-unknown-unknown"
+wasm-debug = "build --target wasm32-unknown-unknown"
+unit-test = "test --lib"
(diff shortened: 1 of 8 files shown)


ben2x4 (approved):
Awesome how easy this was to implement in the DAO contract!

PR #30, opened 2021-09-27
All cw20 balances for treasury

JakeHartnell:


contracts/cw20-gov/src/contract.rs
```diff
@@ -80,8 +80,8 @@ fn verify_png_logo(logo: &[u8]) -> Result<(), ContractError> {
 /// Checks if passed logo is correct, and if not, returns an error
 fn verify_logo(logo: &Logo) -> Result<(), ContractError> {
     match logo {
-        Logo::Embedded(EmbeddedLogo::Svg(logo)) => verify_xml_logo(&logo),
-        Logo::Embedded(EmbeddedLogo::Png(logo)) => verify_png_logo(&logo),
+        Logo::Embedded(EmbeddedLogo::Svg(logo)) => verify_xml_logo(logo),
+        Logo::Embedded(EmbeddedLogo::Png(logo)) => verify_png_logo(logo),
```
ben2x4, line 84: Does this change need to be pushed upstream? Seems incorrect to need a reference of a reference.

contracts/cw-dao/src/state.rs
```diff
@@ -154,6 +154,7 @@ pub struct Ballot {
 // unique items
 pub const CONFIG: Item<Config> = Item::new("config");
 pub const PROPOSAL_COUNT: Item<u64> = Item::new("proposal_count");
+pub const TREASURY_TOKENS: Item<Vec<Addr>> = Item::new("treasury_tokens");
```
ben2x4, line 157: I think a `cw-storage-plus` Map with keys and empty data is going to be a much more efficient data structure. I believe, using Item with a vec will cause the entire vec to get rewritten every time it is modified. 

I'm still trying to wrap my head around the `cw-storage-plus` package but here is my current understanding. The primitive object is Path. An Item is just a nice wrapper around a Path. A map actually uses a new path for each key so modification of the map does not require the whole data structure to be rewritten. This would not work if our vec had duplicates, but in this case we actually have a Set of unique values.  It might be nice to contribute a Set type to cw-storage-plus to make this a little more ergonomic.

contracts/cw-dao/src/msg.rs
```diff
@@ -160,6 +165,13 @@ pub enum QueryMsg {
     Voter { address: String },
     /// Returns Config
     GetConfig {},
+    /// Returns All DAO Cw20 Balances
+    Cw20Balances {
+        start_after: Option<u64>,
```
ben2x4, line 170: I think logically these should be the same sized uint

contracts/cw-dao/src/contract.rs
```diff
@@ -352,6 +382,29 @@ pub fn execute_update_config(
...
 
+pub fn execute_update_cw20_token_list(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    mut to_add: Vec<Addr>,
+    to_remove: Vec<Addr>,
+) -> Result<Response<Empty>, ContractError> {
+    // Only contract can call this method
+    if env.contract.address != info.sender {
```
ben2x4, line 393: fine for now, but seems a little weird you have to use a proposal to do this 🤷‍♂️

contracts/cw-dao/src/contract.rs
```diff
@@ -352,6 +382,29 @@ pub fn execute_update_config(
...
+        return Err(ContractError::Unauthorized {});
+    }
+    TREASURY_TOKENS.update(deps.storage, |mut list| -> StdResult<_> {
```
ben2x4, line 396: This will become much easier with the Map type. It is a set so we can add and remove without worrying about duplicates.

ben2x4 (changes requested):
Good work here! I think my biggest concern is using the Map type to store the token list vector. I believe this will be much more efficient!

contracts/cw20-gov/src/contract.rs
JakeHartnell: This was clippy complaint. :shrug:

Clippy is :crab: :crown:.

ben2x4: the one true king

contracts/cw-dao/src/state.rs
JakeHartnell: Now we use maps. :rocket:

contracts/cw20-gov/src/contract.rs
JakeHartnell: Praise be. :crab: :crown:

contracts/cw-dao/src/contract.rs
JakeHartnell: Maybe a bit heavy weight, but didn't think these methods should just be open to everyone. :shrug:

JakeHartnell:
> Good work here! I think my biggest concern is using the Map type to store the token list vector. I believe this will be much more efficient!

It is nicer with Maps. :world_map:

PR #34, opened 2021-10-04
Refactor balance to voting power in prep for delegation

ben2x4:
This is a primer PR for adding delegation to the cw20-gov token. I converted the existing balances SnapshotMap to voting power and now leverage the cw20-base balance Map to track balances. This let me remove a lot of code and I wanted to separate this into a distinct PR for easier reveiwability. 

The next steps are to use the DELEGATIONS map to track who a token holder has delegated to. This info will be used to update voting power.


JakeHartnell (approved):
Thanks for submitting this separately. :pray: :pray: :pray:

Issue #28, opened 2021-09-25
Option to refund Proposal Deposits even on failing props


JakeHartnell:
Might be nice for some communities to not have a penalty for creating a failing proposal. We can add an option to the DAO config for this. If it's enabled, in `excute_close` we include a refund message.

JakeHartnell:
Closed with #35

PR #35, opened 2021-10-05
 Add refund option to dao contract #28

PotasticPotatoe:


contracts/cw-dao/src/contract.rs
```diff
@@ -1288,6 +1330,112 @@ mod tests {
...
+    #[test]
+    fn test_close_works_with_refund() {
+        let mut app = mock_app();
+        let voting_period = Duration::Height(2000000);
+        let threshold = Threshold::ThresholdQuorum {
+            threshold: Decimal::percent(51),
+            quorum: Decimal::percent(10),
+        };
+        let proposal_deposit_amount = Uint128::new(10);
+        let (dao_addr, _cw20_addr) = setup_test_case(
```
JakeHartnell, line 1342: If you're going to use this `_cw20_addr` variable like you do below, it shouldn't have the `_` in front of it. Should be `cw20_addr`.

contracts/cw-dao/src/contract.rs
```diff
@@ -1288,6 +1330,112 @@ mod tests {
...
+        assert!(res.is_ok());
+
+        let allowance = Cw20ExecuteMsg::IncreaseAllowance {
+            spender: dao_addr.clone().into(),
+            amount: proposal_deposit_amount,
+            expires: None,
+        };
+
+        let res = app.execute_contract(Addr::unchecked(OWNER), _cw20_addr.clone(), &allowance, &[]);
+        assert!(res.is_ok());
```
JakeHartnell, line 1410: Again, this shouldn't be needed... if it is, we need to fix this.

contracts/cw-dao/src/contract.rs
```diff
@@ -321,6 +331,7 @@ pub fn execute_update_config(
     max_voting_period: Duration,
     proposal_deposit_amount: Uint128,
     proposal_deposit_token_address: String,
+    refund_failed_proposals: Option<bool>,
```
JakeHartnell, line 334: Not your fault, but Clippy isn't going to let this pass because this function has too many arguments. We should make it take one argument, an `UpdateContfigMsg`.

Would look like so:
```rust
pub fn execute_update_config(
    deps: DepsMut,
    env: Env,
    info: MessageInfo,
    config: UpdateConfigMsg,
    ...
}
```

`UpdateConfigMsg`:
```rust
pub struct UpdateConfigMsg {
        threshold: Threshold,
        max_voting_period: Duration,
        proposal_deposit_amount: Uint128,
        proposal_deposit_token_address: String,
        refund_failed_proposals: Option<bool>,
}
```

contracts/cw-dao/src/contract.rs
```diff
@@ -1288,6 +1330,112 @@ mod tests {
...
+        let cw20 = Cw20Contract(_cw20_addr.clone());
+
+        let allowance = Cw20ExecuteMsg::IncreaseAllowance {
+            spender: dao_addr.clone().into(),
+            amount: proposal_deposit_amount,
+            expires: None,
+        };
+
+        let res = app.execute_contract(Addr::unchecked(OWNER), _cw20_addr.clone(), &allowance, &[]);
+        assert!(res.is_ok());
```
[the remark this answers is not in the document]
PotasticPotatoe: Initial DAO Balance was around 10 (could change to be higher).

My impression was that the user proposing gives the DAO money so that we can then create a proposal (with which the dao can use). However if that is wrong..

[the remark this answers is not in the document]
JakeHartnell: They need to send deposit funds along with the create proposal TX. CC @ben2x4.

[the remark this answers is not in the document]
JakeHartnell: > My impression was that the user proposing gives the DAO money so that we can then create a proposal (with which the dao can use). However if that is wrong..

My bad, this totally works. Guess I had using `send` on Cw20... but allowance works just as well. Silly me.

contracts/cw-dao/src/msg.rs
```diff
@@ -135,6 +137,7 @@ pub enum ExecuteMsg {
     },
 }
 
+
```
JakeHartnell, line 140:
```suggestion
```

PR #38, opened 2021-10-14
Update cw and plus versions

orkunkl:


ben2x4:
This looks great, thank you for taking care of it. 

Went ahead and fixed some test utility that was breaking. Seems like something changed with the updates

PR #44, opened 2021-10-17
Fix setting customer app.toml docker build script

ben2x4:
This change adds the custom app.toml script as the right time in the build in order to enable the REST api.


JakeHartnell (approved):
LFG

PR #48, opened 2021-10-17
Docker build uses user inputed address to fix Keplr issues

ben2x4:
Keplr has encoding issues when using the account 0 of a chain. This PR modifies the docker build script to instantiate a user inputted address as Account 1.

It also disables CORS for the RPC endpoint.

### DIFF
--- a/scripts/deploy_local.sh
+++ b/scripts/deploy_local.sh
@@ -89,7 +89,7 @@ CW_DAO_INIT="{
   \"max_voting_period\": {
     \"height\": 100
   },
-  \"proposal_deposit_amount\": \"1000000\",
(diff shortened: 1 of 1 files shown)


scripts/deploy_local.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+if [ "$1" = "" ]
```
elsehow, line 3: the idea here is that the user needs to pass in their own wasm address? if so, the message could be a bit more descriptive, and/or we should list something in the README to describe what the user should pass into this script.

JakeHartnell (approved):

Issue #50, opened 2021-10-17
Reduce entanglement with cw-dao-dapp


elsehow:
Currently, running a development environment (by our instructions) requires running the [frontend repo](https://github.com/DA0-DA0/cw-dao-dapp) in order to get a wasm address for one of your private keys. This links the two repos together a bit too tightly.

Ideally, developers should get a wasm address for their private key without the frontend repo.

JakeHartnell:
We can just provide a default account for them if they want. This will be less of an issue when we use a Dockerized Juno instance as they will just be able to use their Juno accounts.

JakeHartnell:
This is done, we have a default account.

PR #53, opened 2021-10-18
Make variable names in deploy script match those in `cw-dao-dapp` env.local

elsehow:
Removes ambiguity about which address to place in which .env.local field. 

(This issue is related to #50)


scripts/deploy_local.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
elsehow, line 1: this was causing really difficult-to-diagnose issues for me. it turns out that `sh` is aliased to `bash` on most systems (but not on mine, where it's actually `sh`), and the script was using some `bash` syntax here and there.

PR #148, opened 2022-01-30
Don't allow creation of claims if >= 100 are outstanding

ezekiiel:
There exists a world wherein a user of the staking contract creates so many claims that the process of iterating over them to redeem them causes an out of gas error.

Ideally, we'd sidestep this issue by adding the ability to redeem a single claim to the contracts. Sadly, cw-controllers has [no such](https://docs.rs/cw-controllers/latest/cw_controllers/struct.Claims.html) method and the fields of `Claims` are private. Short of submitting a change there we can limit the number of outstanding claims allowed.

Open to thoughts on if this is the approach we'd like to take. Ideal one is certainly to modify cw-controllers to allow this sort of behavior.

## Testing

Want to verify that 100 claims is a conservative enough threshold that one can still submit a claim that that point.

1. Spin up contracts in docker.
2. Import default account into docker container keyring.
2. Create a staking contract with a one second unstaking duration
3. Create 100 claims:
    ```
    for i in {0..100}; do docker exec -it cosmwasm junod tx wasm execute juno1j08452mqwadp8xu25kn9rleyl2gufgfjnv0sn8dvynynakkjukcqkfwuqa '{"unstake":{"amount":"1"}}' --from default --fees 20000ujunox --chain-id testing --gas auto --gas-adjustment 2 -y; sleep 6; done
    ```
4. Verify those claims have been created and are pending:
    ```
    docker exec -it cosmwasm junod query wasm contract-state smart juno1j08452mqwadp8xu25kn9rleyl2gufgfjnv0sn8dvynynakkjukcqkfwuqa '{"claims": {"address":"juno10j9gpw9t4jsz47qgnkvl5n3zlm2fz72k67rxsg"}}' --output json | jq '.data.claims | length'
    ```
5. Claim yo tokens:
    ```
    docker exec -it cosmwasm junod tx wasm execute juno1j08452mqwadp8xu25kn9rleyl2gufgfjnv0sn8dvynynakkjukcqkfwuqa '{"claim":{}}' --from default --fees 20000ujunox --chain-id testing --gas auto --gas-adjustment 2 -y
    ```

### DIFF
--- a/contracts/stake-cw20/src/contract.rs
+++ b/contracts/stake-cw20/src/contract.rs
@@ -13,7 +13,7 @@ use crate::msg::{
     StakedBalanceAtHeightResponse, StakedValueResponse, TotalStakedAtHeightResponse,
     TotalValueResponse,
 };
-use crate::state::{Config, BALANCE, CLAIMS, CONFIG, STAKED_BALANCES, STAKED_TOTAL};
+use crate::state::{Config, BALANCE, CLAIMS, CONFIG, MAX_CLAIMS, STAKED_BALANCES, STAKED_TOTAL};
 use crate::ContractError;
 use cw2::set_contract_version;
 pub use cw20_base::allowances::{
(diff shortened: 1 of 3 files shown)


JakeHartnell (approved):

Callum-A (approved):
LGTM!

PR #154, opened 2022-02-01
Add UpdateStakingContract method and tests

JakeHartnell:
Needed to provide a migration path for DAOs when new and better staking contracts inevitably come out.

### DIFF
--- a/contracts/cw3-dao/src/contract.rs
+++ b/contracts/cw3-dao/src/contract.rs
@@ -397,6 +397,7 @@ pub fn execute_update_staking_contract(
     if env.contract.address != info.sender {
         return Err(ContractError::Unauthorized {});
     }
+    let new_staking_contract = deps.api.addr_validate(new_staking_contract.as_str())?;
 
     // Replace the existing staking contract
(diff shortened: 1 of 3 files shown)


contracts/cw3-dao/src/contract.rs
```diff
@@ -384,6 +387,25 @@ pub fn execute_update_config(
...
+    info: MessageInfo,
+    new_staking_contract: Addr,
+) -> Result<Response<Empty>, ContractError> {
+    // Only contract can call this method
+    if env.contract.address != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // Replace the existing staking contract
+    STAKING_CONTRACT.save(deps.storage, &new_staking_contract)?;
```
ezekiiel, line 402: Let's validate the address here to prevent a DAO from getting locked up.

```suggestion
    let new_staking_contract = deps.api.addr_validate(new_staking_contract.as_str())?;

    // Replace the existing staking contract
    STAKING_CONTRACT.save(deps.storage, &new_staking_contract)?;
```

ezekiiel (approved):
❤️

PR #163, opened 2022-02-03
Make admin optional stake cw20

ben2x4:
This contract is more widely useful then just DAO's and therefore I think it is best if admin is optional. When an admin is not set the staking configs are immutable. This is an important user safeguard as the admin can lock the users funds forever and it could be quite dangerous if the admin was malicious.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,7 +283,7 @@ dependencies = [
 
 [[package]]
 name = "cw3-dao"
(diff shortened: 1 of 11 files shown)


ezekiiel (approved):
LGTM! Thanks :)

entrancedjames (approved):
Lgtm!

JakeHartnell (approved):

PR #164, opened 2022-02-03
[DRAFT] Stake cw20 rewards contract

ben2x4:
Basic contract for distributing staking rewards to stakers in `stake-cw20`. These rewards are autocompounding. To claim any user can claim the available rewards and it is a very cheap operation. These rewards are calculated block by block and sent to the `stake-cw20` contract. 

# TODO
- Tests
- Query pending rewards
- Optional admin: Refund? Extend reward duration?


JakeHartnell:
What's the status of this? Is this branch / PR still relevant?

PR #165, opened 2022-02-03
Add `start_height` field to proposal queries

ezekiiel:
The amount of voting power a given address has in a proposal is determined by their voting power on the block when the proposal as started. We currently do not expose this information via proposal queries which makes it quite the detective operation to determine if a given address can vote in a proposal.

### DIFF
--- a/contracts/cw3-dao/src/contract.rs
+++ b/contracts/cw3-dao/src/contract.rs
@@ -542,6 +542,7 @@ fn query_proposal(deps: Deps, env: Env, id: u64) -> StdResult<ProposalResponse> 
         expires: prop.expires,
         threshold,
         deposit_amount: prop.deposit,
+        start_height: prop.start_height,
     })
 }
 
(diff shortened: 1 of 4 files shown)


ben2x4 (commented):
LGTM, though would love someone a little more familiar with this contract to review.

ben2x4 (approved):

contracts/cw3-dao/src/contract.rs
```diff
@@ -542,6 +542,7 @@ fn query_proposal(deps: Deps, env: Env, id: u64) -> StdResult<ProposalResponse>
         expires: prop.expires,
         threshold,
         deposit_amount: prop.deposit,
+        start_height: prop.start_height,
```
maurolacy, line 545: So, this field was already there, and you are just exposing it.

maurolacy (approved):
LGTM.

contracts/cw3-dao/src/contract.rs
ezekiiel: Yep :)

Issue #147, opened 2022-01-30
Allow only signers or DAO members to execute transactions


JakeHartnell:
https://twitter.com/poolpitako/status/1487486538825060352?s=20&t=yDwEhQi75PaHN6EHX-8OPw

> can it be so that only contract signers can execute by default, to not get ^^ situations or MEV extraction

Inspired by @ValarDragon. :pray: 

Think this makes a lot of sense... right now anyone can execute a transaction after it's been passed. This can allow for some weird situations. Think a better default would be to have only signers or DAO members be able to execute transactions.

Could make it a setting potentially...

Relevant article: https://blog.gnosis.pm/gnosis-safe-mev-how-to-mitigate-it-347e13535e34

verminator23:
@JakeHartnell Great idea. I can tackle this

verminator23:
@JakeHartnell Is it sufficient to just check if the sender has a staked balance?

ezekiiel:
> Is it sufficient to just check if the sender has a staked balance?

Yup :) or in the case of a multisig you can check if they have voting power.

JakeHartnell:
Closed with: https://github.com/DA0-DA0/dao-contracts/pull/166

PR #166, opened 2022-02-04
Vt only signers execute

verminator23:
Only allow signers / members to execute proposals in cw3-dao and cw3-multisig. Addresses issue #147


JakeHartnell:
This is great. I think it would be good to have this behind a config option.

Sometimes it's nice to let anyone execute. For example, a DAO could build an auto-execute bot that would automatically execute proposals upon passing.

Think it depends on the DAO / community... maybe @ValarDragon has thoughts on this?

Not sure what the default should be. CC @elsehow / @ezekiiel / @bmorphism,

ezekiiel:
I like the execute bot idea. Am pro putting it behind a config option. Can add a template for this once it lands :)

verminator23:
Great idea. Will add a setting to the config

verminator23:
Made this a config option @ezekiiel @JakeHartnell. I set the default to be true, but lmk if you guys prefer the opposite

contracts/cw3-dao/src/contract.rs
```diff
@@ -59,6 +59,7 @@ pub fn instantiate(
         proposal_deposit: msg.proposal_deposit_amount,
         refund_failed_proposals: msg.refund_failed_proposals,
         image_url: msg.image_url,
+        only_members_execute: true,
```
JakeHartnell, line 62: Can we update the instantiate msg to include this option?

verminator23:
> One last comment, and then I think this is good to go.

Added it

JakeHartnell (approved):
LGTM!

PR #170, opened 2022-02-09
add stake-cw20-external-rewards

ben2x4:
This contract distributes external staking rewards to stakers in the `stake-cw20` contract. External means the rewards is either a native or cw20 that is not the same as the cw20 being staked. 

The contract calculates rewards based on blocks but only distributes them at predefined epochs. This is all configurable in the instatiation message.

### DIFF
--- a/contracts/stake-cw20-external-rewards/LICENSE
+++ b/contracts/stake-cw20-external-rewards/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+   4. Redistribution. You may reproduce and distribute copies of the
+      Work or Derivative Works thereof in any medium, with or without
+      modifications, and in Source or Object form, provided that You
+      meet the following conditions:
+
+      (a) You must give any other recipients of the Work or
+          Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices
+          stating that You changed the files; and
+
+      (c) You must retain, in the Source form of any Derivative Works
+          that You distribute, all copyright, patent, trademark, and
+          attribution notices from the Source form of the Work,
+          excluding those notices that do not pertain to any part of
+          the Derivative Works; and
+
+      (d) If the Work includes a "NOTICE" text file as part of its
+          distribution, then any Derivative Works that You distribute must
+          include a readable copy of the attribution notices contained
+          within such NOTICE file, excluding those notices that do not
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an addendum to the NOTICE text from the Work, provided
+          that such additional attribution notices cannot be construed
+          as modifying the License.
+
+      You may add Your own copyright statement to Your modifications and
+      may provide additional or different license terms and conditions
+      for use, reproduction, or distribution of Your modifications, or
+      for any such Derivative Works as a whole, provided Your use,
+      reproduction, and distribution of the Work otherwise complies with
+      the conditions stated in this License.
+
+   5. Submission of Contributions. Unless You explicitly state otherwise,
+      any Contribution intentionally submitted for inclusion in the Work
+      by You to the Licensor shall be under the terms and conditions of
+      this License, without any additional terms or conditions.
+      Notwithstanding the above, nothing herein shall supersede or modify
+      the terms of any separate license agreement you may have executed
+      with Licensor regarding such Contributions.
+
+   6. Trademarks. This License does not grant permission to use the trade
+      names, trademarks, service marks, or product names of the Licensor,
+      except as required for reasonable and customary use in describing the
+      origin of the Work and reproducing the content of the NOTICE file.
+
+   7. Disclaimer of Warranty. Unless required by applicable law or
+      agreed to in writing, Licensor provides the Work (and each
+      Contributor provides its Contributions) on an "AS IS" BASIS,
+      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+      implied, including, without limitation, any warranties or conditions
+      of TITLE, NON-INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A
+      PARTICULAR PURPOSE. You are solely responsible for determining the
+      appropriateness of using or redistributing the Work and assume any
+      risks associated with Your exercise of permissions under this License.
+
+   8. Limitation of Liability. In no event and under no legal theory,
+      whether in tort (including negligence), contract, or otherwise,
+      unless required by applicable law (such as deliberate and grossly
+      negligent acts) or agreed to in writing, shall any Contributor be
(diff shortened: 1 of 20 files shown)


contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+    };
+    CONFIG.save(deps.storage, &config)?;
+    Ok(Response::default())
+}
+
+pub fn validate_instantiate_msg(env: &Env, msg: &InstantiateMsg) -> Result<(), ContractError> {
+    if env.block.height > msg.start_block {
+        return Err(ContractError::StartBlockAlreadyOccurred {});
+    }
+    if msg.start_block > msg.end_block {
```
ezekiiel, line 46: Should this be `>=`? Currently could instantiate a contract that starts and ends at the same block and it would be valid so long as that block was a multiple of `payment_block_delta`.

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+        return Err(ContractError::StartBlockAfterEndBlock {});
+    }
+    if msg.start_block % msg.payment_block_delta != 0 {
+        return Err(ContractError::StartBlockNotDivisibleByPaymentDelta {});
+    }
+    if msg.end_block % msg.payment_block_delta != 0 {
+        return Err(ContractError::EndBlockNotDivisibleByPaymentDelta {});
+    }
```
ezekiiel, line 54: Seems like this is what we actually want here. Slightly less restrictive.

(Note: invented a new error type here. Probably don't actually accept the suggestion.)

```suggestion
    if (msg.end_block - msg.start_block) % msg.payment_block_delta != 0 {
        return Err(ContractError::RewardsDurationNotDivisibleByPaymentDelta {});
    }
```

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+    let duration = Uint128::from(((msg.end_block - msg.start_block) / msg.payment_block_delta) + 1);
+    let calculated_total = duration
+        .checked_mul(msg.payment_per_block)
+        .map_err(cosmwasm_std::StdError::overflow)?
+        .checked_mul(Uint128::from(msg.payment_block_delta))
+        .map_err(cosmwasm_std::StdError::overflow)?;
```
ezekiiel, line 60: Should we really multiply by `payment_block_delta` here? Seems like `calculated_total = payment_per_block * duration` is what we want.

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+    config: &Config,
+    info: &MessageInfo,
+    amount: Uint128,
+) -> StdResult<CosmosMsg> {
+    Ok(match config.clone().denom {
+        Denom::Native(denom) => BankMsg::Send {
+            to_address: info.sender.to_string(),
+            amount: vec![Coin { amount, denom }],
+        }
+        .into(),
```
ezekiiel, line 202:
```suggestion
    Ok(match &config.denom {
        Denom::Native(denom) => BankMsg::Send {
            to_address: info.sender.to_string(),
            amount: vec![Coin { amount, denom: denom.to_string() }],
        }
        .into(),
```

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+pub fn query_info(deps: Deps) -> StdResult<InfoResponse> {
+    let config = CONFIG.load(deps.storage)?;
+    Ok(InfoResponse {
+        start_block: config.start_block,
+        end_block: config.end_block,
+        payment_per_block: config.payment_per_block,
+        total_amount: config.total_amount,
+        denom: config.denom,
+        staking_contract: config.staking_contract.to_string(),
+        payment_block_delta: config.payment_block_delta,
```
ezekiiel, line 264: Might be nice to include if the contract has been funded in here.

ezekiiel (commented):
Looking good :) Just couple minor comments

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+pub fn execute(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
+        ExecuteMsg::Claim {} => try_claim(deps, _env.clone(), info, _env.block.height),
+        ExecuteMsg::Receive(msg) => try_receive(deps, _env, info, msg),
+        ExecuteMsg::ClaimUpToBlock { block } => try_claim(deps, _env, info, block),
```
ezekiiel, line 77: Cool to have this. The while loop with queries inside seems like a potential out of gas issue but this should provide a way out.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yeah, this would would result in one payment. Probable not the best practice and a seperate contract should be written for these cases, but nothing invalid about this.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yes, good suggestion

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+    up_to_block: u64,
+) -> StdResult<(Uint128, u64)> {
+    let delta = config.payment_block_delta;
+    let mut current_block = last_claimed;
+    let mut amount = Uint128::zero();
+    while current_block <= min(up_to_block, config.end_block) {
+        let stake_info = get_stake_info_at_height(deps.querier, config, address, current_block)?;
+        if stake_info.balance > Uint128::zero() {
+            amount += ((Uint128::from(delta) * config.payment_per_block) * stake_info.balance)
+                .checked_div(stake_info.total)?
```
ezekiiel, line 184: This div will round down so unless things divide pretty nicely this will end up distributing slightly (on order of `10 ** -6 juno`) less rewards than it has been given I believe.

Probably fine but I wonder if you have thoughts about that. I've got the same issue with the DAO Up! contracts.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: ah, I see. I was kinda a dum dum here. I need to not divide by payment_block_delta when caculating duration, then I do not need to multiply by it later.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yeah emergency escape procedure should it ever occur

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yeah, i think rounding down is acceptable. Not sure there is any other option. I guess you could add a refund method for the dust at the end but idk if it really matters for most sane tokens. I could see this being an issue for tokens with really low supply and no decimals for example :man_shrugging:

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: good catch

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn instantiate(
+    deps: DepsMut,
+    env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    validate_instantiate_msg(&env, &msg)?;
+    let config = Config {
+        start_block: msg.start_block,
```
verminator23, line 26: What's the rationale for having the user pass in start_block, end_block, and payment per block? It would be more elegant to accept a duration and an amount, and then calculate payment per unit time within the contract. I do something similar [here](https://github.com/CosmWasm/cw-tokens/blob/4ca9307f10b94f545427963c78bf10a0ef9c7641/contracts/cw20-streams/src/contract.rs#L61).

contracts/stake-cw20-external-rewards/src/contract.rs
verminator23, line 77: Not sure why ClaimUpToBlock is necessary. Probably should just claim up to the latest block every time

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    let mut config = CONFIG.load(deps.storage)?;
+    if config.funded {
+        return Err(ContractError::AlreadyFunded {});
+    }
+    if config.denom != Denom::Cw20(info.sender) {
+        return Err(ContractError::IncorrectDenom {});
+    }
+    if amount != config.total_amount {
```
verminator23, line 106: Maybe change to amount >= config.total_amount, and refund any extra funds that were sent

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+    deps: Deps,
+    address: &Addr,
+    config: &Config,
+    last_claimed: u64,
+    up_to_block: u64,
+) -> StdResult<(Uint128, u64)> {
+    let delta = config.payment_block_delta;
+    let mut current_block = last_claimed;
+    let mut amount = Uint128::zero();
+    while current_block <= min(up_to_block, config.end_block) {
```
verminator23, line 180: Oh I see. ClaimUpToBlock can be used to prevent out of gas issue here

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1324 @@
...
+}
+
+pub fn try_fund(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    let mut config = CONFIG.load(deps.storage)?;
+    if config.funded {
```
verminator23, line 100: Shouldn't you be able to top up this contract whenever the rewards get depleted. Otherwise, you'd have to instantiate a new contract every time you want to allocate more rewards

verminator23:
This is a solid start. There are a few major issues though:
1) The contract has to loop over every unclaimed block height to redeem all rewards for an account. So stakers would potentially have to submit several transactions to claim all of their rewards (otherwise their transactions will run out of gas). This UX is not ideal IMO. This is primarily due to the limitations of querying the stake cw20 contract, which indexes stakes based on block height. If you want to accommodate more traditional staking mechanisms (like Synthetix or Sushiswap staking rewards), which provide better UX, you may want to consider a different staking contract implementation.
2) The contract can't be topped up, so you can't allocate more funds to issue additional rewards. You'd have to deploy an entirely new rewards contract to do so.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: I like that the parameters are explicit and the instatiator is required to double check everything will function as expected.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: This is a mechanism to protect against a rare edge case where the users has not claimed their rewards to a very long time and claim all results in the operation hitting the per block gas limit.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: I really like require the instatiator to be precise with their inputs. Failing the instatiate requires them to revaluate the inputs and ensure they are doing what they intend to do.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yeah this is a good additional feature

ben2x4:
> This is a solid start. There are a few major issues though:
> 
> 1. The contract has to loop over every unclaimed block height to redeem all rewards for an account. So stakers would potentially have to submit several transactions to claim all of their rewards (otherwise their transactions will run out of gas). This UX is not ideal IMO. This is primarily due to the limitations of querying the stake cw20 contract, which indexes staking info based on block height. If you want to accommodate more traditional staking mechanisms ([similar to Synthetix staking rewards](https://github.com/Synthetixio/synthetix/blob/develop/contracts/StakingRewards.sol)), which provide better UX, you may want to consider a different staking contract implementation.
> 2. The contract can't be topped up, so you can't allocate more funds to issue additional rewards. You'd have to deploy an entirely new rewards contract to do so.

I agree that point 2 is a great feature to implement before production deployment. 

In regards to point 1, need to look a little more deeply into how they implement this. Would love to chat.

JakeHartnell:
Can we please rebase this branch to make it easier to merge in PRs potentially? :pray:

contracts/stake-cw20-external-rewards/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
JakeHartnell, line 190: Update this to be either your name or DAO DAO please. : )

contracts/stake-cw20-external-rewards/README.md
```diff
@@ -0,0 +1,108 @@
+# CosmWasm Starter Pack
```
JakeHartnell, line 1: The most basic README would be nice. Can remove all of this template stuff.

contracts/stake-cw20-external-rewards/src/contract.rs
JakeHartnell: Personally, agree with @ben2x4 on this one.

contracts/stake-cw20-external-rewards/src/contract.rs
JakeHartnell: Is there a test case for this? If not we should make a TODO / issue. Can be added in a separate PR.

contracts/stake-cw20-external-rewards/src/contract.rs
JakeHartnell: This is the least of our worries IMO. Dust is just a part of crypto...

contracts/stake-cw20-external-rewards/src/contract.rs
JakeHartnell: Me like.

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1328 @@
...
+            block_height: config.start_block,
+            time: Timestamp::default(),
+        });
+    let (amount, _) = get_amount_owed(
+        deps,
+        &address,
+        &config,
+        last_claimed.block_height,
+        env.block.height,
+    )?;
```
JakeHartnell, line 283: This could get expensive?

donkey-donkey:
have you all checked out the repo from cw20 - reward. its just like anchor on terra and yield protocol on ethereum
https://github.com/CosmWasm/cw-tokens/tree/cw20-reward/contracts/cw20-reward

verminator23:
> have you all checked out the repo from cw20 - reward. its just like anchor on terra and yield protocol on ethereum https://github.com/CosmWasm/cw-tokens/tree/cw20-reward/contracts/cw20-reward

Ben also needs it to support multiple reward tokens for a single staking contract. But I agree, this is a good start.

ben2x4:
This approach is flawed, implementing this with a different method

PR #171, opened 2022-02-10
Add address management smart contract for managing a favorites list

ezekiiel:
Small contract for managing a sorted list of addresses. Might be used
for keeping a list of favorite contracts.

Addresses are added in the form:

```rust
struct AddressItem {
	pub addr: Addr,
	pub priority: u32,
}
```

One may then query for the contract list and addresses will be
returned in sorted order with the highest priority ones returned
first.

Placing many, countless addresses into this contract may cause it to
become unusable as it will run into out of gas issues. The contract
will only become locked at the point when the number of addresses
becomes so enormous that a single `log(N)` operation over a B-Tree set
excedes the gas limit.


contracts/address-manager/src/contract.rs
```diff
@@ -0,0 +1,109 @@
...
+    ADMIN.save(deps.storage, &admin)?;
+    Ok(Response::default()
+        .add_attribute("method", "update_admin")
+        .add_attribute("new_admin", admin))
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::GetAddresses {} => query_get_addresses(deps),
```
JakeHartnell, line 89: We may want pagination for this?

contracts/address-manager/src/error.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("Unauthorized")]
+    Unauthorized {},
+    // Add any other custom errors you like here.
+    // Look at https://docs.rs/thiserror/1.0.21/thiserror/ for details.
```
JakeHartnell, line 12:
```suggestion
```

contracts/address-manager/rustfmt.toml
```diff
@@ -0,0 +1,15 @@
+# stable
```
JakeHartnell, line 1: Do we need files like this if they exist in the root of this repo?

contracts/address-manager/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
JakeHartnell, line 190: Update please with "DAO DAO" or yourself... @elsehow do we want a license with each contract?

contracts/address-manager/.gitignore
```diff
@@ -0,0 +1,15 @@
+# Build results
```
JakeHartnell, line 1: Do we need files like this if they exist in the root of this repo?

contracts/address-manager/.github/workflows/Basic.yml
```diff
@@ -5,34 +5,33 @@ on: [push, pull_request]
 name: Basic
 
 jobs:
+
```
JakeHartnell, line 8: Feel like we only need one file like this in the repo root.

contracts/address-manager/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
JakeHartnell, line 1: Do we need files like this if they exist in the root of this repo?

contracts/address-manager/.gitignore
ezekiiel: Good point. Probably not because this is a workspace and build output ends up in the root. Removing..

contracts/address-manager/LICENSE
ezekiiel: Looks like we don't actually [need this](https://infra.apache.org/licensing-howto.html). Having the LICENCE file in the root seems like it is good enough

contracts/address-manager/src/contract.rs
ezekiiel: Good idea. Added optional start and end bounds to this query.

contracts/address-manager/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "address-manager"
```
JakeHartnell, line 2: Prefix with `cw-`?

contracts/address-manager/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+[package.metadata.scripts]
+optimize = """docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  cosmwasm/rust-optimizer:0.12.4
+"""
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0-beta4" }
+cosmwasm-storage = { version = "1.0.0-beta4" }
```
JakeHartnell, line 33: You can leave off the `4` here and it will use the latest version. For example: `"1.0.0-beta"`.

contracts/address-manager/src/contract.rs
```diff
@@ -0,0 +1,117 @@
...
+    ADMIN.save(deps.storage, &admin)?;
+    Ok(Response::default()
+        .add_attribute("method", "update_admin")
+        .add_attribute("new_admin", admin))
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::GetAddresses {start, end} => query_get_addresses(deps, start, end),
```
JakeHartnell, line 85: Maybe a query to check if a single address is in the list?

JakeHartnell: `QueryMsg::CheckAddress { addr }`?

JakeHartnell (commented):
This is great! Just a few comments, and then I think this is ready. : )

ezekiiel:
Renamed to cw-address-manager and added a `CheckAddress` query.  Also tested using the DAO Up! DAO today. See most recent proposals [here](https://testnet.daodao.zone/dao/juno192c76yeywn7xe84ddw9jjtstecnztqhf4cg7cex07gnyxavchulqv7zp3d).

JakeHartnell (approved):
LGTM!

JakeHartnell:
Closing in favor of the `cw4-group` fork approach as we discussed @ezekiiel.

PR #173, opened 2022-02-11
Stake CW20 Rewards

Callum-A:
Picked up the TODOs left on PR #164.

- Added a query pending rewards route.
- Pulled in Main as the current stake-cw20-rewards was using an older version of the stake-cw20 contract. (That's why other commits are listed under this PR)
- Added integration tests
  - Checks initial config setup
  - Checks funding
  - Checks rewards
  - Checks compounding each block
  - Checks invalid claim rewards calls
 
Feedback appreciated.

### DIFF
--- a/contracts/stake-cw20-rewards/src/contract.rs
+++ b/contracts/stake-cw20-rewards/src/contract.rs
@@ -424,15 +424,15 @@ mod tests {
     fn test_query_config() {
         let mut app = mock_app();
         let total_payment = Uint128::from(TOTAL_PAYMENT);
-        let start_block = (&mut app).block_info().height + 2; // We need to add 2 due to setup contract advancing the block twice
-        let end_block = (&mut app).block_info().height + 2 + TOTAL_REWARDS_DURATION;
+        let start_block = app.block_info().height + 2; // We need to add 2 due to setup contract advancing the block twice
+        let end_block = app.block_info().height + 2 + TOTAL_REWARDS_DURATION;
         let (rewards, stake, token) =
             setup_test_case(&mut app, vec![], total_payment, start_block, end_block);
-        let resp = query_config(&app, rewards.clone());
+        let resp = query_config(&app, rewards);
         assert_eq!(
             GetConfigResponse {
-                token_address: token.clone(),
-                staking_contract: stake.clone(),
+                token_address: token,
+                staking_contract: stake,
                 payment_per_block: total_payment
                     .checked_div(Uint128::from(end_block - start_block))
                     .unwrap_or_default(),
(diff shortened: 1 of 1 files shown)


JakeHartnell:
Upstream PR needs a rebase, then it will be easier to review this. CC @ben2x4. :pray:

contracts/stake-cw20-rewards/src/contract.rs
```diff
@@ -161,5 +163,486 @@ pub fn query_config(deps: Deps) -> StdResult<GetConfigResponse> {
     })
 }
 
+pub fn query_rewards(deps: Deps, env: Env) -> StdResult<GetPendingRewardsResponse> {
```
ezekiiel, line 166: What do you think about renaming this to `query_pending_rewards`?

contracts/stake-cw20-rewards/src/contract.rs
```diff
@@ -161,5 +163,486 @@ pub fn query_config(deps: Deps) -> StdResult<GetConfigResponse> {
...
+    let config = CONFIG.load(deps.storage)?;
+    let last_claim = LAST_CLAIM.load(deps.storage).unwrap_or(config.start_block);
+    let blocks = Uint128::from(min(&env.block.height, &config.end_block) - last_claim);
+    let pending_rewards = if config.funded {
+        blocks
+            .checked_mul(config.payment_per_block)
+            .map_err(StdError::overflow)?
+    } else {
+        Uint128::zero()
+    };
```
ezekiiel, line 176: Does this contract have a `payment_block_delta` like the other one? If so I don't think this is quite accurate. If not, we should discuss if they should be consistent. https://github.com/DA0-DA0/dao-contracts/pull/170#discussion_r802210734

ezekiiel (commented):
A rebase on main would be amazing for reviewing this 🙏

contracts/stake-cw20-rewards/src/contract.rs
Callum-A: This contract doesn't have a delta like external, as rewards are paid every block so I believe the maths here is correct in line with @ben2x4 's description in PR #164 but correct me if I'm misunderstanding.

contracts/stake-cw20-rewards/src/contract.rs
Callum-A: I agree this needs to be renamed.

JakeHartnell:
@ezekiiel / @Callum-A, is this PR still relevant? It's been open for a while.

JakeHartnell:
Yolo.

PR #174, opened 2022-02-12
Remove cw-distribution

Callum-A:
- Deletes cw-distribution files
- Removes cw-distribution reference from root Cargo.toml
- Removes cw-distribution reference from root README.md


JakeHartnell (approved):
Thanks!

PR #179, opened 2022-02-13
[hold] Add cw4-group contract which returns members in sorted order.

ezekiiel:
This is an implementation of the [cw4 spec](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw4/README.md). Unlike the [cw4-group](https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw4-group)
contract in cw-plus this contract returns queries for the list of
members in order of highest weights first.

As it needs to sort the members by weight in order to get the list of
members the member list query is less gas efficent than the cw-plus
version. Happily, queriers will hit limits on the amount of text that
is possible to return from a query before they approach the gas limits
for a query.

Testing on the uni-2 testnet suggests that this cap to the number of
addresses one can add to the contract is somewhere around `5991`.
At this point queries can still be made, they just can't list all of the 
members.

This was tested using the following shell script:

```bash
CONTRACT_ADDR=juno1rsdnksm93kqtausqf48k0zjez42lzq9988zsr6u38c6un5v02g4qyxmstd

function query_members() {
    junod query wasm contract-state smart $CONTRACT_ADDR '{"list_members":{}}' --output json | jq '.data.members | length'
}

function generate_address() {
    local body=$(cat /dev/urandom | env LC_ALL=C tr -dc 'a-zA-Z0-9' | fold -w 59 | head -n 1)
    echo juno$body
}

function generate_members() {
    count=$1
    for i in $(seq $count); do
	addr=$(generate_address)
	weight=$RANDOM
	echo -n "{\"addr\":\"$addr\",\"weight\":$weight} "
    done
}

function generate_member_list() {
    count=$1
    echo -n "["
    extra_comma=$(generate_members $count | tr ' ' ,)
    echo -n ${extra_comma%?}
    echo -n "]"
}

function add_members() {
    members_unescaped=$(generate_member_list 500)
    members=$(echo $members_unescaped | tr '"' '\"')
    junod tx wasm execute $CONTRACT_ADDR "{\"update_members\":{\"remove\":[],\"add\":$members}}" --from ekez --fees 50000ujunox --chain-id uni-2 --gas auto --gas-adjustment 5 -y
}

while true; do
    old=$(query_members)
    current=$(query_members)
    add_members
    while [ "$old" == "$current" ]; do
	current=$(query_members)
	sleep 10
    done
    echo "-> $current"
done
```


ezekiiel:
I have moved all of the cw-plus for this contract here and added additional tests for queries returning members in sorted order.

orkunkl:
I have not started reviewing, but just one thought: maybe `weight` multi-index could solve this?

contracts/cw4-group-sorted/src/msg.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    /// Change the admin
+    UpdateAdmin { admin: Option<String> },
+    /// Applys a diff to the existing members. Remove is applied
+    /// after add, so if an address is in both, it is removed
+    UpdateMembers {
+        remove: Vec<String>,
+        add: Vec<Member>,
```
entrancedjames, line 25: are there any benefits of coupling these two? (suppose in the future we have a call to this execute method for, it would be weird to pass an empty set to the add and only add to the remove).

ezekiiel: We're trying to be compatible with the [cw4 spec](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw4/README.md) which requires that these are implemented. Think an advantage of this approach is that gas fees have a high base cost and then scale more slowly with an increase in required compute. Doing 'more' in one operation is liable to be more gas efficient than two smaller operations.

ezekiiel:
> @orkunkl: I have not started reviewing, but just one thought: maybe weight multi-index could solve this?

I'm having trouble understanding the docs for that type. Is there any chance you know of an example of using that I might refer to?

fwiw - I think it's probably impossible to have both constant time insert and remove and iterate over slices of the keys in sorted order. Sorting will always have a `nlog(n)` in it.

JakeHartnell (approved):

contracts/cw4-group-sorted/src/contract.rs
```diff
@@ -0,0 +1,189 @@
...
+use crate::error::ContractError;
+use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
+use crate::query::{Member, MemberListResponse};
+use crate::state::{
+    initialize_members, list_members_sorted, remove_member, update_member, ADMIN, HOOKS, MEMBERS,
+    TOTAL,
+};
+
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:string-manager";
```
Callum-A, line 18: Maybe a bit pedantic but do we want to call it string-manager over something like cw4-group-sorted?

janitachalam (approved):

contracts/cw4-group-sorted/src/contract.rs
JakeHartnell:
```suggestion
const CONTRACT_NAME: &str = "crates.io:cw4-group-sorted";
```

maurolacy:
> I have not started reviewing, but just one thought: maybe `weight` multi-index could solve this?



> > @orkunkl: I have not started reviewing, but just one thought: maybe weight multi-index could solve this?

Agreed.

> 
> I'm having trouble understanding the docs for that type.

It's tricky to use multi-indexes, but worth it. Perhaps you can ask for clarification of the docs in the original repo?

> Is there any chance you know of an example of using that I might refer to?

See the [`ListMembersByPoints`](https://github.com/confio/poe-contracts/blob/c733eb265c78bd0a20e1cc172d61047ddfeaa17a/packages/tg4/src/query.rs#L18) query impl, in `poe-contracts` contracts like `tg4-stake`.

What we did was extending the `cw4` spec to support "by points" queries. That is the [`tg4`](https://github.com/confio/poe-contracts/tree/main/packages/tg4) spec. I would recommend doing something similar. No need to depend on `poe-contracts`, but it's better IMO to have a new spec (`cw4-plus`, `cw5`, or so), than creating a specific contract for this functionality alone. 

> 
> fwiw - I think it's probably impossible to have both constant time insert and remove and iterate over slices of the keys in sorted order. Sorting will always have a `nlog(n)` in it.

ezekiiel:
Closing for now. Don't think we need this for the new DAO contracts.

PR #180, opened 2022-02-17
Hooks in staking contract

ben2x4:
This PR adds the ability to add hooks to the staking contract, which are contracts that will be notified when users stake and unstake funds. This will allow cool features to be built on top of the staking contract such as delegated voting and external staking rewards.

I have also taken made some improvements to the governance of modifying hooks as it is extremly high risk as an improper hook contract would lock all funds until the hook is removed. To better enable this, I have split the admin role into an owner and manager. The owner can add/remove hooks, change the owner, change the admin, and change the unbonding duration. The manager can do all of this excepct change the owner. This allows the DAO to delegate responsibilities of managing hooks to a more reponsive party such as a multisig while maintaining authority to remove the manager in case they attempt to lock funds. 

This is especially important in the wasmswap context as a staking contract would need to be instatiated and managed for each pool. The amount of management will most likely exceed the capacity of the whole DAO to make each individual contract change. 

TODO:

Need to write actual tests for the hooks, but this requires a dummy contract to ensure hooks are functioning as intended.


contracts/stake-cw20/src/state.rs
```diff
@@ -3,12 +3,14 @@ use serde::{Deserialize, Serialize};
...
 use cw_controllers::Claims;
+use cw_controllers::Hooks;
 use cw_storage_plus::{Item, SnapshotItem, SnapshotMap, Strategy};
 use cw_utils::Duration;
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct Config {
-    pub admin: Option<Addr>,
+    pub owner: Option<Addr>,
+    pub manager: Option<Addr>,
```
entrancedjames, line 13: should we add comments describing the differences in behavior for these two roles (and their use cases).

entrancedjames (approved):
Overall LGTM, tho I'm new to the hooks system and have some questions.

contracts/stake-cw20/src/state.rs
orkunkl: Agree

JakeHartnell: +1

contracts/stake-cw20/src/contract.rs
```diff
@@ -66,42 +75,52 @@ pub fn execute(
...
-                "admin",
-                config
-                    .admin
-                    .map(|a| a.to_string())
-                    .unwrap_or_else(|| "None".to_string()),
-            ))
-        }
-    }
+    config.owner = new_owner;
+    config.manager = new_manager;
```
ezekiiel, line 104: Probably want to validate these addresses.

contracts/stake-cw20/src/contract.rs
```diff
@@ -278,6 +302,38 @@ pub fn execute_fund(
...
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    addr: Addr,
+) -> Result<Response, ContractError> {
+    let config: Config = CONFIG.load(deps.storage)?;
+    if config.owner != Some(info.sender.clone()) && config.manager != Some(info.sender) {
+        return Err(ContractError::Unauthorized {});
+    };
+    HOOKS.add_hook(deps.storage, addr.clone())?;
```
ezekiiel, line 315: Does `HOOKS` validate this address?

ezekiiel: We should also be careful here because a valid address must be a contract address.

contracts/stake-cw20/src/contract.rs
```diff
@@ -278,6 +302,38 @@ pub fn execute_fund(
...
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    addr: Addr,
+) -> Result<Response, ContractError> {
+    let config: Config = CONFIG.load(deps.storage)?;
+    if config.owner != Some(info.sender.clone()) && config.manager != Some(info.sender) {
+        return Err(ContractError::Unauthorized {});
+    };
+    HOOKS.remove_hook(deps.storage, addr.clone())?;
```
ezekiiel, line 331: Validate address?

ezekiiel (commented):
This looks great. Just a couple questions around address validation. Seems like a bad hook will not stop removal of said hook which is nice.

ben2x4:
> Just double-checking, but all this contract is doing is applying a side-effect on specific behaviors on this contract to other contracts. Is there any expectation on downstream contracts? For example what if they fail on downstream side-effects?

The contract is notifying via hooks other contracts when a user modifies their amount staked. There is a critical expectations that these contracts will not fail. It is critical that whoever is governoring these hooks is not malicious as they have the ability to lock all staked funds by adding an invalid hook. Hooks can be removed if a contracts is broken.

contracts/stake-cw20/src/state.rs
ben2x4: :heavy_check_mark:

contracts/stake-cw20/src/contract.rs
ben2x4: is the addr type validated when it is parsed from Json?

orkunkl:
> > Just double-checking, but all this contract is doing is applying a side-effect on specific behaviors on this contract to other contracts. Is there any expectation on downstream contracts? For example what if they fail on downstream side-effects?
> 
> The contract is notifying via hooks other contracts when a user modifies their amount staked. There is a critical expectations that these contracts will not fail. It is critical that whoever is governoring these hooks is not malicious as they have the ability to lock all staked funds by adding an invalid hook. Hooks can be removed if a contracts is broken.

`SubMsg::new` sets the reply method as `Never` as default. Even if the hook contract execution fails, txs will not so super safe it to use here.

contracts/stake-cw20/src/contract.rs
ezekiiel: Nope :/ https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Addr.html

contracts/stake-cw20/src/contract.rs
ezekiiel: Nope :/ https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Addr.html

JakeHartnell (approved):

ben2x4:
> > > Just double-checking, but all this contract is doing is applying a side-effect on specific behaviors on this contract to other contracts. Is there any expectation on downstream contracts? For example what if they fail on downstream side-effects?
> > 
> > 
> > The contract is notifying via hooks other contracts when a user modifies their amount staked. There is a critical expectations that these contracts will not fail. It is critical that whoever is governoring these hooks is not malicious as they have the ability to lock all staked funds by adding an invalid hook. Hooks can be removed if a contracts is broken.
> 
> `SubMsg::new` sets the reply method as `Never` as default. Even if the hook contract execution fails, txs will not so super safe it to use here.

oh interesting, I think this actually presents its own problems? Like if a sub-contract gets into a weird state and is failing, it might not be super obvious. Weird things might then occur becuase the hooked contract becomes out of sync. Probable a super edge case, but definitely some new concerns to think through here.

verminator23:
> oh interesting, I think this actually presents its own problems? Like if a sub-contract gets into a weird state and is failing, it might not be super obvious. Weird things might then occur becuase the hooked contract becomes out of sync. Probable a super edge case, but definitely some new concerns to think through here.

@ben2x4 Yeah I agree. I think having the tx fail if the contract receiving the hook fails might actually be more ideal. If there's a critical bug in the dependent contract, you'd want to block execution until a fix is applied. And resume normal operations afterwards. Would prevent the dependent contract from reaching an invalid state.

verminator23:
Overall LGTM. The address validation checks that @ezekiiel mentioned would be nice. Also a few tests with a dummy contract that receives the hooks would be good

contracts/stake-cw20/src/contract.rs
ben2x4: cool added validation

ben2x4: ty for looking into it

contracts/stake-cw20/src/contract.rs
ben2x4: not technically needed here as validated when added, but added anyways for best practices

contracts/stake-cw20/src/contract.rs
ben2x4: validated

contracts/stake-cw20/src/contract.rs
```diff
@@ -39,13 +42,19 @@ pub fn instantiate(
...
     msg: InstantiateMsg,
 ) -> Result<Response<Empty>, ContractError> {
-    let admin = match msg.admin {
+    let owner = match msg.owner {
+        Some(admin) => Some(deps.api.addr_validate(admin.as_str())?),
+        None => None,
+    };
+
+    let manager = match msg.manager {
         Some(admin) => Some(deps.api.addr_validate(admin.as_str())?),
```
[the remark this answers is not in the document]
ben2x4: good catch

contracts/stake-cw20/src/contract.rs
```diff
@@ -39,13 +42,19 @@ pub fn instantiate(
     _info: MessageInfo,
     msg: InstantiateMsg,
 ) -> Result<Response<Empty>, ContractError> {
-    let admin = match msg.admin {
+    let owner = match msg.owner {
+        Some(admin) => Some(deps.api.addr_validate(admin.as_str())?),
```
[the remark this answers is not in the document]
ben2x4: ty

ben2x4:
Changed all msg addr types to strings

contracts/stake-cw20/src/contract.rs
```diff
@@ -67,42 +75,61 @@ pub fn execute(
...
     duration: Option<Duration>,
 ) -> Result<Response, ContractError> {
+    let new_owner = match new_owner {
+        Some(a) => Some(deps.api.addr_validate(&*a)?),
+        None => None,
+    };
+    let new_manager = match new_manager {
+        Some(a) => Some(deps.api.addr_validate(&*a)?),
+        None => None,
+    };
```
ezekiiel, line 102:
```suggestion
    let new_owner = new_owner.map(|new_owner| deps.api.addr_validate(new_owner)).transpose()?;
    let new_manager = new_manager.map(|new_manager| deps.api.addr_validate(new_manager)).transpose()?;
```

contracts/stake-cw20/src/contract.rs
```diff
@@ -279,6 +311,40 @@ pub fn execute_fund(
         .add_attribute("amount", amount))
 }
 
+pub fn execute_add_hook(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    addr: String,
+) -> Result<Response, ContractError> {
+    let addr = deps.api.addr_validate(&addr)?;
```
ezekiiel, line 320: Would be SO nice if we could work out a way to validate that the address here is a contract address and not just a wallet address. Unclear to me if CosmWasm supports a way to do this though as none of their code does so far as I can tell.

ezekiiel (approved):
Happy to merge this now. Before we deploy need some tests but there is a dependency issue so I'm OK with waiting on that.

contracts/stake-cw20/src/contract.rs
maurolacy: We have that around in some places. Basically, you do a raw query for `ContractInfo`. It's a little tricky to handle the error / success conditions well, though.

contracts/stake-cw20/src/contract.rs
```diff
@@ -67,42 +75,61 @@ pub fn execute(
...
-    match config.admin {
-        None => Err(ContractError::NoAdminConfigured {}),
-        Some(current_admin) => {
-            if info.sender != current_admin {
-                return Err(ContractError::Unauthorized {
-                    expected: current_admin,
-                    received: info.sender,
-                });
-            }
+    if Some(info.sender.clone()) != config.owner && Some(info.sender.clone()) != config.manager {
```
orkunkl, line 104: Using pattern matching would make this code a lot simpler

verminator23 (approved):

contracts/stake-cw20/src/contract.rs
ben2x4: would be nice, but would rather not have messy validation code and think existing validation is good enough.

contracts/stake-cw20/src/contract.rs
ben2x4: I am not sure pattern matching makes it that much simpler, happy to take a look at a suggested change.

Issue #169, opened 2022-02-07
Reject proposals when they can't pass


JakeHartnell:
Currently, `cw3-dao` and `cw3-multisig` only show proposals as rejected when time has run out. This is because they really only pay attention to the `yes` votes.

From the `README`:
> Before the proposal has expired, any voter with non-zero weight can add their vote. Only "Yes" votes are tallied. If enough "Yes" votes were submitted before the proposal expiration date, the status is set to "Passed".

We should also tally "No" votes, and update the proposal status if it reaches a sufficient amount to prevent the proposal from passing.

JakeHartnell:
Closed with #181

PR #181, opened 2022-02-17
Reject proposals when they cannot pass

Callum-A:
Addresses Issue #169 

Adds an is_rejected method to both Proposal types which does the same as is_passed however checks no votes for a majority.
Added unit tests  for the logic on the Proposal side.
Added integration tests for both a DAO and Multisig.

Feedback appreciated.

### DIFF
--- a/contracts/cw3-multisig/src/state.rs
+++ b/contracts/cw3-multisig/src/state.rs
@@ -94,6 +94,9 @@ impl Proposal {
         if status == Status::Open && self.expires.is_expired(block) {
             status = Status::Rejected;
         }
+        if status == Status::Open && self.is_rejected(block) {
+            status = Status::Rejected;
+        }
 
         status
     }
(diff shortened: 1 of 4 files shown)


contracts/cw3-multisig/src/state.rs
```diff
@@ -107,17 +107,17 @@ impl Proposal {
         self.status = self.current_status(block);
     }
 
-    // returns true iff this proposal is sure to pass (even before expiration if no future
-    // sequence of possible votes can cause it to fail)
-    pub fn is_passed(&self, block: &BlockInfo) -> bool {
+    /// Helper function to check if a certain vote count has reached threshold.
+    /// Only called from is_rejected and is_passed for no and yes votes
+    fn does_vote_count_reach_threshold(&self, vote_count: u64, block: &BlockInfo) -> bool {
```
maurolacy, line 112: Good generalisation. A little unclear what this does, but I think with the comment above, it's OK.

maurolacy (commented):
Looks good. ~~It would be good to have unit tests for checking and documenting the "reject early" case.~~ (already done)

Will backport this to cw-plus at some point.

maurolacy (approved):
Here are the unit tests. It seems I was reviewing an outdated version before.

maurolacy:
Returning to the upstream version of `cw3-flex-multisig` could be a good idea at some point.

Need to confirm / check this, but in principle you could be contributing / adding these there directly there.

Callum-A:
> Returning to the upstream version of `cw3-flex-multisig` could be a good idea at some point.
> 
> Need to confirm / check this, but at least in principle you could be contributing / adding these there directly there.

I can definitely look into this upstream

contracts/cw3-multisig/src/state.rs
Callum-A: I'm going to look into making it clearer via making the comments better today.

Callum-A: Have attempted to improve the comments in my latest commits, feedback welcome.

ezekiiel (approved):
Looks great. Thanks for all the tests!

JakeHartnell (approved):

JakeHartnell:
Awesome work!

PR #182, opened 2022-02-19
External staking rewards

ben2x4:
WIP in progress but wanted to get some feed back if possible cc @verminator23


contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    env: Env,
+    info: MessageInfo
+) -> Result<Response<Empty>, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if config.admin != Some(info.sender.clone()){
+        return Err(Unauthorized {})
+    };
+    let coin = info.funds.first()?;
+    let amount = coin.clone().amount;
+    let denom = coin.clone().denom;
```
ezekiiel, line 99:
```suggestion
    let { amount, denom } = coin;
```

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+pub fn execute_fund (
+    mut deps: DepsMut,
+    env: Env,
+    sender: Addr,
+    amount: Uint128
+) -> Result<Response<Empty>, ContractError> {
+    update_rewards(&mut deps, &env, &sender);
+    
+    let reward_config = REWARD_CONFIG.load(deps.storage)?;
+    let new_reward_config = if reward_config.periodFinish < env.block.height {
```
ezekiiel, line 115: If I execute fund a second time before the initial rewards finish is the intended behavior to overwrite that previous reward rate?

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    Ok(Response::new().add_message(transfer_msg).add_attribute("action", "claim").add_attribute("amount", rewards))
+}
+
+pub fn get_transfer_msg(recipient: Addr, amount: Uint128, denom: Denom) -> StdResult<CosmosMsg> {
+    match denom {
+        Denom::Native(denom) => {
+            Ok(
+               BankMsg::Send {
+                   to_address: recipient.into_string(),
+                   amount: vec![Coin{ denom: denom, amount }]
```
ezekiiel, line 189:
```suggestion
                   amount: vec![Coin{ denom, amount }]
```

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    let reward_config = REWARD_CONFIG.load(deps.storage)?;
+    let total_staked = get_total_staked(deps.as_ref(), &config.staking_contract)?;
+    let staked_balance = get_staked_balance(deps.as_ref(), &config.staking_contract, addr)?;
+    let current_block = min(env.block.height, reward_config.periodFinish);
+    let last_update_block = LAST_UPDATE_BLOCK.load(deps.storage).unwrap_or_default();
+
+    let reward_per_token = REWARD_PER_TOKEN.update::<_, StdError>(deps.storage, |r| {
+        let additional_reward_per_token = match total_staked {
+            Uint128::zero() => Uint128::zero(),
+            _ => ((reward_config.rewardRate * max(Uint128::from(current_block - last_update_block), Uint128::zero()))
```
ezekiiel, line 218: Is this `max` call needed? A uint should never be smaller than zero.

ezekiiel (commented):
Just an initial pass. I'll take another more detailed look soon. Unsure about the logic around multiple fund messages being executed. Could well be correct.

Previously the idea was to have two staking rewards contracts. This seems to handle both cw20 tokens and native tokens. Should we close open PRs for those other contracts?

ben2x4:
> Just an initial pass. I'll take another more detailed look soon. Unsure about the logic around multiple fund messages being executed. Could well be correct.
> 
> Previously the idea was to have two staking rewards contracts. This seems to handle both cw20 tokens and native tokens. Should we close open PRs for those other contracts?

Two contract is right, but different reason. One contract is for external rewards, ie rewards that are not the same as the staking token, and one contract is for auto compounding rewards that are the same token.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: is that what happens here? intended behavior is to extend the current rewards period with the same rate

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: is that how overflow works? will it just stop at 0?

contracts/stake-cw20-external-rewards/NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright 2021 Ben <ben2x4@tutanota.com>
```
Callum-A, line 1: Do we need a NOTICE? I notice (no pun intended) none of our other contracts have them

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+    CONFIG.save(deps.storage, &config)?;
+
+    let reward_config = RewardConfig {
+        period_finish: 0,
+        reward_rate: Default::default(),
+        reward_duration: 100000,
+    };
+    REWARD_CONFIG.save(deps.storage, &reward_config)?;
+
+    Ok(Response::new())
```
Callum-A, line 60: Maybe worth adding some attributes here?

contracts/stake-cw20-external-rewards/src/state.rs
```diff
@@ -0,0 +1,32 @@
...
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{Addr, Uint128, Uint256};
+use cw20::Denom;
+
+use cw_storage_plus::{Item, Map};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct Config {
+    pub owner: Option<Addr>,
```
Callum-A, line 11: Is there a reason for both an owner and manager? Are there certain actions an owner can do and a manager cannot?

Callum-A (dismissed):
All looks good, just a couple queries and niceties

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+    addr: &Addr,
+    reward_per_token: Uint256,
+    staking_contract: &Addr,
+) -> StdResult<Uint128> {
+    let _config = CONFIG.load(deps.storage)?;
+    let staked_balance = get_staked_balance(deps, staking_contract, addr)?;
+    let user_reward_per_token = USER_REWARD_PER_TOKEN
+        .load(deps.storage, addr.clone())
+        .unwrap_or_default();
+    let reward_factor = reward_per_token.checked_sub(user_reward_per_token)?;
```
ezekiiel, line 293: 1. I stake some tokens at a reward per token rate of 2.
2. The reward period for those ends.
3. The rewards are updated with a reward per token of 1.
4. I attempt to claim my rewards and this overflows.

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+        let result = numerator.checked_div(denominator)?;
+
+        // try_into() shouldn't cause problems here because final results fits into Uint256
+        result
+            .checked_div(Uint512::from(scale_factor()))?
+            .try_into()
+            .unwrap()
+    };
+
+    Ok(prev_reward_per_token + additional_reward_per_token)
```
ezekiiel, line 278: If I am hitting this line because I have added new funds to the contract after the previous funding set has expired should the reward rate really be the new one + the old one? Seems to me that in that case the old one shouldn't matter because it has already been paid out.

ezekiiel: In other words, this line seems to imply to me that the rewards rate will only ever increase. The ability to fund a rewards contract after it has completed a round of rewards seems to imply to me that is not the case as a new funding round should set a new `rewards_per_token` rate separate from the one that completed.

contracts/stake-cw20-external-rewards/src/contract.rs
ezekiiel: Note: this only applies about the `reward_per_token` value not being monotonically increasing in the other comment isn't off base.

contracts/stake-cw20-external-rewards/src/state.rs
ben2x4: yes, main difference is owner can change both the owner and manager address. Manager can not alter the owner. The idea here is a DAO might want to delegate the management of the contract to an operations multisig. The owner/manager model allows this delegation will still leaving the dao the ability to reassert control if the manager ever misbehaves.

Callum-A: Awesome, I noticed these are documented under the InstantiateMsg, so I'm happy with how it is.

contracts/stake-cw20-external-rewards/src/contract.rs
ezekiiel: This line below seems like it isn't taking into account the old rate and is just setting the rate based on the amount sent:
```rust
rewardRate: amount / reward_config.rewardDuration
```

contracts/stake-cw20-external-rewards/src/contract.rs
ezekiiel: Oh I see. Are you worried here about an overflow when you do `current_block - last_update_block`? I guess that is possible if the rewards time is expired and someone has caused `LAST_UPDATE_BLOCK` to be updated.

You probably just want something like:

```rust
if current_block < last_update_block {
  Uint128::zero()
} {
  Uint128::new(current_block - last_update_block)
}
```

Rusts overflow behavior is to wrap. So something like `0u32 - 1` would wrap around to `2^32`. Taking the max here won't help as you'll just get the overflowed value. [This](https://huonw.github.io/blog/2016/04/myths-and-legends-about-integer-overflow-in-rust/) blog post has some nice info about that.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: ok, yeah think I see what you are saying here. Maybe need to look at the tests as well, will get back to you.

ben2x4: ok, here is what is happening, needed to refresh context a bit.

1. If rewards are ended, ie `reward_period_finish` < `current_block_height`, we will start a new rewards period with `amount` distributed evenly over `rewards_duration` number of blocks.

2. If rewards are ongoing we will increase the remaining rewards by `amount` evenly distributed over the remaining blocks in the reward period.

I think confusion here is `reward_config.periodFinish < env.block.height == true` means rewards are currently not ongoing so the reward amount should be overwritten.

ben2x4: This is how the synthetix reward contract functions https://github.com/Synthetixio/synthetix/blob/45fd1febc8595057ea8d157730b36111c775d177/contracts/StakingRewards.sol#L113

Interested if you think a different approach is better suited.

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: ah lol, good catch, will implement better overflow handling

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: hmm, i think this suggestion is invalid. not sure proper syntax here to do the destructuring

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: this is resolved, as last update block will only get updated to the end of the current rewards period

contracts/stake-cw20-external-rewards/NOTICE
ben2x4: always have included in my contracts :shrug:

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yup makes sense

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response<Empty>, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let owner = match msg.owner {
+        Some(owner) => Some(deps.api.addr_validate(owner.as_str())?),
+        None => None,
+    };
```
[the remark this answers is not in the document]
ben2x4: ok resolved this, needed to convert all addr types to strings in contract messages

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+        owner,
+        manager,
+        staking_contract: msg.staking_contract,
+        reward_token: msg.reward_token,
+    };
+    CONFIG.save(deps.storage, &config)?;
+
+    let reward_config = RewardConfig {
+        period_finish: 0,
+        reward_rate: Default::default(),
```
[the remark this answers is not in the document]
ben2x4: :+1:

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+    let manager = match msg.manager {
+        Some(manager) => Some(deps.api.addr_validate(manager.as_str())?),
+        None => None,
+    };
+
+    let config = Config {
+        owner,
+        manager,
+        staking_contract: msg.staking_contract,
+        reward_token: msg.reward_token,
```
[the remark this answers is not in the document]
ben2x4: reward token is a denom type so used a match statment to validate when it is a cw20.

contracts/stake-cw20-external-rewards/src/contract.rs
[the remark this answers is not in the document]
ben2x4: done

contracts/stake-cw20-external-rewards/src/msg.rs
```diff
@@ -0,0 +1,55 @@
...
+use stake_cw20::hooks::StakeChangedHookMsg;
+
+use crate::state::{Config, RewardConfig};
+pub use cw_controllers::ClaimsResponse;
+
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, PartialEq)]
+pub struct InstantiateMsg {
+    pub owner: Option<Addr>,
+    pub manager: Option<Addr>,
+    pub staking_contract: Addr,
```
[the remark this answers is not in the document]
ben2x4: done

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+
+pub fn execute_receive(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    wrapper: Cw20ReceiveMsg,
+) -> Result<Response<Empty>, ContractError> {
+    let msg: ReceiveMsg = from_binary(&wrapper.msg)?;
+    let config = CONFIG.load(deps.storage)?;
+    let sender = deps.api.addr_validate(&*wrapper.sender)?;
```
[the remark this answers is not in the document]
ben2x4: no, good catch

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+    if config.reward_token != Denom::Cw20(info.sender) {
+        return Err(InvalidCw20 {});
+    };
+    match msg {
+        ReceiveMsg::Fund { .. } => execute_fund(deps, env, sender, wrapper.amount),
```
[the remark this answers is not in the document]
ben2x4: good catch

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+    env: Env,
+    info: MessageInfo,
+) -> Result<Response<Empty>, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    let coin = info
+        .funds
+        .clone()
+        .pop()
+        .ok_or(ContractError::InvalidFunds {})?;
```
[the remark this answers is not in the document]
ben2x4: good suggestion

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+}
+
+pub fn execute_fund(
+    mut deps: DepsMut,
+    env: Env,
+    sender: Addr,
+    amount: Uint128,
+) -> Result<Response<Empty>, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if config.owner != Some(sender.clone()) && config.manager != Some(sender.clone()) {
```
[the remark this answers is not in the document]
ben2x4: the some requires it

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+                + (amount / Uint128::from(reward_config.period_finish - env.block.height)),
+            reward_duration: reward_config.reward_duration,
+        }
+    };
+
+    REWARD_CONFIG.save(deps.storage, &new_reward_config)?;
+
+    Ok(Response::new()
+        .add_attribute("action", "fund")
+        .add_attribute("amount", amount))
```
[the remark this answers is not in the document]
ben2x4: :+1:

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: yes hit is true, it is a montonically increasing value

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: it is monotonically increasing

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,1994 @@
...
+        RewardConfig {
+            period_finish: env.block.height + reward_config.reward_duration,
+            reward_rate: amount / Uint128::from(reward_config.reward_duration),
+            reward_duration: reward_config.reward_duration,
+        }
+    } else {
+        RewardConfig {
+            period_finish: reward_config.period_finish, // period finish needs to be incremented by the rewards duration
+            reward_rate: reward_config.reward_rate
+                + (amount / Uint128::from(reward_config.period_finish - env.block.height)),
```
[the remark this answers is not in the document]
ben2x4: equal sign leads to path without the divid by zero concern.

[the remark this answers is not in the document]
ben2x4: I think current condition is correct

contracts/stake-cw20-external-rewards/src/msg.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::state::{Config, RewardConfig};
+pub use cw_controllers::ClaimsResponse;
+
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, PartialEq)]
+pub struct InstantiateMsg {
+    pub owner: Option<String>,
+    pub manager: Option<String>,
+    pub staking_contract: String,
+    pub reward_token: Denom,
+}
```
ezekiiel, line 16: If we're instantiating contracts like this from a frontend it would be really nice to be able to specify rewards duration here.

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -0,0 +1,2057 @@
...
+
+    let reward_token = match msg.reward_token {
+        Denom::Native(denom) => Denom::Native(denom),
+        Cw20(addr) => Cw20(deps.api.addr_validate(&addr.to_string())?),
+    };
+
+    let config = Config {
+        owner,
+        manager,
+        staking_contract: deps.api.addr_validate(&msg.staking_contract)?,
```
ezekiiel, line 51: Could we run a `GetConfig {}` query here to smoke check that the address supplied is actually a staking contract?

ezekiiel (approved):
Great work! Two little comments. LG2M.

JakeHartnell:
@ben2x4, let's clean up the unused variable linting errors and merge. :rocket:

JakeHartnell (approved):

contracts/stake-cw20-external-rewards/src/msg.rs
ben2x4: done, good logic

contracts/stake-cw20-external-rewards/src/contract.rs
ben2x4: good call out, used total staked query as it is more specific to the staking contract.

ezekiiel (approved):

JakeHartnell (approved):

PR #186, opened 2022-03-03
Remove usage of Addr in contract messages.

ezekiiel:
The Addr type is not validated during deserialization yet it still
implements both serialization and deserialization. This means that an
invalid Addr can be constructed in the event that it is passed to a
contract via a message.

This removes all usage of Addr in message types. Before this change
two bugs are present:

1. For the DAO and Multisig contracts tokens can be added to the
treasury which are invalid.
2. For the staking contract this allows setting the admin of the
contract to an invalid address.

### DIFF
--- a/contracts/cw3-dao/src/contract.rs
+++ b/contracts/cw3-dao/src/contract.rs
@@ -143,9 +143,9 @@ pub fn instantiate(
                 admin: Some(env.contract.address.to_string()),
                 label,
                 msg: to_binary(&stake_cw20::msg::InstantiateMsg {
-                    admin: Some(env.contract.address),
+                    admin: Some(env.contract.address.to_string()),
                     unstaking_duration,
-                    token_address: cw20_addr.addr(),
+                    token_address: cw20_addr.addr().to_string(),
                 })?,
             };
 
@@ -454,13 +454,13 @@ pub fn execute_update_staking_contract(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    new_staking_contract: Addr,
+    new_staking_contract: String,
 ) -> Result<Response<Empty>, ContractError> {
     // Only contract can call this method
     if env.contract.address != info.sender {
         return Err(ContractError::Unauthorized {});
     }
-    let new_staking_contract = deps.api.addr_validate(new_staking_contract.as_str())?;
+    let new_staking_contract = deps.api.addr_validate(&new_staking_contract)?;
 
     // Replace the existing staking contract
     STAKING_CONTRACT.save(deps.storage, &new_staking_contract)?;
@@ -474,8 +474,8 @@ pub fn execute_update_cw20_token_list(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    to_add: Vec<Addr>,
-    to_remove: Vec<Addr>,
+    to_add: Vec<String>,
+    to_remove: Vec<String>,
 ) -> Result<Response<Empty>, ContractError> {
     // Only contract can call this method
     if env.contract.address != info.sender {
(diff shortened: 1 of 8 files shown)


JakeHartnell (approved):

Callum-A (approved):
LGTM!

contracts/cw3-dao/src/contract.rs
```diff
@@ -143,9 +143,9 @@ pub fn instantiate(
                 admin: Some(env.contract.address.to_string()),
                 label,
                 msg: to_binary(&stake_cw20::msg::InstantiateMsg {
-                    admin: Some(env.contract.address),
+                    admin: Some(env.contract.address.to_string()),
```
webmaster128, line 146: `to_string()` always creates a clone. You can use `env.contract.address.into()` instead, which is a free conversion from Addr to String.

contracts/cw3-dao/src/contract.rs
```diff
@@ -454,13 +454,13 @@ pub fn execute_update_staking_contract(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    new_staking_contract: Addr,
+    new_staking_contract: String,
```
webmaster128, line 457: Perfect. You get the `Addr` type safely below.

contracts/cw3-dao/src/contract.rs
```diff
@@ -771,9 +773,9 @@ pub fn reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractEr
                         admin: Some(env.contract.address.to_string()),
                         label: env.contract.address.to_string(),
                         msg: to_binary(&stake_cw20::msg::InstantiateMsg {
-                            admin: Some(env.contract.address),
+                            admin: Some(env.contract.address.to_string()),
                             unstaking_duration,
-                            token_address: cw20_addr,
+                            token_address: cw20_addr.to_string(),
```
webmaster128, line 778: `cw20_addr.into()` is again a free alternative that avoids the clone.

contracts/cw3-dao/src/msg.rs
```diff
@@ -163,12 +163,12 @@ pub enum ExecuteMsg {
     UpdateConfig(Config),
     /// Updates token list
     UpdateCw20TokenList {
-        to_add: Vec<Addr>,
-        to_remove: Vec<Addr>,
+        to_add: Vec<String>,
+        to_remove: Vec<String>,
```
webmaster128, line 167: 👌

contracts/cw3-dao/src/msg.rs
```diff
@@ -163,12 +163,12 @@ pub enum ExecuteMsg {
...
     },
     /// Update Staking Contract (can only be called by DAO contract)
     /// WARNING: this changes the contract controlling voting
-    UpdateStakingContract { new_staking_contract: Addr },
+    UpdateStakingContract { new_staking_contract: String },
```
webmaster128, line 171: 👍

contracts/cw3-dao/src/tests.rs
```diff
@@ -2032,7 +2026,7 @@ fn test_update_staking_contract() {
 
     // Nobody can call call update staking contract method directly
     let update_staking_contract_msg = ExecuteMsg::UpdateStakingContract {
-        new_staking_contract: Addr::unchecked("Better_Staking_Contract"),
+        new_staking_contract: "Better_Staking_Contract".to_string(),
```
webmaster128, line 2029: Heads up: upper case addresses become invalid in the latest MockApi implementation. But you can use "better_staking_contract" when you run into trouble.

contracts/stake-cw20/src/contract.rs
```diff
@@ -89,7 +90,7 @@ pub fn execute_update_config(
                 });
             }
 
-            config.admin = new_admin;
+            config.admin = new_admin.map(|a| deps.api.addr_validate(&a)).transpose()?;
```
webmaster128, line 93: Nice one

webmaster128 (commented):
> This means that an invalid Addr can be constructed in the event that it is passed to a contract via a message.

Right, the `Addr` type must not be used for untrusted data, such as incoming messages. Deserialization is too primitive to be able to validate an address against a chain specific address format.

PR #189, opened 2022-03-06
Instantiate DAO with existing staking contract

Callum-A:
We now have an added enum in instantiate to create a DAO with an existing staking contract.

- Similar to how we handle CW20s.
- If we instantiate a new CW20 the contract will error if the user specifies to use an existing staking contract. As this action is invalid
- Existing CW20s can use both a new staking contract or an older one
- Test cases cover all possible combinations

Feedback appreciated.

### DIFF
--- a/contracts/cw3-dao/src/contract.rs
+++ b/contracts/cw3-dao/src/contract.rs
@@ -3,7 +3,9 @@ use crate::helpers::{
     get_and_check_limit, get_deposit_message, get_proposal_deposit_refund_message,
     get_staked_balance, get_total_staked_supply, get_voting_power_at_height, map_proposal,
 };
-use crate::msg::{ExecuteMsg, GovTokenMsg, InstantiateMsg, ProposeMsg, QueryMsg, VoteMsg};
+use crate::msg::{
+    ExecuteMsg, GovTokenMsg, InstantiateMsg, ProposeMsg, QueryMsg, StakingContractMsg, VoteMsg,
+};
 use crate::query::{
     ConfigResponse, Cw20BalancesResponse, ProposalListResponse, ProposalResponse,
     ThresholdResponse, TokenListResponse, VoteInfo, VoteListResponse, VoteResponse,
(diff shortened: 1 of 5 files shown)


contracts/cw3-dao/src/contract.rs
```diff
@@ -136,22 +132,33 @@ pub fn instantiate(
...
+                        })?,
+                    };
+                    let msg = SubMsg::reply_on_success(msg, INSTANTIATE_STAKING_CONTRACT_REPLY_ID);
 
-            msgs.append(&mut vec![msg]);
+                    msgs.append(&mut vec![msg]);
+                }
+                StakingContractMsg::UseExistingStakingContract { addr } => {
+                    // Validate the address
+                    let staking_contract_addr = deps.api.addr_validate(&addr)?;
```
JakeHartnell, line 157: We might want to do a bit more validation here... for example:
- Does this contract's staking token match the cw20 the DAO is configured with?
- Is it a staking contract?

Callum-A: This was a concern of mine, what would be the best way to achieve this? Send a query for a known route and fail over safely if it is not present?

Callum-A: Could query using the GetConfig route, gives us the token address to then validate.

contracts/cw3-dao/src/contract.rs
```diff
@@ -136,22 +132,43 @@ pub fn instantiate(
...
 
-            msgs.append(&mut vec![msg]);
+                    let resp: stake_cw20::msg::GetConfigResponse = deps.querier.query_wasm_smart(
+                        &staking_contract_addr,
+                        &stake_cw20::msg::QueryMsg::GetConfig {},
+                    )?;
+
+                    if cw20_addr.addr() != resp.token_address {
```
ezekiiel, line 164: Love this.

ezekiiel (approved):
This is great, as always. :) Thanks for the tests.

One thought: right now we're doing these checks to make sure we're not using an existing staking contract. We could avoid needing to do those by making that configuration an invalid instantiate message.

For example, we could combine the cw20 and staking information into an enum which enumerates the valid configurations. Invalid messages would then fail to deserialize and mean we wouldn't have to make those checks.

I'm happy with this current setup, but if you're feeling fancy that would be nice.

ezekiiel:
> New Staking - New Token
> New Staking - Existing Token
> Existing Staking - Existing Token

This is what I was imagining. Happy to just merge this as is though. Having it written out seems like the size of that enum might make it a little cumbersome. I can think more about this with the new contracts. :)

Callum-A:
> This is what I was imagining. Happy to just merge this as is though. Having it written out seems like the size of that enum might make it a little cumbersome. I can think more about this with the new contracts. :)

I agree the enum names might become unwieldly. Happy for other inputs on this however.

JakeHartnell (approved):

JakeHartnell (approved):
LGTM

PR #190, opened 2022-03-07
Core Contract Implementation

ezekiiel:
This adds a new `cw-governance` contract and two modules intended for
testing: a voting module (`cw20-balance-voting`) and a governance
module (`cw-govmod-sudo`).

These contracts implement the spec described in [this](https://docs.google.com/document/d/1OjUZAl8ty6jZ59ryEyUUuifzectb3RigUSa8RLm4GsY/edit?usp=sharing) design document.

The `cw-governance` contract will serve as the governance contract for
both DAOs and multisigs in the v1 contracts. Before we can deploy a
frontend for these new contracts the following contracts will need to
be completed:

1. A list contract. WIP [here](https://github.com/DA0-DA0/dao-contracts/pull/179).
2. A voting power module based around staked token balances. This
   ought to interact with the current staking contract via the new
   hooks being added. Hooks are being worked on
   [here](https://github.com/DA0-DA0/dao-contracts/pull/180).
3. A voting power module which implements the current cw4-groups
   functionality.
4. A governance module which implements the current proposal
   functionality.
5. A cw20 treasury contract.
6. A NFT treasury contract.

The treasury contracts are a new edition to the design. Treasury
contracts serve as a listing of assets associated with a governance
contract. They ought to implement the appropriate receive messages for
their asset (for example, cw20 receive) and support toggling
receiving an asset triggering that asset being added to their asset
lists.

I believe that this approach is preferable to implementing these
methods on the DAO contract as it makes the governance contract more
flexible and reduces its attack surface. For example, this means that
the addition of a new NFT specification or modification to an existing
one's receive messages does not require a migration of the governance
contract.

These treasury contracts may be added to the DAO as items and frontend
will be responsible for rendering their contents.

### Reviewing

For reviewing this PR I'd recommend one starts by reading the design
doc linked above. Having read that, take a look at the
`cw-governance-macros` package which provides the macros which other
packages implementing the needed interfaces must use on their query
enums.

The `cw-governance-interface` package may be interesting as well as it
uses those macros to derive a standard query enum as well as the
relevant query response types.

The `cw-governance` package is far an away the most important contract
here. The `cw-govmod-sudo` and `cw20-balances-voting` contracts exist
for testing purposes only and are unlikely to ever be used in
production.


ezekiiel:
Note for self: need to add `may_propose` query type to voting modules. This enables proposal deposits. Governance modules can't be responsible for handling this as doing so would make an assumption about the underlying system for allowing proposals. For example, in a NFT voting DAO a "proposal deposit" is likely to look much different than in a cw20 based one.

contracts/cw-governance-interface/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "cw-governance-interface"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0-beta5" }
```
orkunkl, line 9: new cw versions are here

contracts/cw-governance-interface/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod voting;
```
orkunkl, line 1: This module should be, rather a package like https://github.com/CosmWasm/cw-plus/tree/main/packages/cw20/src

orkunkl, line 1: This module should be, rather a package like https://github.com/CosmWasm/cw-plus/tree/main/packages/cw20/src

contracts/cw-governance/src/contract.rs
```diff
@@ -0,0 +1,359 @@
...
+        .map(|info| info.into_wasm_msg(env.contract.address.clone()))
+        .map(|wasm| SubMsg::reply_on_success(wasm, GOV_MODULE_REPLY_ID))
+        .collect();
+
+    Ok(Response::default()
+        .add_attribute("action", "execute_update_governance_modules")
+        .add_submessages(to_add))
+}
+
+pub fn execute_set_item(
```
orkunkl, line 175: Hmm leftover code?

contracts/cw-governance-interface/src/lib.rs
JakeHartnell: Yes please.

contracts/cw-governance/src/contract.rs
JakeHartnell: Yeah, what does this do? What's the use case?

contracts/cw-governance/src/contract.rs
```diff
@@ -0,0 +1,359 @@
...
+) -> Result<Response, ContractError> {
+    let addr = deps.api.addr_validate(&addr)?;
+    ITEMS.save(deps.storage, key.clone(), &addr)?;
+    Ok(Response::default()
+        .add_attribute("action", "execute_set_item")
+        .add_attribute("key", key)
+        .add_attribute("addr", addr))
+}
+
+pub fn execute_remove_item(deps: DepsMut, key: String) -> Result<Response, ContractError> {
```
JakeHartnell, line 188: Ditto @orkunkl's comment above her. What's this for?

JakeHartnell (changes requested):
Requesting changes so this isn't accidentally merged too soon. :smile:

contracts/cw-governance/src/contract.rs
```diff
@@ -0,0 +1,359 @@
...
+        }
+        ExecuteMsg::SetItem { key, addr } => execute_set_item(deps, key, addr),
+        ExecuteMsg::RemoveItem { key } => execute_remove_item(deps, key),
+    }?;
+
+    Ok(response.add_attribute("sender", info.sender))
+}
+
+pub fn execute_proposal_hook(msgs: Vec<CosmosMsg<Empty>>) -> Result<Response, ContractError> {
+    Ok(Response::default()
```
orkunkl, line 106: I could still not wrap the design around my head, but reflecting the proposal hook without any auth looks scary at first sight.

contracts/cw-governance/src/contract.rs
```diff
@@ -0,0 +1,359 @@
...
+            config.image_url.unwrap_or_else(|| "None".to_string()),
+        ))
+}
+
+pub fn execute_update_voting_module(
+    env: Env,
+    module: ModuleInstantiateInfo,
+) -> Result<Response, ContractError> {
+    let wasm = module.into_wasm_msg(env.contract.address);
+    let submessage = SubMsg::reply_on_success(wasm, VOTE_MODULE_UPDATE_REPLY_ID);
```
orkunkl, line 133: DAOs should be able to select different voting systems for different proposals 🤔

contracts/cw-governance/src/contract.rs
```diff
@@ -0,0 +1,359 @@
...
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
+        ExecuteMsg::ExecuteProposalHook { .. } => {
+            if !GOVERNANCE_MODULES.has(deps.storage, info.sender.clone()) {
+                return Err(ContractError::Unauthorized {});
+            }
+        }
+        _ => {
+            if info.sender != env.contract.address {
```
orkunkl, line 85: instead of pattern matching, using if-else clauses could make this code much simpler to read.

contracts/cw-governance/src/contract.rs
ezekiiel: There is validation above that checks that this is coming from a governance module in the execute method. I think it makes sense to move that into the function though for future readers though.

contracts/cw-governance/src/contract.rs
ezekiiel: Agree. I'm just going to move this validation into the execute methods themselves. At one point it was a lot simpler to do it like this, but now this code is a little unwieldy.

contracts/cw-core/examples/schema.rs
```diff
@@ -3,8 +3,10 @@ use std::fs::create_dir_all;
 
 use cosmwasm_schema::{export_schema, remove_schemas, schema_for};
 
-use cw_governance::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
-use cw_governance::query::DumpStateResponse;
+use cw_core::{
```
orkunkl, line 6: `cw_core` feels like it is a package that is part of cosmwasm std or smth.
Maybe best to rename to `cw_dao_core` or `cw_da0_core`

JakeHartnell: > `cw_core` feels like it is a package that is part of cosmwasm std or smth. Maybe best to rename to `cw_dao_core` or `cw_da0_core`

Completely agree with this.

JakeHartnell: +1 cw_dao_core

janitachalam (approved):

Callum-A (approved):

JakeHartnell (approved):
Just need to fix merge conflicts. :tada:

contracts/cw-core/examples/schema.rs
JakeHartnell: Still want this to happen... will make a ticket.

JakeHartnell: https://github.com/DA0-DA0/dao-contracts/issues/255

PR #193, opened 2022-03-14
Contracts v1 - single choice voting module.

ezekiiel:
This is an implementation of single choice voting for the v1
contracts. It is largely based on the existing single choice voting
implementation and thus inherits some of its issues (more discussion
of this below).

This implementation also fixes an overflow bug and an issue with
rejected proposals not being closable.

#### Vote ordering matters.

Our present voting system (and the ones used by the cw-plus voting
contracts) determine if a vote passes or fails by which option (yes or
no) reaches the threshold first. For example, with a 50% passing
threshold and 100 tokens there are two possible outcomes for the same
set of votes depending on ordering:

1. Wallet A votes no first with 50 tokens - proposal fails.
2. Wallet B votes yes first with 50 tokens - proposal passes.

This also has interesting effects when dealing with `ThresholdQuorum`
passing thresholds. These behaviors are currenly demonstrated in the
`test_close_cotes_quorum` and `test_close_votes` tests.

I am unsure what the appropriate solution to this is, but think we
should put our heads together a little.

#### Early proposal expiration reduces hostile takeover cost.

Proposals currently have a `latest` field which the proposer may
set. This field overrules the governance module's default max voting
period if it is smaller than the default. The result of this is that
in governance systems with a `ThresholdQuorum` the cost of a hostile
takeover becomes `quorum * total_supply * token_price` as a hostile
actor may do the following:

1. Purchase `quorum * total_supply` tokens.
2. Create a proposal that expires the block after it is created.
3. Vote yes on that proposal in the same block that it is created.

To mitigate this I think we should remove the `latest` field from
proposals.


contracts/cw-govmod-single/src/contract.rs
```diff
@@ -0,0 +1,411 @@
...
+    error::ContractError,
+    msg::{ExecuteMsg, InstantiateMsg, QueryMsg},
+    proposal::{advance_proposal_id, Proposal, Status, Vote, Votes},
+    query::{ProposalResponse, VoteInfo, VoteListResponse, VoteResponse},
+    state::{Ballot, Config, BALLOTS, CONFIG, PROPOSALS, PROPOSAL_COUNT},
+    threshold::Threshold,
+    utils::{get_total_power, get_voting_power},
+};
+
+const CONTRACT_NAME: &str = "crates.io:cw-govmod-single";
```
JakeHartnell, line 21: We should come up with a good naming scheme for these modules...

JakeHartnell (approved):
:rocket:

contracts/cw-govmod-single/src/contract.rs
```diff
@@ -0,0 +1,411 @@
...
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    msg.threshold.validate()?;
+
+    let dao = info.sender;
+
+    let config = Config {
+        threshold: msg.threshold,
+        max_voting_period: msg.max_voting_period,
+        only_members_execute: msg.only_members_execute,
```
janitachalam, line 43: just curious, are there situations in which it's important for non-members to be able to execute?

contracts/cw-govmod-single/src/contract.rs
```diff
@@ -0,0 +1,411 @@
...
+    info: MessageInfo,
+    threshold: Threshold,
+    max_voting_period: Duration,
+    only_members_execute: bool,
+    dao: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    threshold.validate()?;
+    let dao = deps.api.addr_validate(&dao)?;
```
janitachalam, line 293: nit: dao_addr?

contracts/cw-govmod-single/src/tests.rs
```diff
@@ -0,0 +1,680 @@
...
+    );
+}
+
+/// The current behavior for passing proposals is that the first
+/// option to reach the threshold wins. For example, with a 50%
+/// passing threshold if 50% of voting power votes no that proposal
+/// fails even if the other 50% would have voted yes. The same goes if
+/// the yes and no were reversed.
+///
+/// TODO(zeke): is this the behavior that we want?
```
janitachalam, line 444: I would vote that if it's still possible for the proposal to pass, it should not be rejected

contracts/cw-govmod-single/src/contract.rs
JakeHartnell: Yes, if you have cron based execution for example (an outside bot that looks for passed proposals and automatically executes them).

Issue #194, opened 2022-03-14
Add a voting power module based on staked token balances.


ezekiiel:
The last contract to start for the v1 contracts is a voting power module based on staked token balance. An example voting power module can be found [here](https://github.com/DA0-DA0/dao-contracts/tree/zeke/contracts-v1/contracts/cw20-balance-voting) which demonstrates implementing the voting power interface.

I believe that this contract can be separated from the current staking contract and register itself as a hook receiver to handle balance updates. Would be nice to support vote delegation here as well but I think we can do that as a follow up issue.

Callum-A:
Going to take a look at this, assigning to myself for now.

Callum-A:
Closed with #199

PR #199, opened 2022-03-15
Staked balances v1 voting contract

Callum-A:
Addresses #194 
Staked balances V1 voting contract

Also allows existing staking contract to be used as per my existing PR.

Full test coverage.

Feedback appreciated


contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,257 @@
...
+                    staking_code_id,
+                    unstaking_duration,
+                } => {
+                    let msg = WasmMsg::Instantiate {
+                        code_id: staking_code_id,
+                        funds: vec![],
+                        admin: Some(env.contract.address.to_string()),
+                        label: env.contract.address.to_string(),
+                        msg: to_binary(&stake_cw20::msg::InstantiateMsg {
+                            admin: Some(env.contract.address.to_string()),
```
ezekiiel, line 71: Same thing here. We probably either want to make the DAO the admin or add methods to this contract to let the DAO manage it. Probably simpler to just make the DAO the admin. :)

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,257 @@
...
+        staking_contract,
+        &stake_cw20::msg::QueryMsg::StakedBalanceAtHeight {
+            address: address.to_string(),
+            height: Some(env.block.height),
+        },
+    )?;
+    to_binary(
+        &cw_governance_interface::voting::VotingPowerAtHeightResponse {
+            power: res.balance,
+            height: env.block.height,
```
ezekiiel, line 179: We likely want to use the height returned by the staked balance query here. I think that in the event that someone has staked on the same block that we query on the `height` field will be the previous block because the state change hasn't been committed yet.

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,257 @@
...
+    let res: stake_cw20::msg::TotalStakedAtHeightResponse = deps.querier.query_wasm_smart(
+        staking_contract,
+        &stake_cw20::msg::QueryMsg::TotalStakedAtHeight {
+            height: Some(env.block.height),
+        },
+    )?;
+    to_binary(
+        &cw_governance_interface::voting::TotalPowerAtHeightResponse {
+            power: res.total,
+            height: env.block.height,
```
ezekiiel, line 195: Same as other comment. Probably want to use the height returned by the query.

ezekiiel (commented):
AMAZING! Just a couple very small comments. My vote would be to merge this and then look at integrating hooks once other things are more far along as this is likely the most functional contracts-v1 contract at the moment. :)

JakeHartnell (approved):

JakeHartnell (approved):

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,272 @@
...
+    _env: Env,
+    _info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {}
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
```
ezekiiel, line 146: What do you think about adding a query here to get the DAO that this contract is associated with? I don't see it being used terribly often but seems like it could be useful + we already have that info stored in state.

contracts/cw20-staked-balance-voting/src/msg.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {}
+
+#[voting_query]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
+    TokenContract {},
```
ezekiiel, line 54: Can actually derive this option with `#[token_query]`.

ezekiiel (approved):
Great work!

PR #200, opened 2022-03-16
introduce multiple choice proposals

janitachalam:
Introduce multiple choice proposals which will be integrated into the v1 (modular) version of contracts.

Some notes / unresolved issues: 

1. This implements multiple choice voting where only a single choice may be selected by each voter. I envision ranked-choice voting as another type of voting contract (not difficult to implement) given that the logic for calculating the votes diverges significantly.
2. There is a lot of duplicated code that could be shared between the different contracts / I ended up copying some code myself but I didn't do anything about it for now because I figured this may be less of an issue after the contracts refactor. 
3. The way that I interpreted thresholds with multiple choice voting is that in the AbsolutePercentage case, the threshold must be met by any of the given choices for it to pass. In the quorum case, there must be quorum votes cast in total for any of the choices, and threshold of those votes must be met by any of the given choices. 
4. Haven't determined what to do about the case where there is more than one choice with equal weight of votes.


JakeHartnell:
Will be great to have this for testing how multiple voting modules might play together. :pray:

ezekiiel:
This is so great. :) Haven't looked much at the code yet but thanks so much for putting it together and even integrating it with the new contracts design! I've got a single choice voting module that I'm working on now. Suspect there will be code duplication here. I'll try and find those spots and then break out some packages we can use.

JakeHartnell:
Let's try an get this into the `v1-contracts` branch! It will need a bit of a refactor, but I think it's a good opportunity to test out the new DAO contracts design.

Will also be useful for testing to have two governance modules.

janitachalam:
> Let's try an get this into the `v1-contracts` branch! It will need a bit of a refactor, but I think it's a good opportunity to test out the new DAO contracts design.
> 
> Will also be useful for testing to have two governance modules.

Sounds good, I can work on integrating it

contracts/cw3-multiple-choice/src/msg.rs
```diff
@@ -0,0 +1,165 @@
...
+/// The total_weight used for calculating success as well as the weights of each
+/// individual voter used in tallying should be snapshotted at the beginning of
+/// the block at which the proposal starts (this is likely the responsibility of a
+/// correct cw4 implementation).
+/// See also `ThresholdResponse` in the cw3 spec.
+/// This is similar to thresholds for 'yes' votes in single choice voting, but in this case in order for any
+/// of the multiple choice options to pass, it must have threshold votes in its favor.
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum Threshold {
```
ezekiiel, line 109: There are a ton of subtle details we've found in the new implementation that you'll likely want to move in here. See this comment for an overview: https://github.com/DA0-DA0/dao-contracts/blob/00e9b4d70848c4f79c519dd9a9f184373f99681c/contracts/cw-govmod-single/src/threshold.rs#L7-L38

I can make another pass later and go over the logic here for that sort of stuff.

contracts/cw3-multiple-choice/src/msg.rs
```diff
@@ -0,0 +1,165 @@
...
+        match self {
+            Threshold::AbsolutePercentage {
+                percentage: percentage_needed,
+            } => valid_percentage(percentage_needed),
+            Threshold::ThresholdQuorum {
+                threshold,
+                quorum: quroum,
+            } => {
+                valid_percentage(threshold)?;
+                valid_percentage(quroum)
```
ezekiiel, line 133: We likely want to support zero percent quorums (allows purely plurality voting like US elections). For `cw-govmod-single` I added a `validate_quorum` function to handle this.

contracts/cw3-multiple-choice/src/state.rs
```diff
@@ -0,0 +1,172 @@
...
+        status
+    }
+
+    /// update_status sets the status of the proposal to current_status.
+    /// (designed for handler logic)
+    pub fn update_status(&mut self, block: &BlockInfo) {
+        self.status = self.current_status(block);
+    }
+
+    pub fn is_passed(&self, block: &BlockInfo) -> bool {
```
ezekiiel, line 76: We likely also want a `is_rejected` function here as well to allow proposals to be properly shown as rejected in the UI when there is no way they can pass. You should be able to tweak this one: https://github.com/DA0-DA0/dao-contracts/blob/5483427cfedb1ae2844a6e403369c7f3d601a6d1/contracts/cw-govmod-single/src/proposal.rs#L128

contracts/cw3-multiple-choice/src/state.rs
```diff
@@ -0,0 +1,172 @@
...
+        match self.threshold {
+            Threshold::AbsolutePercentage {
+                percentage: percentage_needed,
+            } => {
+                for c in &self.votes.vote_counts {
+                    if *c >= votes_needed(self.total_weight, percentage_needed) {
```
ezekiiel, line 82: What happens here if multiple options receive over `percentage_needed`? For example, with three options and a 33% passing threshold they could all get the needed votes. I think we should do some brainstorming about how we should do thresholds for multiple choice voting.

ezekiiel: Also, we should consider how no votes should work with multiple choice voting. For example, if I create a proposal with three options, all of which send the entire treasury to me, how should we determine that none have passed? Can people vote no for the entire thing?

ezekiiel (commented):
Really, really, really great work. Thank you! This is a monumental task. Super cool that you've put this together. Left a bunch of comments and can make another pass on governance specifics / overflows later once we have things ironed out.

Might be worthwhile to make a thread in the Discord to discuss how we should handle thresholds for multiple choice votes.

ezekiiel:
@janitachalam can we close this in favor of https://github.com/DA0-DA0/dao-contracts/pull/200?

JakeHartnell:
Closing in favor of #279

PR #201, opened 2022-03-18
Update local deploy script

the-frey:
Juno now has configuration options to switch on CORS for local dev, meaning we can simplify the deployment file.


scripts/deploy_local.sh
```diff
@@ -4,41 +4,37 @@
...
+    -e STAKE_TOKEN=$DENOM \
+    -e GAS_LIMIT="$GAS_LIMIT" \
+    -e UNSAFE_CORS=true \
+    -p 1317:1317 -p 26656:26656 -p 26657:26657 \
     --mount type=volume,source=junod_data,target=/root \
-    ghcr.io/cosmoscontracts/juno:$IMAGE_TAG /opt/setup_junod.sh $1
-
-# Add custom app.toml to junod_data volume
-docker run -v junod_data:/root --name helper busybox true
-docker cp docker/app.toml helper:/root/.juno/config/app.toml
```
JakeHartnell, line 34: Can probably delete this file now too! Thanks for this.

JakeHartnell: `rm -rf docker`

JakeHartnell (approved):

scripts/deploy_local.sh
the-frey: Yep, good point. Forgot that

JakeHartnell (approved):

PR #206, opened 2022-03-21
Move voting and interface logic into packages.

ezekiiel:
This moves logic around voting (types / meets threshold functions) and the interface / macros packages into a `packages/` subfolder.


JakeHartnell (approved):

Callum-A (approved):
Yeah this all looks good to me!

PR #210, opened 2022-03-22
Consistent use of `Name {}` style enums.

### LINKED ISSUE #209: Enum variants should all use `Name {}` form.
In the interest of being consistent and making enums easy to write out in the frontend we should use the `Name {}` form for all our enum variants. I messed some of these up while writing the `cw-govmod-single` and `cw-governance` contract.

Specifically, the `Admin` and `TokenInfo` enums both have fields that don't follow this format.

ezekiiel:
Resolves #209.


Callum-A (approved):
Jumping in just to give my stamp of approval, this all looks good : )

PR #211, opened 2022-03-22
fix local deploy script 

janitachalam:
InstantiateMsg for cw3 dao contract had some new fields added recently


JakeHartnell (approved):
:pray:

entrancedjames (approved):

PR #212, opened 2022-03-22
Add DAO initial balance to staked governance token instantiation

### LINKED ISSUE #208: Support DAO initial balance in staked balance voting module.
The old contracts support allocating a token amount to the DAO on instantiation. We should support this in the `TokenInfo::New` variant.

NoahSaso:
Resolves #208 

Sending part of the initial token supply to the DAO was left out of the contract v1 rewrite. This adds it back in.

### DIFF
--- a/contracts/cw20-staked-balance-voting/src/contract.rs
+++ b/contracts/cw20-staked-balance-voting/src/contract.rs
@@ -5,6 +5,7 @@ use cosmwasm_std::{
     Uint128, WasmMsg,
 };
 use cw2::set_contract_version;
+use cw20::Cw20Coin;
 use cw_utils::parse_reply_instantiate_data;
 
 use crate::error::ContractError;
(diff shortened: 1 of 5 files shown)


Callum-A (approved):
Cheers for picking this up Noah, this all looks good to me.

JakeHartnell (approved):

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -89,11 +90,21 @@ pub fn instantiate(
...
+            if let Some(initial_dao_balance) = msg.initial_dao_balance {
+                if initial_dao_balance > Uint128::zero() {
+                    initial_balances.push(Cw20Coin {
+                        address: info.sender.to_string(),
+                        amount: initial_dao_balance,
+                    });
+                }
+            }
+
             let initial_supply = initial_balances
```
ezekiiel, line 108: Likely want to move push of DAO balance after this check. This allows a DAO to be created where the only address with a balance is a DAO.

PR #216, opened 2022-03-23
Named groups contract

### LINKED ISSUE #214: Implement a named groups contract.
This contract should allow DAOs to manage list of addresses and associated them with a named group. It should support adding and removing addresses from a group as well as listing all members of a group and querying for what group an address is in.

For example: a dao may create a `core team` group with a set of addresses. A governance module may then only allow proposal creation by members of that group.

This contract may be added / removed from a DAO as an item via the the `SetItem` and `RemoveItem` execute messages. Likely it will be standard that this contract is stored under the `groups` key.

In order to make it so that a DAO may be instantiated with a group (this is important for DAO creation UX) we should add the ability for a DAO to be instantiated with a set of items. This can be done by adding a field to the `cw-governance` contract's instantiate message which takes a mapping of `name -> ModuleInstantiateInfo`. On instantiation the DAO should dispatch submessages to create those items and then store them in its item map.

NoahSaso:
Resolves #214 

Creates a named group of address system, such that addresses can be grouped together for various purposes. This could be used in governance modules to give certain addresses special permissions, such as a `core team` making certain types of proposals. Or maybe only certain groups can vote on certain proposals. Anything can happen!

It can be instantiated with a set of initial groups or not.

Execute:
- Update
- RemoveGroup
- UpdateOwner

Query:
- Dump [all groups]
- ListGroups
- ListAddresses
- IsAddressInGroup


contracts/cw-named-groups/src/contract.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    // Load groups into dump.
+    let groups = GROUPS.keys(deps.storage, None, None, Order::Ascending);
+    for group in groups.flatten() {
+        let addresses = GROUPS.load(deps.storage, &group)?;
+        dump.push(Group {
+            name: group,
+            addresses: addresses.into_iter().map(|addr| addr.to_string()).collect(),
+        });
+    }
```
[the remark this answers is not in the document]
ezekiiel: This sort of stuff isn't terribly important imo, but this is liable to be gas sensitive so might as well be as effective as we can.

contracts/cw-named-groups/src/contract.rs
```diff
@@ -0,0 +1,286 @@
...
+        addresses: addrs.into_iter().collect(),
+    })
+}
+
+fn query_list_groups(deps: Deps, address: String) -> StdResult<ListGroupsResponse> {
+    // Validate address.
+    let addr = deps.api.addr_validate(&address)?;
+    // Return groups, or an empty set if failed to load (address probably doesn't exist).
+    // It doesn't make sense to ask for the addresses in a group if the group doesn't exist, which is why
+    // we don't return an error in query_list_addresses; however, here in query_list_groups, it makes sense
```
ezekiiel, line 217:
```suggestion
    // we return an error in query_list_addresses; however, here in query_list_groups, it makes sense
```

contracts/cw-named-groups/src/helpers.rs
```diff
@@ -0,0 +1 @@
+
```
ezekiiel, line 1: Probably want to remove this file.

contracts/cw-named-groups/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+pub mod contract;
+mod error;
+pub mod helpers;
```
ezekiiel, line 3:
```suggestion
```

contracts/cw-named-groups/src/msg.rs
```diff
@@ -0,0 +1,58 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    Add {
+        group: String,
+        addresses: Option<Vec<String>>,
+    },
+    Remove {
+        group: String,
+        addresses: Option<Vec<String>>,
+    },
```
ezekiiel, line 28: What do you think about adding an execute message to change the owner of this contract? This might be useful if a DAO needs to migrate its router contract.

contracts/cw-named-groups/src/msg.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
+    Dump {},
+    ListGroups { address: String },
+    ListAddresses { group: String },
```
ezekiiel, line 36: We should paginate these queries. Dump is OK to run out of gas, but for these it would be nice if they didn't. See `ListProposals` in `contracts/cw-governance` for an example of doing this.

contracts/cw-named-groups/src/msg.rs
```diff
@@ -0,0 +1,58 @@
+use std::collections::HashSet;
+
+use cosmwasm_std::Addr;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Group {
+    pub name: String,
+    pub addresses: HashSet<String>,
```
ezekiiel, line 10: When this comes in as JSON how does this look? Will the frontend just provide a list? Will deserialization fail if the provided list has duplicate values?

ezekiiel: Might be worthwhile to just take a `Vec<String>` in the instantiate message if these don't have clean answers. Can't find anything about this in the `serde_json` docs.

contracts/cw-named-groups/src/tests.rs
```diff
@@ -0,0 +1,630 @@
...
+
+    #[test]
+    fn instantiate_with_groups() {
+        let groups = vec![group_factory(1), group_factory(2), group_factory(3)];
+
+        let (app, contract_addr) = instantiate(Some(groups.clone())).unwrap();
+
+        // Ensure there are the expected groups.
+        let dump_result = dump(&app, &contract_addr);
+        assert_eq!(dump_result.groups, groups);
```
ezekiiel, line 75: Let's test that the `ADDRESSES` map is properly instantiated here as well. Don't think that works at the moment.

ezekiiel (commented):
Great work. Thanks for putting this together. Only thing we really need here is some pagination in the list queries. :)

contracts/cw-named-groups/src/contract.rs
```diff
@@ -0,0 +1,286 @@
...
+    if let Some(addresses) = addresses {
+        // Only attempt to add if addresses are provided.
+        if !addresses.is_empty() {
+            // Validate addresses.
+            let addrs = validate_addresses(deps.api, addresses)?;
+
+            // Add group to each address's group list.
+            for addr in addrs.iter() {
+                add_to_map(deps.storage, ADDRESSES, addr.clone(), vec![group.clone()])?;
+            }
```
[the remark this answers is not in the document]
NoahSaso: My recently acquired understanding of Rust tells me I have to clone group either way since it's in a loop.

contracts/cw-named-groups/src/contract.rs
```diff
@@ -0,0 +1,286 @@
...
+    if let Some(addresses) = addresses {
+        // Only attempt to remove if addresses are provided.
+        if !addresses.is_empty() {
+            // Validate addresses.
+            let addrs = validate_addresses(deps.api, addresses)?;
+
+            // Remove group from each address's group list.
+            let group_set_to_remove = vec![group.clone()];
+            for addr in addrs.iter() {
+                remove_from_map(
```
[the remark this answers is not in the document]
NoahSaso: They are each called with one of each: a single element vector and arbitrary length vector.

contracts/cw-named-groups/src/msg.rs
NoahSaso: Good catch! Should have been a `Vec` :)

contracts/cw-named-groups/src/tests.rs
NoahSaso: What do you mean? `GROUPS.save` on line 58 is only called if there are any initial groups, and `ADDRESSES.save` is the same. It doesn't seem like there's any function to instantiate an empty Map, as `.save` and `.update` take keys and values to store. The API makes it opaque if the underlying data store has been instantiated or not, but I assume just the `Map::new` in `state.rs` takes care of that. What am I missing?

NoahSaso: I understand what you mean! Another test should be stronger, not this one. Resolving this because I added more code to verify both `List...` queries return the right data after instantiating with initial groups.

Callum-A (approved):
Clean work!

orkunkl:
The storage could become expensive, let's keep this unmerged, I will work out a new design on a new PR targeting this one.

orkunkl:
You don't need HashSet it will add hashing runtime complexity which could become expensive in a blockchain context.
Multi-key design will ensure the uniqueness of the member.
The change will be bigger than expected, it is better for me to throw my state design here. Something like this will do:

```rust
pub struct Groups<'a>{
   groups: Map<'a, (&'a str, &'a Addr), bool>,
   addr_index: Map<'a, (&'a Addr, &'a str), bool>
}

impl<'a> Groups<'a> {
    pub const fn new(storage_key: &'a str) -> Self {
        Groups{
            groups: Map::new(storage_key),
            // TODO: fix key
            addr_index: Map::new(storage_key)
        }
    }

    /// No need for a create_group contract, add will create non existing groups too
    pub fn add<'a>(
        &self,
        storage: &mut dyn Storage,
        name: &'a str,
        addresses: &[Addr]
    ) -> StdResult<()> {
        for addr in addresses {
            self.groups.save(storage, (name, addr), &true)?;
            self.addr_index.save(storage, (addr, name), &true)?;
        }

        Ok(())
    }

    pub fn delete<'a>(
        &self,
        storage: &mut dyn Storage,
        name: &'a str,
        addresses: &[Addr]
    ) -> StdResult<()> {
        for addr in addresses {
            // TODO: do existing check
            self.groups.remove(storage, (name, addr))?;
            self.addr_index.remove(storage, (addr, name))?;
        }

        Ok(())
    }
}

pub const GROUPS: Groups = Groups::new("groups");
```

You don't even need separate functions for add delete and update. You can do something similar to https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw4-group/src/contract.rs#L114

contracts/cw-named-groups/src/contract.rs
```diff
@@ -222,19 +193,34 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
 }
 
 fn query_dump(deps: Deps) -> StdResult<DumpResponse> {
-    // Create dump of all groups.
-    let dump = GROUPS
+    let mut groups_map: HashMap<String, Vec<String>> = HashMap::new();
```
orkunkl, line 196: No need for hash map here too. It will use hashing internally for keys. Just use `Vec<(String, Vec<String)>`

contracts/cw-named-groups/src/contract.rs
```diff
@@ -222,19 +193,34 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
 }
 
 fn query_dump(deps: Deps) -> StdResult<DumpResponse> {
```
orkunkl, line 195: This part could be costly. I will optimize here using `Iterator`.

contracts/cw-named-groups/src/contract.rs
```diff
@@ -245,8 +231,12 @@ fn query_list_addresses(
     offset: Option<u32>,
     limit: Option<u32>,
 ) -> StdResult<ListAddressesResponse> {
-    let addresses = GROUPS
-        .load(deps.storage, &group)
+    // Retrieve all addresses under this group, returning error if group not found.
+    let addresses: Vec<Addr> = GROUPS
+        .groups_to_addresses
+        .prefix(&group)
```
orkunkl, line 237: 🚀

contracts/cw-named-groups/src/contract.rs
```diff
@@ -268,12 +258,16 @@ fn query_list_groups(
...
+    // Return groups, or an empty vec if failed to load (address probably doesn't exist).
     // It doesn't make sense to ask for the addresses in a group if the group doesn't exist, which is why
     // we return an error in query_list_addresses; however, here in query_list_groups, it makes sense
-    // to return an empty list when an address is not in any groups since conceptually the structure
-    // is One Group to Many Addresses.
-    let groups = ADDRESSES.load(deps.storage, addr).unwrap_or_default();
+    // to return an empty list when an address is not in any groups since this is a valid case.
+    let groups = GROUPS
+        .addresses_to_groups
+        .prefix(&addr)
```
orkunkl, line 267: Let's move this function under `Group` struct, best to hide complexity

orkunkl (approved):
LGTM

Callum-A (approved):
Awesome works!

PR #217, opened 2022-03-23
Add `ReverseProposals` query to cw-govmod-single.

### LINKED ISSUE #213: Add missing queries in `cw-govmod-single`.
The module is currently missing the `ReverseProposals` and `Tally` queries. These are reasonably frequently used in the frontend.

ezekiiel:
Resolves #213 

This adds the `ReverseProposals` query to `cw-govmod-single`. It does not add the `Tally` query as in the new design that query is strictly a subset of the `Proposal` query with a little extra work to compute the current turnout percent on the contract side.

Where possible we want to limit the amount of contract side compute happening and limit complexity in the contracts so I'm in favor of just not adding it.

### DIFF
--- a/contracts/cw-govmod-single/src/contract.rs
+++ b/contracts/cw-govmod-single/src/contract.rs
@@ -14,6 +14,7 @@ use crate::{
     error::ContractError,
     msg::{DepositInfo, ExecuteMsg, InstantiateMsg, QueryMsg},
     proposal::{advance_proposal_id, Proposal, Status},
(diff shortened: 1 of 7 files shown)


ezekiiel:
Unrelated but this also loosens the quorum requirements to allow for a zero percent quorum.

Callum-A (approved):
Nice works, all looks good on my end!

JakeHartnell (approved):

PR #220, opened 2022-03-24
Fix threshold for no for non v1 side contracts

Callum-A:
Need to get some eyes on this to check my logic all works clear.

New logic is no votes can reject proposals when they are greater than 1 - threshold. (Or for absolute count total_weight - count). 

Going to need some eyes on this to sanity check.

### DIFF
--- a/contracts/cw3-dao/src/state.rs
+++ b/contracts/cw3-dao/src/state.rs
@@ -103,22 +103,19 @@ impl Proposal {
         self.status = self.current_status(block);
     }
 
-    /// Helper function to check if a certain vote count has reached threshold.
-    /// Only called from is_rejected and is_passed for no and yes votes
-    /// Handles the different threshold types accordingly.
-    /// This function returns true if and only if vote_count is greater than the threshold which
-    /// is calculated.
-    /// In the case where we use yes votes, this function will return true if and only if the
-    /// proposal will pass.
-    /// In the case where we use no votes, this function will return true if and only if the
-    /// proposal will be rejected regardless of other votes.
-    fn does_vote_count_reach_threshold(&self, vote_count: Uint128, block: &BlockInfo) -> bool {
+    /// Returns true if this proposal is sure to pass (even before expiration, if no future
+    /// sequence of possible votes could cause it to fail).
+    pub fn is_passed(&self, block: &BlockInfo) -> bool {
         match self.threshold {
             Threshold::AbsolutePercentage {
                 percentage: percentage_needed,
             } => {
-                vote_count
-                    >= votes_needed(self.total_weight - self.votes.abstain, percentage_needed)
+                let opinions = self.total_weight - self.votes.abstain;
+                // If all abstain we cannot pass the proposal
+                if opinions.is_zero() {
+                    return false;
+                }
+                self.votes.yes >= votes_needed(opinions, percentage_needed)
             }
             Threshold::ThresholdQuorum { threshold, quorum } => {
                 // we always require the quorum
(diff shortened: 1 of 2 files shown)


ethanfrey:
Nice, the main issue is it checked if no passed threshold, when it should check if no passes `1-threshold`.

This is a good fix.

ezekiiel:
Just need to add the logic around the case where the passing threshold is 100% as discussed in Discord. https://github.com/DA0-DA0/dao-contracts/pull/221

contracts/cw3-dao/src/state.rs
```diff
@@ -128,27 +121,112 @@ impl Proposal {
...
-    /// already rejected.
+    /// Returns true if this proposal is sure to be rejected (even before expiration, if
+    /// no future sequence of possible votes could cause it to pass).
     pub fn is_rejected(&self, block: &BlockInfo) -> bool {
-        self.does_vote_count_reach_threshold(self.votes.no, block)
+        match self.threshold {
+            Threshold::AbsolutePercentage {
+                percentage: percentage_needed,
+            } => {
+                let options = self.total_weight - self.votes.abstain;
```
ezekiiel, line 141: As discussed in the Discord, let's add a special case for if `options` is zero. This will reject proposals in the all abstain case.

ezekiiel: Otherwise this is looking really great. :) Appreciate your help iterating on this threshold stuff.

contracts/cw3-dao/src/state.rs
```diff
@@ -128,27 +121,112 @@ impl Proposal {
                 if self.expires.is_expired(block) {
                     // If expired, we compare vote_count against the total number of votes (minus abstain).
                     let opinions = self.votes.total() - self.votes.abstain;
-                    vote_count >= votes_needed(opinions, threshold)
+                    self.votes.yes >= votes_needed(opinions, threshold)
```
ezekiiel, line 124: Re: "failure on abstain" - if `opinions` is zero here `votes_needed(0, N)` will return `0` for all values of `N`. This means that if all the votes are abstain and `opinions` is zero a proposal will be considered passed even if `self.votes.yes` is zero as `0 >= 0`.

contracts/cw3-dao/src/state.rs
```diff
@@ -128,27 +125,116 @@ impl Proposal {
...
-                    vote_count >= votes_needed(opinions, threshold)
+                    // If they are all abstains the proposal does not pass
+                    if opinions.is_zero() {
+                        return false;
+                    }
+                    self.votes.yes >= votes_needed(opinions, threshold)
                 } else {
                     // If not expired, we must assume all non-votes will be cast against
-                    // vote_count
                     let possible_opinions = self.total_weight - self.votes.abstain;
```
ezekiiel, line 135: Need a check that this isn't zero as as well.

contracts/cw3-multisig/src/state.rs
```diff
@@ -134,27 +131,124 @@ impl Proposal {
...
-                    vote_count >= votes_needed(opinions, threshold)
+                    // If all abstains the proposal does not pass
+                    if opinions == 0 {
+                        return false;
+                    }
+                    self.votes.yes >= votes_needed(opinions, threshold)
                 } else {
                     // If not expired, we must assume all non-votes will be cast against
-                    // vote_count
                     let possible_opinions = self.total_weight - self.votes.abstain;
```
ezekiiel, line 141: Options check here as well.

ezekiiel (approved):
Looks great! Thanks for all help iterating on this.

contracts/cw3-dao/src/state.rs
```diff
@@ -128,27 +125,120 @@ impl Proposal {
...
-    /// already rejected.
+    /// Returns true if this proposal is sure to be rejected (even before expiration, if
+    /// no future sequence of possible votes could cause it to pass).
     pub fn is_rejected(&self, block: &BlockInfo) -> bool {
-        self.does_vote_count_reach_threshold(self.votes.no, block)
+        match self.threshold {
+            Threshold::AbsolutePercentage {
+                percentage: percentage_needed,
+            } => {
+                let options = self.total_weight - self.votes.abstain;
```
janitachalam, line 153: nit: opinions

contracts/cw3-dao/src/state.rs
```diff
@@ -103,22 +103,19 @@ impl Proposal {
...
                 percentage: percentage_needed,
             } => {
-                vote_count
-                    >= votes_needed(self.total_weight - self.votes.abstain, percentage_needed)
+                let opinions = self.total_weight - self.votes.abstain;
+                // If all abstain we cannot pass the proposal
+                if opinions.is_zero() {
+                    return false;
+                }
+                self.votes.yes >= votes_needed(opinions, percentage_needed)
```
entrancedjames, line 118: !opinions.is_zero() && (self.votes.yes >= votes_needed(opinions, percentage_needed))

contracts/cw3-dao/src/state.rs
```diff
@@ -128,27 +125,120 @@ impl Proposal {
                 if self.expires.is_expired(block) {
                     // If expired, we compare vote_count against the total number of votes (minus abstain).
                     let opinions = self.votes.total() - self.votes.abstain;
-                    vote_count >= votes_needed(opinions, threshold)
+                    // If they are all abstains the proposal does not pass
+                    if opinions.is_zero() {
```
entrancedjames, line 129: ditto for this and one below

janitachalam (approved):
LGTM

contracts/cw3-dao/src/state.rs
ezekiiel: Think I like having this broken out into two things. Makes it a little easier on the eyes. :) Possible that we save a branch with the single conditional, but we have much less gas efficient code elsewhere.

Callum-A: I'm not the biggest fan here, it makes it harder to read imo?

entrancedjames: then it would be 

if opinions.is_zero() {
  false 
} else {
  self.votes.yes >= votes_neeeded(opinions, percentage_needed)
}

anyhow thats a nit (just feels weird we return for 1 and not another)

contracts/cw3-multisig/src/state.rs
```diff
@@ -106,25 +106,22 @@ impl Proposal {
...
-            } => vote_count >= weight_needed,
+            } => self.votes.yes >= weight_needed,
             Threshold::AbsolutePercentage {
                 percentage: percentage_needed,
             } => {
-                vote_count
-                    >= votes_needed(self.total_weight - self.votes.abstain, percentage_needed)
+                let opinions = self.total_weight - self.votes.abstain;
+                // If all abstains we cannot pass the propsal
+                if opinions == 0 {
```
entrancedjames, line 121: is this safe, we used Uint128::zero last time, why not here?

entrancedjames (commented):
Did we reach 100% test coverage for the does_vote_count_reach_threshold()

for both multi-sig and cw3-dao. we probably have but I'm just curious here

entrancedjames (approved):
Besides those comments, overall looks good and given we're passing the old tests this means this is working as expected.

contracts/cw3-multisig/src/state.rs
Callum-A: types are different multisig uses u64

PR #221, opened 2022-03-25
v1 contracts - Properly determine if a proposal is rejected.

ezekiiel:
This updates the rejection logic so that proposals are considered
rejected if the number of no votes > 100% - threshold.

This also adds a number of tests that check our behavior in the event
that a proposal results in a tie or a proposal has only abstain
votes. The current behavior, and the behavior of the old contracts is
to pass the proposal in both cases. We should do some deep thinking
about if this is the desired behavior.

### DIFF
--- a/contracts/cw-govmod-single/src/proposal.rs
+++ b/contracts/cw-govmod-single/src/proposal.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{Addr, BlockInfo, CosmosMsg, Empty, StdResult, Storage, Uint128};
+use cosmwasm_std::{Addr, BlockInfo, CosmosMsg, Decimal, Empty, StdResult, Storage, Uint128};
 use cw_utils::Expiration;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


contracts/cw-govmod-single/src/proposal.rs
```diff
@@ -130,10 +114,115 @@ impl Proposal {
...
+                    }
+                }
+
+                // We need to invert the threshold, when nos exceed
+                // this inverted threshold we know the yes can never
+                // reach the threshold.
+                votes_meet_threshold(self.votes.no, options, Decimal::one() - percentage_needed)
+            }
+            Threshold::ThresholdQuorum { threshold, quorum } => {
+                match (
```
ezekiiel, line 161: @Callum-A you might consider taking a look at the logic here. I've made some modifications to your changes to handle the case where all votes are abstain or there is a 100% passing threshold. You might consider moving this behavior over to the other PR.

ezekiiel: I've also switched to using a match statement and checking conditions accordingly. This seems more clear as it makes explicit that we cover all four cases of expiration / quorum being met.

Callum-A (approved):
Yeah this all looks good now, cheers Zeke. Here's to electronic voting.

PR #224, opened 2022-03-25
Add a majority passing threshold option.

ezekiiel:
The majority threshold option allows the creator to specify that they
would like a proposal to pass if the majority of votes for that
proposal are yes. Otherwise, a proposal will pass if the yes votes are
greater than or equal to the number of votes required for passing.

Why this is needed:

If the passing threshold is 60% folks likely expect that if 6 out of
10 members vote yes that proposal will pass. This implies that yes >=
threshold.

On the other hand, this makes it difficult to specify that users would
like a majority of votes to be the passing threshold as specifying a
50% passing threshold will allow a vote with 5 yes and 5 no votes to
pass. The majority option addresses this by comparing yes > threshold
for that case.

This PR also adds a special case so if all the votes for a proposal are
abstain the proposal fails.


Callum-A (approved):
This architecture is so clean, this all looks great!

PR #226, opened 2022-03-26
Initial items on DAO Instantiation

NoahSaso:
Part 2 of #214

Allow the DAO to set an initial list of items on instantiation. These items can be:
- already existing addresses
- info used to instantiate a new contract, storing the corresponding address automatically

### DIFF
--- a/contracts/cw-governance/src/contract.rs
+++ b/contracts/cw-governance/src/contract.rs
@@ -71,9 +71,8 @@ pub fn instantiate(
 
     GOVERNANCE_MODULE_COUNT.save(deps.storage, &(gov_module_msgs.len() as u64))?;
 
-    let mut response = Response::new();
-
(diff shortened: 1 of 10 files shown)


janitachalam (approved):

contracts/cw-governance/src/contract.rs
```diff
@@ -59,7 +71,48 @@ pub fn instantiate(
 
     GOVERNANCE_MODULE_COUNT.save(deps.storage, &(gov_module_msgs.len() as u64))?;
 
-    Ok(Response::new()
+    let mut response = Response::new();
```
ezekiiel, line 74: It may be more readable to make a `Vec<Submsg<Empty>>` here instead of adding these to the response directly. You can then call `add_submessages ` later like we do with the `gov_module_msgs`.

PR #227, opened 2022-03-28
Add cw-governance functionality for managing cw20 and cw721 assets.

ezekiiel:
Adds:
    
1. Management of cw20 and cw721 token lists.
2. Automatic addition of received cw20 and cw721 assets to list.
3. Toggling on / off of automatic addition.
4. Associated queries for listing assets and listing asset balances.

After some thinking I've opted to implement this functionality inside
of the core contract instead of inside separate treasury
modules. Implementing treasury modules is nice because it makes us
flexible in the event that a new token / NFT / unknown asset standard
comes out and we would like to support it. It adds a fair amount of
complexity though:

1. Someone wanting to send tokens to a DAO must send those tokens to
   the appropriate treasury contract.
2. Queries on the frontend for asset lists require two additional
   queries (what are treasury contracts, what is address of treasury
   contract I care about).

We can always add new treasury contracts later down the line.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,6 +192,8 @@ dependencies = [
  "cw20 0.12.1",
  "cw20-balance-voting",
  "cw20-base 0.12.1",
+ "cw721",
+ "cw721-base",
  "schemars",
  "serde",
  "thiserror",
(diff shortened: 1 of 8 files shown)


Callum-A (approved):
Super nice work, I'm happy to get this in! Awesome

PR #231, opened 2022-04-04
Few improvements to named-groups PR

orkunkl:
This PR tries to implement address grouping without hash map. In its current state it does not work, there is a problem with moved values. Should not be difficult to fix

### DIFF
--- a/contracts/cw-named-groups/src/contract.rs
+++ b/contracts/cw-named-groups/src/contract.rs
@@ -1,5 +1,3 @@
-use std::collections::HashMap;
-
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
@@ -12,6 +10,8 @@ use crate::msg::{
     DumpResponse, ExecuteMsg, Group, InstantiateMsg, IsAddressInGroupResponse,
     ListAddressesResponse, ListGroupsResponse, QueryMsg,
 };
+
+use crate::ContractError::Std;
 use crate::state::{GROUPS, OWNER};
 
 // version info for migration info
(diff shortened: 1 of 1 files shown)


contracts/cw-named-groups/src/contract.rs
```diff
@@ -193,36 +193,26 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+    let (latest_name, _) = groups.first().ok_or(StdError::generic_err("not found"))?;
+
+    let mut dump: Vec<Group> = vec![];
+    let mut tmp: Vec<String> = vec![];
+    for (name, addr) in groups {
+        if name == *latest_name {
+            tmp.push(addr.to_string());
+        }
+        else {
+            dump.push(Group{ name, addresses: tmp })
```
ezekiiel, line 211: Given a set of groups of `{('water', 'fish'), ('water', 'plankton'), ('land', 'dog')}` it's not clear to me this works.

1. Set `latest_name = 'water'`.
2. While `latest_name == name` push to tmp => tmp becomes `['plankton', 'dog']`.
3. `latest_name != 'land'` so we add `Group { name: 'land', addresses: ['plankton', 'dog'] }`

Am I misunderstanding this logic? Haven't looked much at this contract so could be missing something.

ezekiiel: Think likely what we want to do here is
```rust
dump.push(Group{ name: latest_name, addresses: tmp });
latest_name = name;
tmp = vec![addr.to_string()]
```

janitachalam: I think ekez's proposed logic is correct. 

If the Map structure supported some kind of GROUP BY operation for composite keys where we could collapse entries by a prefix, that would be cool, but perhaps too niche of an operation. 

But also I'm curious, why is it so important to avoid using a hash map here since we're not persisting it to storage? How much of an optimization gain would it be?

orkunkl: HashMap builds runs the keys through a hashing algorithm. Hashing is somewhat an expensive operation for this desired logic and can be easily done by pushing data to vectors.

orkunkl:
Implemented on other branch, closing.

PR #238, opened 2022-04-13
First version of proposal hooks

Callum-A:
Pulling this up to start to get some feedback on my first implementation.

Tried to make it as portable as possible for the different proposal modules we will want to support.

Feedback appreciated!


contracts/cw-proposal-single/src/contract.rs
```diff
@@ -7,9 +7,11 @@ use cosmwasm_std::{
 use cw2::set_contract_version;
 use cw_storage_plus::Bound;
 use cw_utils::{Duration, Expiration};
+use proposal_hooks::{new_proposal_hooks, new_vote_hooks, proposal_status_changed_hooks};
 
 use voting::{Vote, Votes};
 
+use crate::state::HOOKS;
 use crate::{
     error::ContractError,
```
ezekiiel, line 16: This is total bikeshedding but we can combine these lol
```suggestion
use crate::{
    state::HOOKS,
    error::ContractError,
```

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -338,6 +369,52 @@ pub fn execute_update_config(
...
+
+    let validated_address = deps.api.addr_validate(&address)?;
+
+    HOOKS
+        .add_hook(deps.storage, validated_address.clone())
+        .map_err(ContractError::HookError)?;
+
+    Ok(Response::default()
+        .add_attribute("action", "add_hook")
+        .add_attribute("address", validated_address))
```
ezekiiel, line 392: We could lose the clone here by using the initial value. Ignore this if you think it makes the code less readable.

```suggestion
        .add_hook(deps.storage, validated_address)
        .map_err(ContractError::HookError)?;

    Ok(Response::default()
        .add_attribute("action", "add_hook")
        .add_attribute("address", address))
```

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -338,6 +369,52 @@ pub fn execute_update_config(
...
+    let config = CONFIG.load(deps.storage)?;
+    if config.dao != info.sender {
+        // Only DAO can remove hooks
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let validated_address = deps.api.addr_validate(&address)?;
+
+    HOOKS
+        .remove_hook(deps.storage, validated_address.clone())
```
ezekiiel, line 410: Same as above about losing the clone by using `address` below. Please ignore if you feel the code is less readable that way.

ezekiiel (commented):
Great work! Love how you broke this into a seprate package. Will need to add hooks into `execute_execute` as well here and then I think it will be well covered.

We should also think about the granularity at which we should allow subscribing to hooks. In this current model you subscribe to all hooks, though I could imagine a world where one only wants so subscribe to vote hooks to do incentivized voting.

Should we support more granular subscription like that? @JakeHartnell any thoughts?

JakeHartnell:
> Great work! Love how you broke this into a seprate package. Will need to add hooks into `execute_execute` as well here and then I think it will be well covered.
> 
> We should also think about the granularity at which we should allow subscribing to hooks. In this current model you subscribe to all hooks, though I could imagine a world where one only wants so subscribe to vote hooks to do incentivized voting.
> 
> Should we support more granular subscription like that? @JakeHartnell any thoughts?

More granular subscription would be _really_ nice. If only because of frequency and gas usage, the vote hook fires a lot, but `ProposalStatusChange` doesn't happen as often.

JakeHartnell (approved):
:fire:

Callum-A:
Circling with @ezekiiel 

- We want to remove bad hooks on failure so change to use reply_on_fail and `id = idx * 2 + 1` and `id = idx * 2` for the IDs. Can then work backwards based on odd or even
- Add the reply route and clean up
- May need our own hooks package to achieve this!

JakeHartnell (approved):

JakeHartnell (approved):
LFG

contracts/cw-proposal-single/src/msg.rs
```diff
@@ -109,6 +109,18 @@ pub enum ExecuteMsg {
...
+    },
+    RemoveProposalHook {
+        address: String,
+    },
+    AddVoteHook {
+        address: String,
+    },
+    RemoveVoteHook {
+        address: String,
+    },
```
ezekiiel, line 123: Think we need to generate the schema for these new message types.

ezekiiel (commented):
Amazing work @Callum-A. This is really, really great.

Last things are just to generate some schemas and update a README then lets merge this and test gas limits.

ezekiiel (approved):
:shipit:

JakeHartnell (approved):

PR #241, opened 2022-04-13
Remove division when computing if majority threshold is reached.

ezekiiel:
For the majority threshold instead of checking if `votes > threshold /
2`
we can instead check if `votes * 2 > threshold`. This removes a
division and means that we don't have to worry about a round down in
the division reporting a majority before one has actually been
obtained.

This also removes the `votes_needed` function which is no longer used
and adds a number of new tests to make sure this logic is working as
expected.


JakeHartnell (approved):
Clean.

Callum-A (approved):

PR #247, opened 2022-04-16
Active Check

Callum-A:
Allows voting modules to support an `IsActive {}` query route which can return true or false if the DAO is allowed to create proposals or not.

Feedback appreciated.


ezekiiel:
Rugged my the main merge. :( For my PR the easiest fix here was to just make a new branch and cherry-pick the commit I wanted.

PR #248, opened 2022-04-19
Update v1 contract schemas

JakeHartnell:
Updates v1 contract schemas so we can use the tool by @pyramation: https://github.com/pyramation/cosmwasm-typescript-gen

Note for this tool to work, we also need to export all responses returned by queries. For the AST to work, queries response types must follow the `QueryNameResponse` pattern. Had to do a bit of manipulation to get this to work for some types.

We may want to make sure all our queries follow this pattern. For examples, `ListProposals` is different from the `ProposalListResponse` imported from the `cw3` package (the response type doesn't follow the `QueryNameResponse` pattern).

Much of this has to do with the state of `cosmwasm-typescript-gen`, but figured it was worth mentioning.


Callum-A (approved):

ezekiiel:
This is great. Thank you!

PR #251, opened 2022-04-21
Run all voting tests against staking contract.

ezekiiel:
This creates a general purpose testing package in `packages/testing`
and uses that package to run all of our voting logic against our
staking contract as well as our `cw20_balances_voting` contract.

Followup work will add tests for our cw4 style voting as well.

From the perspective of the proposal module this sort of testing is
important because the method via which voting power is determined
should never impact the results of a vote.

In the process this PR also updates all of our cosmwasm
dependencies to their latest versions so that we can make sure to test
against the same code we'll be running.


ezekiiel:
Build error with conflicting versions of cosmwasm dependencies. Will follow up with a commit for that soon.

ezekiiel:
Build now failing for weird reasons only in `cargo wasm` and not in the docker container. Will investigate tomorrow. :)

janitachalam (approved):

ezekiiel:
Closing this in favor of another PR as the git conflicts here aren't worth the time.

PR #252, opened 2022-04-21
Remove latest field from Proposal

ezekiiel:
The `latest` field previously allowed a proposal to be created that
expires before the default proposal voting period. When a DAO has a
quorum this makes the hostile takeover cost`quorum * threshold * token_price`
as a proposer could create a proposal that instantly
expires and vote on it in one block.


janitachalam (approved):
nice one

PR #254, opened 2022-04-21
Limit the size of proposals.

ezekiiel:
The Juno mainnet allows storage writes in execute messages that are
larger than the amount of data queryable in a query message. This
means that a sufficiently large proposal (around 40,000 bytes) will
be unqueriable after being created.

To prevent this, this commit limits the size, in bytes, that a proposal
may be. A sensible limit for this size was found by repeatedly storing
a contract on the mainnet and creating the largest possible proposal.


Callum-A (approved):

JakeHartnell (approved):

bmorphism:
👀

Issue #246, opened 2022-04-15
Optional active_query macro for voting power contracts


Callum-A:
This macro will allow an interface to return true or false on whether or not they are "active" AKA can the DAO create proposals. This can be used to prevent front running.

PR #257, opened 2022-04-22
Active query

Callum-A:
Closes https://github.com/DA0-DA0/dao-contracts/issues/246
Allows voting modules to support an IsActive {} query route which can return true or false if the DAO is allowed to create proposals or not.

Feedback appreciated.

### DIFF
--- a/contracts/cw20-staked-balance-voting/src/contract.rs
+++ b/contracts/cw20-staked-balance-voting/src/contract.rs
@@ -1,16 +1,20 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdResult, SubMsg,
-    Uint128, WasmMsg,
+    to_binary, Binary, Decimal, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdResult,
+    SubMsg, Uint128, Uint256, WasmMsg,
 };
 use cw2::set_contract_version;
 use cw20::{Cw20Coin, TokenInfoResponse};
 use cw_core_interface::voting::IsActiveResponse;
 use cw_utils::parse_reply_instantiate_data;
+use std::convert::TryInto;
 
 use crate::error::ContractError;
-use crate::msg::{ActiveThreshold, ExecuteMsg, InstantiateMsg, QueryMsg, StakingInfo, TokenInfo};
+use crate::msg::{
+    ActiveThreshold, ActiveThresholdResponse, ExecuteMsg, InstantiateMsg, QueryMsg, StakingInfo,
+    TokenInfo,
+};
 use crate::state::{
     ACTIVE_THRESHOLD, DAO, STAKING_CONTRACT, STAKING_CONTRACT_CODE_ID,
     STAKING_CONTRACT_UNSTAKING_DURATION, TOKEN,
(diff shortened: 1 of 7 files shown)


janitachalam:
in the description (or a linked issue), do you mind explaining what attack scenario this is preventing? (for posterity)

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -226,6 +232,36 @@ pub fn query_dao(deps: Deps) -> StdResult<Binary> {
...
 
+pub fn query_is_active(deps: Deps) -> StdResult<Binary> {
+    let threshold = ACTIVE_THRESHOLD.may_load(deps.storage)?;
+    match threshold {
+        Some(threshold) => {
+            let token_contract = TOKEN.load(deps.storage)?;
+            let staking_contract = STAKING_CONTRACT.load(deps.storage)?;
+            let total_potential_power: TokenInfoResponse = deps
+                .querier
+                .query_wasm_smart(token_contract, &cw20_base::msg::QueryMsg::TokenInfo {})?;
```
ezekiiel, line 243: Seems like we don't need this in the `AbsoluteCount` case. How about moving that into the match statement below?

Callum-A:
> in the description (or a linked issue), do you mind explaining what attack scenario this is preventing? (for posterity)

Added the issue is here: https://github.com/DA0-DA0/dao-contracts/issues/246

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -226,6 +232,36 @@ pub fn query_dao(deps: Deps) -> StdResult<Binary> {
     to_binary(&dao)
 }
 
+pub fn query_is_active(deps: Deps) -> StdResult<Binary> {
+    let threshold = ACTIVE_THRESHOLD.may_load(deps.storage)?;
+    match threshold {
+        Some(threshold) => {
```
ezekiiel, line 238: `if let Some(threshold) = threshold {} else {}` might be cleaner here.

contracts/cw20-staked-balance-voting/src/contract.rs
```diff
@@ -226,6 +232,36 @@ pub fn query_dao(deps: Deps) -> StdResult<Binary> {
...
+                deps.querier.query_wasm_smart(
+                    staking_contract,
+                    &stake_cw20::msg::QueryMsg::TotalStakedAtHeight { height: None },
+                )?;
+            match threshold {
+                ActiveThreshold::AbsoluteCount { count } => to_binary(&IsActiveResponse {
+                    active: actual_power.total >= count,
+                }),
+                ActiveThreshold::Percentage { percent } => {
+                    let count = percent * total_potential_power.total_supply;
```
ezekiiel, line 254: This will round down. Ex: `0.5 * 7 = 3`. In the voting logic at least we take care to round up when possible. What do you think about doing that here?

contracts/cw20-staked-balance-voting/src/msg.rs
```diff
@@ -52,6 +60,7 @@ pub enum ExecuteMsg {}
 
 #[voting_query]
 #[token_query]
+#[active_query]
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum QueryMsg {
```
ezekiiel, line 66: How would you feel about adding a query here for the actual threshold being used as well as if they module is active?

contracts/cw20-staked-balance-voting/src/msg.rs
```diff
@@ -41,9 +41,17 @@ pub enum TokenInfo {
...
+    Percentage { percent: Decimal },
+}
+
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct InstantiateMsg {
     pub token_info: TokenInfo,
+    pub active_threshold: Option<ActiveThreshold>,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
```
ezekiiel, line 57: Could we add a message here to let the DAO update the active threshold?

ezekiiel (commented):
This is great! Nice work. Just some questions about rounding / additional queries / updating the active threshold.

ezekiiel (approved):
Great work! Thanks.

JakeHartnell (approved):

PR #258, opened 2022-04-22
introduce "random shuffle" voting

janitachalam:
This is more of a thought experiment and proof-of-concept contract for the new voting module / hooks design, and is intended to be used in order to play with chaotic power dynamics in a DAO.

Here's how it works: 

- People stake their governance tokens into the pot. There will be a fixed staking amount. Staked tokens do not correlate with voting power, but you must stake because we need there to be some token cost to participating in voting. 
- Once someone stakes or unstakes their tokens, the voting power in the DAO will get randomly re-shuffled. The voting power for any active proposals does not change, however.
- When a proposal is created, only those who have staked their tokens into the pot thus far can vote. This is because it messes up the passing calculations otherwise (thresholds are determined at the beginning of a proposal), and this allows the proposer to be aware of the set of voters they are proposing to. Similar reasoning to why we require staking before voting in the normal voting scenario.

Thoughts on attack scenarios are welcome..


contracts/cw-shuffle-voting/schema/count_response.json
```diff
@@ -0,0 +1,14 @@
+{
```
ezekiiel, line 1: Likely want to delete this file.

contracts/cw-shuffle-voting/schema/state.json
```diff
@@ -0,0 +1,24 @@
+{
```
ezekiiel, line 1: Also likely want to delete this one. :)

contracts/cw-shuffle-voting/src/contract.rs
```diff
@@ -0,0 +1,362 @@
...
+    addr_to_add: Option<Addr>,
+    addr_to_remove: Option<Addr>,
+) -> StdResult<()> {
+    // Generate voting weights based on total staked amount and number of participating addresses.
+    let weights = generate_random_voting_weights(env.block.height, addresses_count);
+
+    // Remove unstaked address if exists
+    if let Some(addr) = addr_to_remove {
+        VOTE_WEIGHTS.remove(deps.storage, &addr, env.block.height)?;
+    }
```
ezekiiel, line 231: What happens here if someone partially unstakes? Ex: I have 50 tokens staked and unstake 25.

contracts/cw-shuffle-voting/src/contract.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    random_weights
+}
+
+// Total voting power will differ from total staked power, given we have generated random weights.
+// Returns the sum of voting power at the given height.
+pub fn query_total_power_at_height(deps: Deps, env: Env, height: Option<u64>) -> StdResult<Binary> {
+    let height = height.unwrap_or(env.block.height);
+    let sum = HEIGHT_TO_TOTAL_POWER.load(deps.storage, height)?;
```
ezekiiel, line 293: What happens if this is called at a block where there were no stake or unstake actions? Seems to me that in that case the key won't exist.

contracts/cw-shuffle-voting/src/state.rs
```diff
@@ -0,0 +1,21 @@
...
+pub const STAKING_CONTRACT_CODE_ID: Item<u64> = Item::new("staking_contract_code_id");
+
+pub const VOTE_WEIGHTS: SnapshotMap<&Addr, Uint128> = SnapshotMap::new(
+    "user_weights",
+    "user_weights__checkpoints",
+    "user_weights__changelog",
+    Strategy::EveryBlock,
+);
+
+pub const HEIGHT_TO_TOTAL_POWER: Map<u64, Uint128> = Map::new("height_to_total_power");
```
ezekiiel, line 19: I think we likely want to use a `SnapshotMap` here as well.

contracts/stake-cw20/src/hooks.rs
```diff
@@ -7,17 +7,30 @@ use serde::{Deserialize, Serialize};
...
+    Stake {
+        addr: Addr,
+        amount: Uint128,
+        staked_addresses_count: u64,
+    },
+    Unstake {
+        addr: Addr,
+        amount: Uint128,
+        staked_addresses_count: u64,
+    },
```
ezekiiel, line 19: If we can avoid it, it would be nice not to change this hook message.

contracts/cw-shuffle-voting/src/contract.rs
```diff
@@ -0,0 +1,362 @@
...
+// The downside to using block_height, of course, is that our random sequences will be predictable.
+// This may not be a big issue given this doesn't allow for more ways to take over the DAO, just to know in advance the
+// distribution of voting powers. In the future we can explore using an on-chain drand-based beacon as a seed for unpredictable randomness: https://github.com/confio/rand
+pub fn generate_random_voting_weights(block_height: u64, addresses_count: u64) -> Vec<Uint128> {
+    let mut rng = Rand32::new(block_height);
+    let mut random_weights: Vec<Uint128> = vec![];
+    let mut i = 0;
+
+    // Generate addresses_count # of random numbers
+    while i < addresses_count {
```
ezekiiel, line 279: This is going to run out of gas for a large enough number of stakers. Need to work out some way to fix that if we want to deploy this to mainnet. :)

contracts/cw-shuffle-voting/src/state.rs
Callum-A: couldnt it just be a snapshot item? not a map?

JakeHartnell:
Going to close this for now, but it's a cool experiment. Might be the basis of a good DAO game!

PR #260, opened 2022-04-23
NFT Voting Module

cypherape:
Add a simple NFT voting module which determines voting power based on the NFT token balance of specific addresses.


contracts/cw721-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,94 @@
...
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    match msg.token_info {
+        TokenInfo::Existing { address } => {
+            let address = deps.api.addr_validate(&address)?;
+            TOKEN.save(deps.storage, &address)?;
```
janitachalam, line 27: "TOKEN" is supposed to be the address of the NFT contract this voting module is going to use, right? we should name it accordingly

also need to handle the new NFT contract case

contracts/cw721-balance-voting/schema/count_response.json
```diff
@@ -0,0 +1,14 @@
+{
```
ezekiiel, line 1: Looks like we need to regenerate these schema files. :) Should be able to run `cargo schema` in the root for that.

ezekiiel (commented):
Thanks for the contribution! NFT voting is definitely something that would be great to have.

For this contract we do need staking to prevent double votes. For example:

1. I have 2 NFTs.
2. I vote on a proposal with 2 voting power.
3. I send my NFTs to another wallet.
4. That wallet votes on a proposal with 2 voting power.

Now we've double voted. I think the way forward here would be to write a modified `stake-cw20` contract which lets you stake NFTs. Then, we could modify this contract to query that contract so that it can have voting power at height information and prevent that double voting.

contracts/cw721-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub fn query_token_contract(deps: Deps) -> StdResult<Binary> {
+    let token = TOKEN.load(deps.storage)?;
+    to_binary(&token)
+}
+
+pub fn query_voting_power_at_height(deps: Deps, env: Env, address: String) -> StdResult<Binary> {
+    let token = TOKEN.load(deps.storage)?;
+    let address = deps.api.addr_validate(&address)?;
+    let balance: cw721::TokensResponse = deps.querier.query_wasm_smart(
```
janitachalam, line 66: I think we'd also need some sort of staking contract for cw721s. stake-cw20 could be used for inspiration :) The reason we require staking before voting in the cw20 case is to ensure that someone can't vote then send their tokens to another wallet and vote again, and also so that the proposer knows who has "registered to vote" for their proposal. you can thinking of staking as an alias for "registering your tokens / NFTs to vote"

JakeHartnell:
> Thanks for the contribution! NFT voting is definitely something that would be great to have.
> 
> For this contract we do need staking to prevent double votes. For example:
> 
> 1. I have 2 NFTs.
> 2. I vote on a proposal with 2 voting power.
> 3. I send my NFTs to another wallet.
> 4. That wallet votes on a proposal with 2 voting power.
> 
> Now we've double voted. I think the way forward here would be to write a modified `stake-cw20` contract which lets you stake NFTs. Then, we could modify this contract to query that contract so that it can have voting power at height information and prevent that double voting.

So all of this is true and we now have #309!

However, this contract would work GREAT with non-transferable NFTs. Have a ticket to implement "soul bound" NFTs which we can use for things like titles. See: https://github.com/DA0-DA0/dao-contracts/issues/311

tl;dr this contract might still be useful.

contracts/cw721-balance-voting/src/contract.rs
JakeHartnell: This is implemented as a separate contract in #309

ezekiiel:
Closing as I've merged the voting power queries into the staking contract after some discussion with @Callum-A .

PR #261, opened 2022-04-23
Add IsActive check to proposal single

### LINKED ISSUE #160: Minimum staked percentage of total supply for the DAO to be active
@ezekiiel, think this would solve the race issue.

Callum-A:
Closes #160 
Adds the IsActive check for proposal single. 

The IsActive query route is optional so on failure it assumes a DAO is active.

Is this safe? Maybe it should be mandatory so on error we can assume inactive? Thoughts?

Feedback appreciated


ezekiiel (approved):
Great work!

janitachalam (approved):

PR #264, opened 2022-04-25
Rename governance modules -> proposal modules in cw-core messages.

ezekiiel:
Noticed that we had missed some renaming when I was instantiating the latest contracts.


JakeHartnell (approved):
:heart:

Callum-A (approved):

PR #265, opened 2022-04-26
Allow the core contract to pause itself.

ezekiiel:
Stops literally all execute messages from happening for the duration of the pause.


JakeHartnell (approved):
Looks good! Should there be a way to un-pause the DAO?

ezekiiel:
>  @JakeHartnell - Should there be a way to un-pause the DAO?

I guess I'm imagining a world where this pause method is effectively a "stop the world until we can figure out what's next". I could see this being used in a situation where there is a vulnerability discovered in the core contract or a governance contract when we don't yet have a replacement.

Individual modules might have more granular methods for pausing themselves (for example, a voting module that can be paused on weekends). If you're pausing the core contract that's the big red something has gone horribly wrong and we need to stop everything button. In a perfect world, this should probably never be used.

janitachalam (approved):

PR #271, opened 2022-04-27
Add permissions and block after test to cw4-groups contract.

ezekiiel:
- Tests that new weights are not reflected until the block increasing the weight is done.
- Tests that permissions are properly managed on the member update hook.


JakeHartnell (approved):
Loving these extra tests! :heart:

ezekiiel:
> Loving these extra tests! ❤️

More coming. :)

Callum-A (approved):

PR #278, opened 2022-04-28
Vt rewards distributor

verminator23:


contracts/cw-distributor/src/contract.rs
```diff
@@ -0,0 +1,130 @@
...
+    env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let owner = deps.api.addr_validate(&msg.owner)?;
+    let recipient = deps.api.addr_validate(&msg.recipient)?;
+
+    let reward_token = deps.api.addr_validate(&msg.reward_token)?;
```
ezekiiel, line 28: Might be nice to run some queries against these addresses when we set them to verify that these are indeed token / staking contracts. Could run `TotalStakedAtHeight` and `TokenInfo`.

contracts/cw-distributor/src/contract.rs
```diff
@@ -0,0 +1,130 @@
...
+    reward_token: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if config.owner != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let owner = deps.api.addr_validate(&owner)?;
+    let recipient = deps.api.addr_validate(&recipient)?;
+    let reward_token = deps.api.addr_validate(&reward_token)?;
```
ezekiiel, line 76: Queries here as well if we decide we'd like to run them.

contracts/cw-distributor/src/contract.rs
```diff
@@ -0,0 +1,130 @@
...
+        owner,
+        recipient,
+        reward_token,
+        reward_rate,
+    };
+    CONFIG.save(deps.storage, &config)?;
+    Ok(Response::default())
+}
+
+pub fn execute_distribute(deps: DepsMut, env: Env) -> Result<Response, ContractError> {
```
ezekiiel, line 88: This function will lock tokens in the contract is `config.reward_rate * block_diff` is greater than the contract's token balance. This is likely to happen when the contract is coming to the end of its reward period. How would you feel about making the amount to send the min of remaining balance and `amount`?

ezekiiel: Likely want a test for this situation as well. :)

ezekiiel (commented):
Looks great! Just one thing around locking the contract at the end of the rewards period and then I think this is good to go.

contracts/cw-distributor/src/contract.rs
verminator23: Great suggestion. I'll add that

contracts/cw-distributor/src/contract.rs
verminator23: @ben2x4 Lmk if you think this type of validation is necessary. I'm down to add it

contracts/cw-distributor/src/contract.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use cosmwasm_std::{Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
+use cw2::set_contract_version;
+use stake_cw20::msg::ReceiveMsg;
+
+use crate::error::ContractError;
+use crate::msg::{ConfigResponse, ExecuteMsg, InstantiateMsg, LastPaymentBlockResponse, QueryMsg};
+use crate::state::{Config, CONFIG, LAST_PAYMENT_BLOCK};
+
+const CONTRACT_NAME: &str = "crates.io:cw-distributor";
```
ben2x4, line 13: I would probable change name here and for file to `stake-cw20-reward-distributor` as I don't think the contract is very general purpose.

contracts/cw-distributor/src/contract.rs
ben2x4: yup, had same thought and +1 here

contracts/cw-distributor/src/contract.rs
```diff
@@ -0,0 +1,130 @@
...
+        contract_addr: config.reward_token.into(),
+        msg,
+        funds: vec![],
+    }
+    .into();
+    Ok(Response::default().add_message(send_msg))
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
```
ben2x4, line 115: I would consider combining all of this into one `info` response and including the contracts current balance of the reward token. Having reward token balance makes it easy for the caller to see if the rewards are active and calculate the remaining length without making a seperate query.

ben2x4 (commented):
This is great, I think the minimal amount of state you have included here is really elegant. A couple of small suggestions and then ready to go!

contracts/cw-distributor/src/contract.rs
ben2x4: hmm, I know @ezekiiel  likes these types of checks, but i don't have a strong preference.

contracts/cw-distributor/src/contract.rs
ben2x4: this also make it clear it is part of the stake-cw20 family of contracts

contracts/cw-distributor/src/contract.rs
verminator23: Done

contracts/cw-distributor/src/contract.rs
verminator23: Done

contracts/cw-distributor/src/contract.rs
verminator23: Done

contracts/cw-distributor/src/contract.rs
verminator23: Added these checks + tests

verminator23:
@ezekiiel @ben2x4 I've addressed all of your comments

ben2x4 (approved):
LGTM, amazing job here!

ezekiiel (approved):
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

JakeHartnell:
LFG!

PR #279, opened 2022-05-02
multiple choice proposals v1

janitachalam:
Adds the cw-proposal-multiple contract with single-choice voting ([first-past-the-post](https://www.electoral-reform.org.uk/voting-systems/types-of-voting-system/first-past-the-post/)). Ranked choice to come 

is_passed and is_rejected need the closest review: https://github.com/DA0-DA0/dao-contracts/blob/8552506f9d60489e7e8c3917fddca22d1105236a/contracts/cw-proposal-multiple/src/proposal.rs

Voting strategy design choices: 

- get rid of the percentage option entirely for multiple choice proposals. (it is difficult to reason about a fixed percentage threshold when proposals can have a variable number of options)
- There will always be a quorum for voter turnout (since we are removing the percentage option, it makes sense to have a min bar for turnout after which the highest voted choice will be selected)
- There will be two voting strategies, either single-choice (first-past-the-post, ie the option with the highest number of votes gets selected), or ranked-choice (single-transferable vote, ie multiple choices can be ranked, and votes will cascade to lower preferences until an option receives the majority of votes)
- We will include the abstain / "None of the Above" option — like in single choice proposals, it will contribute towards quorum and will also allow proposals with only bad options to be voted against.
- The voting strategy for a proposal module will be configurable by config, just like how they are for thresholds currently


contracts/cw-proposal-multiple/check_wasm.sh
```diff
@@ -0,0 +1,5 @@
+#!bin/bash
+
+RUSTFLAGS='-C link-arg=-s' cargo wasm
+cp ../../target/wasm32-unknown-unknown/release/cw3_multiple_choice.wasm .
+check_contract cw3_multiple_choice.wasm
```
ezekiiel, line 5: Where is this `check_contract` function defined?

contracts/cw-proposal-multiple/msg.rs
```diff
@@ -0,0 +1,14 @@
+use cosmwasm_std::{CosmosMsg, Empty};
+use cw_utils::Expiration;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct MultipleChoiceProposeMsg {
```
ezekiiel, line 7: Any reason to define this here instead of in `src/msg.rs`?

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,672 @@
...
+        sender.clone(),
+        config.dao.clone(),
+        Some(env.block.height),
+    )?;
+    if sender_power.is_zero() {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // Set the expiration to the minimum of the proposal's `latest`
+    // argument and the configured max voting period.
```
ezekiiel, line 136:
```suggestion
```

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,672 @@
...
+    let expiration = config.max_voting_period.after(&env.block);
+
+    let total_power = get_total_power(deps.as_ref(), config.dao, Some(env.block.height))?;
```
ezekiiel, line 139:
```suggestion
    let total_power = get_total_power(deps.as_ref(), config.dao, None)?;
```

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,672 @@
...
+    info: MessageInfo,
+    proposal_id: u64,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if config.only_members_execute {
+        let power = get_voting_power(
+            deps.as_ref(),
+            info.sender.clone(),
+            config.dao.clone(),
+            Some(env.block.height),
```
ezekiiel, line 276:
```suggestion
            None,
```

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,672 @@
...
+    info: MessageInfo,
+    proposal_id: u64,
+) -> Result<Response<Empty>, ContractError> {
+    let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
+
+    // Only open and expired or rejected proposals may be closed.
+    match prop.status {
+        Status::Rejected => (),
+        Status::Open => {
+            if !prop.expiration.is_expired(&env.block) {
```
ezekiiel, line 341:
```suggestion
            if !prop.expiration.is_expired(&env.block) || prop.is_passed(&env.block) {
```

Can't close a passed proposal.

contracts/cw-proposal-multiple/src/proposal.rs
```diff
@@ -0,0 +1,177 @@
...
+        ) {
+            let (choice_idx, winning_choice) = self.calculate_winning_choice();
+            // Check that the winning choice is not None.
+            if winning_choice.option_type != MultipleChoiceOptionType::None {
+                // If the leading choice cannot possibly be outwon by the "None" option, the proposal has passed.
+                // This means that the difference between them must be greater than the remaining vote power.
+                let none_count = u128::from(self.get_none_vote_count());
+                let winning_choice_count = self.votes.vote_weights[choice_idx].u128();
+                let remaining_power = self.total_power.sub(self.votes.total()).u128();
+                if winning_choice_count - none_count > remaining_power {
```
ezekiiel, line 102:
```
total_power - 7
foo - 3
bar - 2
none - 1
```

Here `winning_choice_count - none_count > remaining_power` but it is still possible for bar to get the next vote and result in a tie. Is this where the 'first across the line' logic comes in?

contracts/cw-proposal-multiple/src/proposal.rs
```diff
@@ -0,0 +1,177 @@
...
+                    return self.votes.vote_weights.iter().all(|x| *x == *first);
+                }
+                false
+            }
+            // Quorum is met and proposal has not expired OR Quorum is not met and proposal is not expired.
+            (true, false) | (false, false) => {
+                // Proposal is rejected if "None" has the majority of the total power because there is no way for
+                // another option to outnumber it.
+                let none_vote_count = Uint128::from(self.get_none_vote_count());
+                does_vote_count_fail(
```
ezekiiel, line 139: Should this be `does_vote_count_pass`? We're checking if `None` has passed the majority threshold right?

contracts/cw-proposal-multiple/src/proposal.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    pub fn get_none_vote_count(&self) -> u64 {
+        return self
+            .votes
+            .vote_weights
+            .iter()
+            .enumerate()
+            .filter(|(idx, _x)| self.choices[*idx].option_type == MultipleChoiceOptionType::None)
+            .count() as u64;
```
ezekiiel, line 175: This is the number of folks who have voted for `None` but not the number of votes for `None` as it doesn't count weights here. Think this should sum the weights of votes for `None`.

contracts/cw-proposal-multiple/src/query.rs
```diff
@@ -0,0 +1,132 @@
...
+/// This defines the different ways tallies can happen.
+/// Every contract should support a subset of these, ideally all.
+///
+/// The total_weight used for calculating success as well as the weights of each
+/// individual voter used in tallying should be snapshotted at the beginning of
+/// the block at which the proposal starts (this is likely the responsibility of a
+/// correct cw4 implementation).
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum ThresholdResponse {
```
ezekiiel, line 51: Seems like we only support quorum here. Any reason to have this entire enum?

contracts/cw-proposal-multiple/src/state.rs
```diff
@@ -0,0 +1,89 @@
...
+
+pub fn parse_id(data: &[u8]) -> StdResult<u64> {
+    match data[0..8].try_into() {
+        Ok(bytes) => Ok(u64::from_be_bytes(bytes)),
+        Err(_) => Err(StdError::generic_err(
+            "Corrupted data found. 8 byte expected.",
+        )),
+    }
+}
+
```
ezekiiel, line 67:
```suggestion
```

legacy/cw3-dao/src/contract.rs
```diff
@@ -217,14 +217,6 @@ pub fn execute_propose(
...
     latest: Option<Expiration>,
 ) -> Result<Response<Empty>, ContractError> {
-    // Check if DAO is Paused
-    let paused = DAO_PAUSED.may_load(deps.storage)?;
-    if let Some(expiration) = paused {
-        if !expiration.is_expired(&env.block) {
-            return Err(ContractError::Paused {});
-        }
-    }
-
```
ezekiiel, line 227: Why are we removing these?

packages/voting/src/deposit.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+pub fn get_return_deposit_msg<T: Proposal>(proposal: &T) -> StdResult<Vec<CosmosMsg>> {
+    match &proposal.deposit_info() {
+        Some(info) => {
+            // If it is closed only issue a refund if we are refunding
+            // failed proposals. Otherwise, the proposal is open and
+            // expired.
+            if proposal.status() == Status::Rejected && !info.refund_failed_proposals {
```
ezekiiel, line 116:
```suggestion
            if proposal.status() == Status::Rejected && !info.refund_failed_proposals || info.deposit.is_zero() {
```

Otherwise this will error with a zero proposal deposit. This is my bug, not yours. May as well fix it here. :)

packages/voting/src/voting.rs
```diff
@@ -26,6 +29,43 @@ pub enum Vote {
     Abstain,
 }
 
+#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, JsonSchema, Debug)]
+pub struct MultipleChoiceVote {
+    // A vote indicates which option the user has selected.
+    pub option_id: u32,
```
ezekiiel, line 35: Think we want this to be a `u64` just to be pedantic so that we are sure it'll be able to index the vec of options without overflowing.

ezekiiel (commented):
Great work @janitachalam! Left some initial feedback. Once that's addresseed it might be nice to get on a call and walk through the vote passing logic. At the very least I'd like to make another pass reading that before we merge this.

contracts/cw-proposal-multiple/check_wasm.sh
janitachalam: sorry, this was for testing

contracts/cw-proposal-multiple/src/query.rs
janitachalam: artifact of older version, removed

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,672 @@
...
+        .may_load(deps.storage, proposal_id)?
+        .ok_or(ContractError::NoSuchProposal { id: proposal_id })?;
+
+    // Check here that the proposal is passed. Allow it to be
+    // executed even if it is expired so long as it passed during its
+    // voting period.
+    let old_status = prop.status;
+    if !prop.is_passed() {
+        return Err(ContractError::NotPassed {});
+    }
```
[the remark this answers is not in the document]
janitachalam: good check, I have changed it to update the current status and then check explicitly for the passed state. let me know what you think

contracts/cw-proposal-multiple/src/proposal.rs
janitachalam: hmm fair point the tie case needs to be handled as well

packages/voting/src/voting.rs
janitachalam: but if we have to cast it to usize and usize is u32, isn't that an issue?

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,671 @@
...
+        // Proposal hook so we can just divide by two for index
+        let idx = msg.id / 2;
+        PROPOSAL_HOOKS.remove_hook_by_index(deps.storage, idx)?;
+        Ok(Response::new())
+    } else {
+        // Vote hook so we can minus one then divid by two for index
+        let idx = (msg.id - 1) / 2;
+        VOTE_HOOKS.remove_hook_by_index(deps.storage, idx)?;
+        Ok(Response::new())
+    }
```
Callum-A, line 670: nit: can move `Ok(Response::new())` to outside if else

janitachalam:
improvements to be made after review with zeke: 

- multiplechoiceoption —> checked multiple choice option (or CheckedProposalOption)
    - wrapper type that adds the none option
    - will add index and vote count 
    - can move validate_choice s there 
    - add a constraint to the number of choices, eg 10 to prevent gas issues 

- calculate vote result: consider finding top choices in one pass 
- test for the overflow case in is_choice_unbeatable
- 
- later: remove option for non members execute? 
- 2 raw dao issues:
- after rebasing copy the update_status and passed checks into proposal_single 

// run a query to query voting modules for validating that this is a dao

contracts/cw-proposal-multiple/src/proposal.rs
```diff
@@ -0,0 +1,207 @@
...
+
+                    let winning_choice = top_choices.first().unwrap();
+                    return Ok(VoteResult::SingleWinner(
+                        self.choices[winning_choice.0].clone(),
+                    ));
+                }
+                Err(StdError::not_found("max vote weight"))
+            }
+
+            VotingStrategy::RankedChoice { quorum: _ } => todo!(),
```
Callum-A, line 184: is this known future work?

ezekiiel (approved):
This is incredible. Great work! Long time coming.

Only a few comments that aren't complements. Couple places I think we could avoid a heap allocations and one function design nit. No functionality issues that I can see though so approved!

Would be great to address those / maybe bump the test coverage a little so we don't regress (even though it's a tiny amount) before merging, but I trust you'll make a good call on what to do pre merge. Great work!

packages/testing/src/lib.rs
```diff
@@ -1,4 +1,6 @@
 #[cfg(not(target_arch = "wasm32"))]
+pub mod helpers;
+#[cfg(not(target_arch = "wasm32"))]
```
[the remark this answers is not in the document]
janitachalam: it doesn't compile otherwise since the testing package is imported in other modules

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,702 @@
...
+    proposal_id: u64,
+    vote: MultipleChoiceVote,
+) -> Result<Response<Empty>, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let mut prop = PROPOSALS
+        .may_load(deps.storage, proposal_id)?
+        .ok_or(ContractError::NoSuchProposal { id: proposal_id })?;
+
+    // Check that this is a valid vote.
+    if vote.option_id as usize >= prop.choices.len() {
```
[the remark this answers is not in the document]
janitachalam: yeah it compiles, I think using usize is only a problem in messages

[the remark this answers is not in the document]
Callum-A: usize works unless it is stored : ) e.g. in a struct in an item

contracts/cw-proposal-multiple/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
Callum-A, line 1: delete

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -0,0 +1,715 @@
...
+        .unwrap_or(IsActiveResponse { active: true });
+
+    if !active_resp.active {
+        return Err(ContractError::InactiveDao {});
+    }
+
+    // Check that the sender is a member of the governance contract.
+    let sender_power = get_voting_power(deps.as_ref(), sender.clone(), config.dao.clone(), None)?;
+    if sender_power.is_zero() {
+        return Err(ContractError::Unauthorized {});
```
Callum-A, line 153: nit: error could be improved here, could be replaced with a MustHaveVotingPowerToPropose error. verbose but does the job

contracts/cw-proposal-multiple/src/state.rs
```diff
@@ -0,0 +1,141 @@
...
+    // Workaround due to not being able to use HashMaps in Cosmwasm.
+    pub index: u32,
+    pub option_type: MultipleChoiceOptionType,
+    pub description: String,
+    pub msgs: Option<Vec<CosmosMsg<Empty>>>,
+    pub vote_count: Uint128,
+}
+
+impl MultipleChoiceOptions {
+    pub fn into_checked(self) -> Result<CheckedMultipleChoiceOptions, ContractError> {
```
Callum-A, line 85: is there tests for this function as an individual unit?

packages/testing/src/helpers.rs
```diff
@@ -0,0 +1,437 @@
+use cosmwasm_std::{to_binary, Addr, Binary, Empty, Uint128};
+use cw20::Cw20Coin;
+use cw20_staked_balance_voting::msg::ActiveThreshold;
+use cw_core::msg::ModuleInstantiateInfo;
+use cw_multi_test::{App, Contract, ContractWrapper, Executor};
+use cw_utils::Duration;
+
+const CREATOR_ADDR: &str = "creator";
+
+pub fn instantiate_with_cw20_balances_governance(
```
Callum-A, line 10: Awesome stuff here

Callum-A (approved):
Absolutely great work, a few nits and questions. Thanks for doing this, this is huge

contracts/cw-proposal-multiple/src/contract.rs
janitachalam: agreed!

contracts/cw-proposal-multiple/src/state.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct CheckedMultipleChoiceOption {
+    // This is the index of the option in both the vote_weights and proposal.choices vectors.
+    // Workaround due to not being able to use HashMaps in Cosmwasm.
+    pub index: u32,
+    pub option_type: MultipleChoiceOptionType,
+    pub description: String,
+    pub msgs: Option<Vec<CosmosMsg<Empty>>>,
```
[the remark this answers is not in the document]
janitachalam: hmm, I guess I have a slight preference for it being an Option to indicate either a text option or the "None of the Above" option which never has msgs

contracts/cw-proposal-multiple/src/state.rs
janitachalam: added!

janitachalam:
re: slight code coverage decrease, I've ensured that the all the critical code paths are test-covered and I think codecov is a little wonky with its calculations as the coverage still decreased even after excluding some test files from the calculation. Think it's good to merge.

PR #280, opened 2022-05-02
multiple choice proposals v1

janitachalam:


elsehow:
This is super awesome.

There are a few ways of dealing with multiple-choice voting. The simplest is [first-past-the-post (FPTP)](https://www.electoral-reform.org.uk/voting-systems/types-of-voting-system/single-transferable-vote/).

Perhaps this contract should be called `cw-proposal-fptp` to make it explicit what kind of counting the contract uses. Future proposals could implement singe transferrable vote (`cw-proposal-stv`), etc etc etc.

Callum-A:
> This is super awesome.
> 
> There are a few ways of dealing with multiple-choice voting. The simplest is [first-past-the-post (FPTP)](https://www.electoral-reform.org.uk/voting-systems/types-of-voting-system/single-transferable-vote/).
> 
> Perhaps this contract should be called `cw-proposal-fptp` to make it explicit what kind of counting the contract uses. Future proposals could implement single transferrable vote (`cw-proposal-stv`), etc etc etc.

I like this idea, it is time for more voting system discussions

PR #284, opened 2022-05-08
Names Registry

Callum-A:
Pulling this up so everyone can see my progress and discuss design choices


contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+) -> Result<Response, ContractError> {
+    let mut config = CONFIG.load(deps.storage)?;
+
+    if info.sender != config.admin {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let new_payment_token_address =
+        new_payment_token_address.unwrap_or_else(|| config.payment_token_address.to_string());
+    let payment_amount = new_payment_amount.unwrap_or(config.payment_amount);
```
elsehow, line 147: Looks like we're not checking whether payment is 0 here. That's fine - as long as it's consistent with `instantiate`.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn instantiate(
+    deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let payment_token_address = deps.api.addr_validate(&msg.payment_token_address)?;
```
elsehow, line 28: I never considered that we might want to set the payment token (I assumed it would be the native token of the chain), but this is definitely a good call I think.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    sender: String,
+    amount: Uint128,
+    name: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if amount < config.payment_amount {
+        return Err(ContractError::InsufficientFunds {});
+    }
+
+    // this is the DAO
```
elsehow, line 110: I think what you probably mean is, "We expect this to be the DAO that is hoping to register a name." The comment threw me for a loop for a second: "wait, is this the name registry DAO? I thought that was the `admin`?"

On that same note, a comment on the `admin` saying, "We expect this to be the Name Registry DAO" could be useful.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let payment_token_address = deps.api.addr_validate(&msg.payment_token_address)?;
+    let admin = deps.api.addr_validate(&msg.admin)?;
+
+    if msg.payment_amount.is_zero() {
```
elsehow, line 31: Perhaps `payment_amount` could have a slightly more descriptive name. What about `payment_required_to_register_name`?

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+
+pub fn register_name(
+    deps: DepsMut,
+    _env: Env,
+    sender: String,
+    amount: Uint128,
+    name: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if amount < config.payment_amount {
```
elsehow, line 106: See note above - `if amount < config.payment_required_to_register_name` is a bit more self-explanatory.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    if NAME_TO_DAO.has(deps.storage, name.clone()) {
+        return Err(ContractError::NameAlreadyTaken {});
+    }
+
+    if DAO_TO_NAME.has(deps.storage, sender.clone()) {
+        return Err(ContractError::NameAlreadyRegistered {});
+    }
+
+    NAME_TO_DAO.save(deps.storage, name.clone(), &sender)?;
+    DAO_TO_NAME.save(deps.storage, sender, &name)?;
```
elsehow, line 126: 🤌 perfect, super clear.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    if info.sender != config.admin {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let new_payment_token_address =
+        new_payment_token_address.unwrap_or_else(|| config.payment_token_address.to_string());
+    let payment_amount = new_payment_amount.unwrap_or(config.payment_amount);
+    let new_admin = new_admin.unwrap_or_else(|| config.admin.to_string());
+
```
elsehow, line 149: I'm certain you did this `Option`al update logic correctly _and_ I'd love to see tests that try to update each of these config fields individually, with the other options not set. The `unwrap_or_else` logic is just tricky + footgunny enough to be worth the extra assurance. (Could you imagine if faulty update logic here gets the contract into a bad state in prod? That would be highkey yikes.)

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    info: MessageInfo,
+    name: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let dao = NAME_TO_DAO.load(deps.storage, name.clone())?;
+
+    // Only name owner and overall name admin can revoke
+    if info.sender != dao && info.sender != config.admin {
+        return Err(ContractError::Unauthorized {});
+    }
```
elsehow, line 175: I might be dumb, but I'm confused here. Shouldn't this be `info.sender != dao || info.sender != config.admin`? You can revoke the name if you are the Naming DAO (admin) *or* if you're the owner of the name. Right? Am I missing something?

Nitpick: instead of "overall name admin" in the comment, I'd probably say something like, "admin of the contract (which we anticipate to be the Name Registry DAO)". Making our expectations explicit is always good.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+pub fn execute_revoke(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    name: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let dao = NAME_TO_DAO.load(deps.storage, name.clone())?;
```
elsehow, line 170: What happens if this name doesn't pertain to a DAO in the map? I assume we get an error, but what is that error?

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    _env: Env,
+    info: MessageInfo,
+    name: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    // Check if name is already taken
+    if NAME_TO_DAO.has(deps.storage, name.clone()) {
+        return Err(ContractError::NameAlreadyTaken {});
+    }
+
```
elsehow, line 194: Is it worth it to also check if the is in the `RESERVED_NAMES` list?

contracts/cw-names-registry/src/error.rs
```diff
@@ -0,0 +1,26 @@
...
+    NameReserved {},
+
+    #[error("This name is not reserved for later use")]
+    NameNotReserved {},
+
+    #[error("This name is already taken")]
+    NameAlreadyTaken {},
+
+    #[error("You already registered a name with this DAO")]
+    NameAlreadyRegistered {},
```
elsehow, line 25: `NameAlreadyRegistered` makes it sound like the name is already taken. Perhaps we could name this one `AddressAlreadyHasOneName`? It's cumbersome, but boneheadedly obvious. 


Some alternatives, in the interest of throwing spaghetti at the wall:

- `AlreadyRegisteredOneName`
- `AlreadyRegisteredName`
- `AlreadyHasName`

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    if RESERVED_NAMES.has(deps.storage, name.clone()) {
+        return Err(ContractError::NameReserved {});
+    }
+
+    if NAME_TO_DAO.has(deps.storage, name.clone()) {
+        return Err(ContractError::NameAlreadyTaken {});
+    }
+
+    if DAO_TO_NAME.has(deps.storage, sender.clone()) {
+        return Err(ContractError::NameAlreadyRegistered {});
```
elsehow, line 122: See comment below about the name of this error. I actually missed that we were doing a one-name-per-DAO check here because I didn't grok the difference between `NameAlreadyTaken` and `NameAlreadyRegistered`.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+}
+
+pub fn query_look_up_dao(deps: Deps, dao: String) -> StdResult<Binary> {
+    let dao = deps.api.addr_validate(&dao)?;
+    let name = DAO_TO_NAME.may_load(deps.storage, dao)?;
+    to_binary(&LookUpDaoResponse { name })
+}
+
+pub fn query_look_up_name(deps: Deps, name: String) -> StdResult<Binary> {
```
elsehow, line 252: I had to read this method a few times, and I'm still just guessing at what it does. That might be my lack of having my head in contracts. But I think a different name could go some way here.

My guess is that this contract tells you (1) whether a name is available or not and (2) if it is unavailable, to what DAO it refers.

Does this method give you a binary (whether or not this name is available)? If so, a more descriptive name might be `look_up_dao_by_name` (and `LookUpDaoByName`).

Similarly, `LookUpNameByDao`, `query_look_up_name_by_dao`.

contracts/cw-names-registry/src/msg.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct LookUpDaoResponse {
+    pub name: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct LookUpNameResponse {
```
elsehow, line 59: I think there may be two routes in here: (1) something I'd call `LookUpDaoByName`, which you might call when you're searching for a DAO, and (2) something I'd call `IsNameAvailableToRegister`.  I'd expect the former to return an Option`<Addr>` and the latter to return a `bool`. Is there a reason not to separate them out?

Now, maybe the answer is, "this way, you only need to do one query to see which DAO has a name + whether that name is available," which, fine, but will users ever need to know these things at the same time? I think people calling this contract (for now, the proposal template) will mostly want to know whether a DAO is available - and we can give them a really simple route to call to find out. 

Not going to bikeshed on this point though - it's almost certainly fine either way.

contracts/cw-names-registry/src/tests.rs
```diff
@@ -0,0 +1,477 @@
...
+        names.clone(),
+        Some(other_token.to_string()),
+        Some(NON_ADMIN_ADDR.to_string()),
+        Some(Uint128::new(25)),
+        Addr::unchecked(NON_ADMIN_ADDR),
+    )
+    .unwrap_err();
+
+    // Update config as admin
+    update_config(
```
elsehow, line 458: See comment above. I'm not sure why this makes me so nervous, but whenever there are lots of `Option<>`s and conditional updating, I get a bit sweaty. It would be good to sense check that these update correctly when we're only setting one field at a time.

contracts/cw-names-registry/src/tests.rs
```diff
@@ -0,0 +1,477 @@
...
+        &mut app,
+        names.clone(),
+        Uint128::new(50),
+        name.to_string(),
+        Addr::unchecked(DAO_ADDR),
+        other_token,
+    )
+    .unwrap_err();
+
+    // Send too little
```
elsehow, line 180: What happens when a DAO sends too much? What should happen when a DAO sends too much? I say: refund the excess. But perhaps that would open up some weird attack or waste gas fees via conditional behavior.

elsehow (changes requested):
This looks _great_! I've left a few minor requests for changes, mostly making names more self-explanatory, along with a few food-for-thought type judgment calls. (If you're not sure whether something is one or the other, let me know).

Some things that are outside of this contract's scope that we should have a group discussion about in a different forum:

- How will we decide which names to reserve?
- How will we receive requests to transfer the reservation? How will we verify those requests?

Looking forward to shipping this!

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,256 @@
...
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let payment_token_address = deps.api.addr_validate(&msg.payment_token_address)?;
+    let admin = deps.api.addr_validate(&msg.admin)?;
+
+    if msg.payment_amount.is_zero() {
+        return Err(ContractError::Unauthorized {});
+    }
```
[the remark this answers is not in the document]
Callum-A: So the reason I error on zero is that it uses the Receive hook for CW20 and you can't send zero so it's actually unreachable.

contracts/cw-names-registry/src/contract.rs
Callum-A: Yeah this uses cw20s, not natives, maybe we could do multiple payment methods?

contracts/cw-names-registry/src/contract.rs
Callum-A: Ah yep lol fair point

contracts/cw-names-registry/src/contract.rs
Callum-A: yeah will flesh out the test

contracts/cw-names-registry/src/contract.rs
Callum-A: yep you're right good spot

contracts/cw-names-registry/src/contract.rs
Callum-A: It'll be the inbuilt error for a non existent key in CW, I'll make an explicit error for it so it'll be more human readable

elsehow (approved):
Awesome, LGTM.

ezekiiel (commented):
Looks great! Just a couple little things. Thanks @Callum-A !

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+    wrapped: Cw20ReceiveMsg,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    // We only take payments from our specified token
+    if info.sender != config.payment_token_address {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // TODO: Is there a way we can verify this is a DAO
```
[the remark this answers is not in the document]
elsehow: Popping in to say that I'm unsure whether we _should_ verify that the registrant is a DAO. Memorializing something I said to @Callum-A yesterday: 

If we saw a bunch of degens snapping names, we would burn those names, shame those degens, and issue a patch as a DAO. In the meantime, why not support an entrepreneurial user snapping up a name for their DAO before they form said DAO? That sounds like something I would do.

Also, while requiring registrants be DAOs raises the barrier to name-snapping, a motivated spoofer will form a single-member DAO with the target name, set proposal length low, and register the name. The only real disincentive to gobbling names is the price we set, which the Name Registry DAO can adjust via governance proposal.

Since we have governance as a backstop, fewer permissions, and less complexity is IMO preferred.

ezekiiel (approved):
LG2M! Great work.

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,330 @@
...
+
+pub fn register_name(
+    deps: DepsMut,
+    _env: Env,
+    sender: String,
+    amount_sent: Uint128,
+    name: String,
+    payment_amount_to_register_name: Uint128,
+) -> Result<Response, ContractError> {
+    if amount_sent != payment_amount_to_register_name {
```
janitachalam, line 165: nit: maybe this can be checked before the call to register_name?

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,330 @@
...
+    if NAME_TO_DAO.has(deps.storage, name.clone()) {
+        return Err(ContractError::NameAlreadyTaken {});
+    }
+
+    if RESERVED_NAMES.has(deps.storage, name.clone()) {
+        return Err(ContractError::NameReserved {});
+    }
+
+    // Only the admin can reserve names
+    if info.sender != config.admin {
```
janitachalam, line 262: nit: move this check first

janitachalam:
looks :fly:

contracts/cw-names-registry/src/contract.rs
Callum-A: i placed it here to reduce duplication as this function is called from both execute_reserve_native and the cw20 receive hook

contracts/cw-names-registry/src/contract.rs
```diff
@@ -0,0 +1,330 @@
...
+pub fn register_name(
+    deps: DepsMut,
+    _env: Env,
+    sender: String,
+    amount_sent: Uint128,
+    name: String,
+    payment_amount_to_register_name: Uint128,
+) -> Result<Response, ContractError> {
+    if amount_sent != payment_amount_to_register_name {
+        return Err(ContractError::InsufficientFunds {});
```
elsehow, line 166: Is `InsufficientFunds` the right error here? It looks like we're also erroring if the funds are too high. I can see this confusing someone.

contracts/cw-names-registry/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("Unauthorized")]
+    Unauthorized {},
+
+    #[error("Insufficient funds sent")]
```
elsehow, line 12: See note above - something like `"Wrong payment amount for name registration sent."`

Error might be `IncorrectPaymountAmount`.

contracts/cw-names-registry/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+    InsufficientFunds {},
+
+    #[error("This name is reserved for later user")]
```
elsehow, line 15: This name is reserved for later use*

elsehow (approved):
Little nits + pushed a README helper. LGTM!

PR #293, opened 2022-05-20
Allow revoting on proposals.

ezekiiel:
This PR adds a config option to the `cw_proposal_single` contract to allow revoting on proposals. If this config option is toggled:

- Proposals will not be passed or rejected until they have expired.
- Voters may cast new votes before the proposal expires.

### DIFF
--- a/contracts/cw-proposal-single/src/contract.rs
+++ b/contracts/cw-proposal-single/src/contract.rs
@@ -57,6 +57,7 @@ pub fn instantiate(
         only_members_execute: msg.only_members_execute,
         dao: dao.clone(),
         deposit_info,
+        allow_revoting: msg.allow_revoting,
     };
 
     CONFIG.save(deps.storage, &config)?;
@@ -86,6 +87,7 @@ pub fn execute(
             threshold,
             max_voting_period,
             only_members_execute,
+            allow_revoting,
             dao,
             deposit_info,
         } => execute_update_config(
@@ -94,6 +96,7 @@ pub fn execute(
             threshold,
             max_voting_period,
             only_members_execute,
+            allow_revoting,
             dao,
             deposit_info,
         ),
@@ -166,6 +169,7 @@ pub fn execute_propose(
             msgs,
             status: Status::Open,
             votes: Votes::zero(),
+            allow_revoting: config.allow_revoting,
             deposit_info: config.deposit_info.clone(),
         };
         // Update the proposal's status. Addresses case where proposal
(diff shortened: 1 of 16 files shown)


contracts/cw-proposal-single/src/contract.rs
```diff
@@ -297,11 +301,29 @@ pub fn execute_vote(
...
         (proposal_id, info.sender.clone()),
         |bal| match bal {
-            Some(_) => Err(ContractError::AlreadyVoted {}),
+            Some(current_ballot) => {
+                if prop.allow_revoting {
+                    if current_ballot.vote == vote {
+                        // Don't allow casting the same vote more than
+                        // once. This seems liable to be confusing
+                        // behavior.
+                        Err(ContractError::AlreadyCast {})
```
JakeHartnell, line 315: Not sure how useful this error message is? Could just overwrite the previous vote regardless of whether it was the same? No strong opinions here though.

contracts/cw-proposal-single/src/proposal.rs
```diff
@@ -308,18 +332,199 @@ mod test {
...
+            threshold,
+            votes,
+            Uint128::new(15),
+            true,
+            true
+        ));
+    }
+
+    #[test]
+    fn test_absolute_threshold() {
```
JakeHartnell, line 423: These are great! A few comments on the test cases would be nice to have.

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -2339,3 +2363,883 @@ fn test_active_threshold_none() {
...
+        proposal.proposal.votes,
+        Votes {
+            yes: Uint128::zero(),
+            no: Uint128::new(100),
+            abstain: Uint128::zero()
+        }
+    );
+
+    // As the clock strikes midnight on the last day of the proposal,
+    // revoting has saved the day!
```
JakeHartnell, line 2480: The people of Floob made the right choice!

JakeHartnell (approved):
Beautiful! :heart:

contracts/cw-proposal-single/src/error.rs
```diff
@@ -36,9 +36,12 @@ pub enum ContractError {
     #[error("Not registered to vote (no voting power) at time of proposal creation.")]
     NotRegistered {},
 
-    #[error("Already voted")]
+    #[error("Already voted. This proposal does not support revoting.")]
     AlreadyVoted {},
 
+    #[error("Already cast that vote. Change your vote to revote.")]
```
Callum-A, line 42: This isn't the most clear as we use AlreadyVoted above, "Already cast a vote with this option."

contracts/cw-proposal-single/src/proposal.rs
Callum-A: Agreed on this front but great coverage

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -2339,3 +2363,883 @@ fn test_active_threshold_none() {
...
+        Threshold::AbsoluteCount {
+            threshold: Uint128::new(11),
+        },
+        Status::Passed,
+        None,
+    );
+}
+
+#[test]
+fn test_large_absolute_threshold() {
```
Callum-A, line 3200: awesome coverage

Callum-A (approved):
Small nits inline with Jake's findings but all good otherwise

contracts/cw-proposal-single/src/contract.rs
ezekiiel: That was my initial thought as well. Agree the error is a little arbitrary. Was trying to think about this from the perspective of a user of a frontend. If I voted once, and then immediately cast the same vote and didn't see an error it is hard to tell the difference between the first vote succeeding and the first one failing. Think in some ways this serves as a confirmation.

Also, in a frontend we build for this we will almost certainly not allow casting the same vote more than one.

contracts/cw-proposal-single/src/error.rs
ezekiiel: Great point. Updated the wording.

PR #295, opened 2022-05-21
Add test demonstrating interesting hook behavior.

ezekiiel:
This adds a test which is currently failing to test for the case where a proposal's state goes from `open -> executed`. Open question is: should we fire two hooks in that case (one for `open -> passed` and one for `passed -> executed`)?

The same issue applies for proposals going from `open -> closed` and skipping rejected. These things happen if a proposal passes / fails due to it expiring and not due to it being voted on.

The test is currently failing. We should discuss what behavior we want here and update logic / the test as needed.


ezekiiel:
After discussion with @Callum-A - think our current behavior is the correct one. We should write docs about this behavior and module authors should be expected to handle this. Closing in favor of: https://github.com/DA0-DA0/dao-contracts/issues/299

PR #297, opened 2022-05-22
Update to CosmWasm 1.0.0!

ezekiiel:
This PR updates all of our dependencies to use CosmWasm 1.0.0!

The actual changes are the result of running:

```bash
git grep -l '1.0.0-beta' | xargs sed -i '' -e 's/1\.0\.0-beta[0-9]?/1.0.0/g' -E
rm Cargo.lock
cargo build
```

After these changes there are no yanked dependencies in our dependency tree per a run of `cargo audit`.

Edit: Seems like CosmWasm made some types implement `Eq` so now clippy wants us to derive that as we get it for free. These changes were generated by running:

```
cargo +nightly clippy --fix --all-targets -- -D warnings
```


janitachalam (approved):

NoahSaso (approved):
🔥

PR #300, opened 2022-05-23
Implement optional admins for SubDAOs

### LINKED ISSUE #283: Optional DAO Admin
Use case: SubDAOs where the parent DAO wants to maintain control over the child DAO.

This should probably go into the core DAO contract. Now some DAOs will not have any admin, but something like `ExcuteMsg::AdminExecuteProposal` could be very useful if a hierarchy is needed between a main DAO and its SubDAOs.

There could additionally be an `UpdateAdmin` only callable by the current admin.

JakeHartnell:
Closes #283

### DIFF
--- a/contracts/cw-core/src/contract.rs
+++ b/contracts/cw-core/src/contract.rs
@@ -459,26 +459,32 @@ pub fn execute_receive_cw721(deps: DepsMut, sender: Addr) -> Result<Response, Co
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
     match msg {
+        QueryMsg::Admin {} => query_admin(deps),
         QueryMsg::Config {} => query_config(deps),
-        QueryMsg::VotingModule {} => query_voting_module(deps),
-        QueryMsg::ProposalModules { start_at, limit } => {
-            query_proposal_modules(deps, start_at, limit)
-        }
-        QueryMsg::DumpState {} => query_dump_state(deps, env),
-        QueryMsg::VotingPowerAtHeight { address, height } => {
-            query_voting_power_at_height(deps, address, height)
-        }
-        QueryMsg::TotalPowerAtHeight { height } => query_total_power_at_height(deps, height),
-        QueryMsg::GetItem { key } => query_get_item(deps, key),
-        QueryMsg::ListItems { start_at, limit } => query_list_items(deps, start_at, limit),
-        QueryMsg::Info {} => query_info(deps),
         QueryMsg::Cw20TokenList { start_at, limit } => query_cw20_list(deps, start_at, limit),
-        QueryMsg::Cw721TokenList { start_at, limit } => query_cw721_list(deps, start_at, limit),
         QueryMsg::Cw20Balances { start_at, limit } => {
             query_cw20_balances(deps, env, start_at, limit)
         }
+        QueryMsg::Cw721TokenList { start_at, limit } => query_cw721_list(deps, start_at, limit),
+        QueryMsg::DumpState {} => query_dump_state(deps, env),
+        QueryMsg::GetItem { key } => query_get_item(deps, key),
+        QueryMsg::Info {} => query_info(deps),
+        QueryMsg::ListItems { start_at, limit } => query_list_items(deps, start_at, limit),
         QueryMsg::PauseInfo {} => query_paused(deps, env),
+        QueryMsg::ProposalModules { start_at, limit } => {
+            query_proposal_modules(deps, start_at, limit)
+        }
+        QueryMsg::TotalPowerAtHeight { height } => query_total_power_at_height(deps, height),
+        QueryMsg::VotingModule {} => query_voting_module(deps),
+        QueryMsg::VotingPowerAtHeight { address, height } => {
+            query_voting_power_at_height(deps, address, height)
+        }
     }
+}
+
+pub fn query_admin(deps: Deps) -> StdResult<Binary> {
+    let admin = ADMIN.load(deps.storage)?;
+    to_binary(&admin)
 }
 
 pub fn query_config(deps: Deps) -> StdResult<Binary> {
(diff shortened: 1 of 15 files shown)


contracts/cw-core/src/contract.rs
```diff
@@ -181,19 +189,67 @@ pub fn execute_pause(
...
         .add_attribute("action", "execute_proposal_hook")
         .add_messages(msgs))
 }
 
+pub fn execute_update_admin(
+    deps: DepsMut,
+    sender: Addr,
+    admin: Option<Addr>,
+) -> Result<Response, ContractError> {
+    let current_admin = ADMIN.load(deps.storage)?;
```
ezekiiel, line 234: Might be nice to use `may_load` and return `ContractError::NoAdmin` here. Otherwise the error seems liable to be pretty unreadable.

contracts/cw-core/src/msg.rs
```diff
@@ -82,43 +85,15 @@ pub struct InstantiateMsg {
...
-        key: String,
-        addr: String,
-    },
-    /// Removes an item from the governance contract's item map.
-    RemoveItem {
-        key: String,
-    },
+    ExecuteProposalHook { msgs: Vec<CosmosMsg<Empty>> },
+    /// Pauses the DAO for a set duration.
+    /// When paused the DAO is unable to execute propopasls
```
ezekiiel, line 95:
```suggestion
    /// When paused the DAO is unable to execute proposals
```

contracts/cw-core/src/contract.rs
```diff
@@ -181,19 +189,67 @@ pub fn execute_pause(
         .add_attribute("until", until.to_string()))
 }
 
+pub fn execute_admin_msgs(
```
ezekiiel, line 192: What do you think about letting the admin execute a message on a paused DAO? Think that could make sense in the theme of "admin" can control the subDAO.

contracts/cw-core/src/state.rs
```diff
@@ -22,6 +22,7 @@ pub struct Config {
     pub automatically_add_cw721s: bool,
 }
 
+pub const ADMIN: Item<Addr> = Item::new("admin");
```
ezekiiel, line 25: I normally prefer to use `Item<Option<Addr>>` in these cases. Seems slightly less error prone as it forces the programmer to recon with the type being optional. Otherwise, might be nice to have a comment here explaining that this field will not always be set.

contracts/cw-core/src/tests.rs
```diff
@@ -821,9 +829,164 @@ fn do_standard_instantiate(auto_add: bool) -> (Addr, App) {
...
+        &ExecuteMsg::UpdateAdmin {
+            admin: Some(Addr::unchecked("rando")),
+        },
+        &[],
+    );
+    assert!(res.is_err());
+
+    // Update Admin can't be called, even by the proposal module
+    let res = app.execute_contract(
+        proposal_module.clone(),
```
ezekiiel, line 902: Think we already have a test for this above. How about checking that the DAO can't call this on itself? This is how a proposal module's message would get executed.
```suggestion
    // Update Admin can't be called, even by the core module
    let res = app.execute_contract(
       core_addr.clone(),
```

ezekiiel (commented):
Looks great! Just some questions / a test suggestion.

contracts/cw-core/src/state.rs
JakeHartnell: Done.

contracts/cw-core/src/contract.rs
```diff
@@ -181,19 +187,81 @@ pub fn execute_pause(
...
+            let res;
+            if let Some(admin) = admin {
+                res = Response::default()
+                    .add_attribute("action", "execute_update_admin")
+                    .add_attribute("new_admin", admin.to_string());
+            } else {
+                res = Response::default()
+                    .add_attribute("action", "execute_update_admin")
+                    .add_attribute("new_admin", "none");
+            }
```
JakeHartnell, line 254: Might be a cleaner way to do this. Open to suggestions.

Do think it's good to have the new admin as an attribute. @ezekiiel / @Callum-A do we have a standard way of encoding optional attributes? Wrote `"none"` here, but hey it's still a draft.

ezekiiel: I like to do something like this:
```rust
admin
    .map(|a| a.into_string())
    .unwrap_or_else(|| "None".to_string()),
```
Not terribly simple, but I think it gets the point across reasonably concisely.

JakeHartnell: Great! Nice and clean!

contracts/cw-core/src/contract.rs
Callum-A: this also could prevent some bricked sub DAO situations!

Callum-A: this also could prevent some bricked sub DAO situations!

contracts/cw-core/src/msg.rs
```diff
@@ -51,6 +51,9 @@ pub struct InitialItem {
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 pub struct InstantiateMsg {
+    /// Optional Admin with the ability to execute DAO messages directly
+    /// Useful for building SubDAOs completely controlled by a parent DAO.
+    pub admin: Option<Addr>,
```
ezekiiel, line 56: Sorry I didn't catch this earlier - the type here needs to be `Option<String>` and then we need to validate it with `let admin = msg.admin.map(|human| deps.api.addr_validate(&human)).transpose()?;` in the instantiate method.

`Addr` types inside of messages will not be validated so this could currently set the admin to a invalid address.

ezekiiel: Also need to run `cargo schema` to get this into the schema files.

ezekiiel (approved):
Great!

Callum-A (approved):

janitachalam (approved):

PR #309, opened 2022-05-28
Add stake_cw721 and cw721-controllers

### LINKED ISSUE #131: NFT DAO
As a user, I would like to create a DAO with membership based on NFT ownership.

For this we'll need a custom staking contract based around NFTs, as well as adding another option to the DAO instantiation message.

cypherape:
Adds the ability to stake Cw721 Tokens. Allows for staked Nft use cases.

Closes #131


contracts/stake-cw721/README.md
```diff
@@ -0,0 +1,22 @@
+# Stake CW20
```
ezekiiel, line 1: Probably want to update this README.

contracts/stake-cw721/src/msg.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    ReceiveNft(Cw721ReceiveMsg),
+    Receive(Cw20ReceiveMsg),
+    Unstake {
+        token_id: String,
+        reward_wallet_address: Option<Addr>,
```
ezekiiel, line 30: When it comes in as part of a message `Addr` does not get validated. Need to use a string and then validate in the receiving method. :)
```suggestion
        reward_wallet_address: Option<String>,
```

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+pub fn instantiate(
+    deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response<Empty>, ContractError> {
+    let owner = match msg.owner {
+        Some(owner) => Some(deps.api.addr_validate(owner.as_str())?),
+        None => None,
+    };
```
ezekiiel, line 44: Instead of the match statement you can use map.

```suggestion
    let owner = msg.owner.map(|h| deps.api.addr_validate(&h)).transpose()?;
```

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+
+    let reward_token = match msg.reward_token_address {
+        Some(reward_token) => Some(deps.api.addr_validate(&*reward_token)?),
+        None => None,
+    };
+
+    let config = Config {
+        owner,
+        manager,
+        nft_address: deps.api.addr_validate(&*msg.nft_address)?,
```
ezekiiel, line 59: Do we really need to deref and then ref here? `&msg.nft_address` should likely work fine.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+        }
+
+        let sender = deps.api.addr_validate(&wrapper.sender)?;
+        let balance = REWARD_BALANCE.load(deps.storage).unwrap_or_default();
+        REWARD_BALANCE.save(
+            deps.storage,
+            &balance
+                .checked_add(wrapper.amount)
+                .map_err(StdError::overflow)?,
+        )?;
```
ezekiiel, line 118: Think the `update` method does what you want here. Something like:
```suggestion
        REWARD_BALANCE.update(deps.storage, |old| old.unwrap_or_default().checked_add(wrapper.amount).map_err(StdError::overflow))?;
```

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+}
+
+pub fn execute_unstake(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    token_id: String,
+    reward_wallet_address: Option<Addr>,
+) -> Result<Response, ContractError> {
+    let validated_reward_wallet = match reward_wallet_address {
```
ezekiiel, line 180: Can use map method from instantiate here instead of the match.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+        Some(validated_reward_wallet) => {
+            Some(deps.api.addr_validate(validated_reward_wallet.as_str())?)
+        }
+        None => None,
+    };
+    let sender = deps.api.addr_validate(info.sender.as_str())?;
```
ezekiiel, line 186: `info.sender` should already be an `Addr`. We can trust the value so no need to validate here. :)

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+                .claims;
+            if outstanding_reward_claims.len() >= MAX_CLAIMS as usize {
+                return Err(ContractError::TooManyClaims {});
+            }
+
+            REWARD_CLAIMS.create_claim(
+                deps.storage,
+                &validated_reward_wallet
+                    .clone()
+                    .unwrap_or_else(|| DEFAULT_ADDR.clone()),
```
ezekiiel, line 319: Seems like we just shouldn't create the claim if the reward wallet is `None`? Sending to an invalid address seems liable to cause confusion.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+use cw20::Cw20ReceiveMsg;
+use cw721::Cw721ReceiveMsg;
+use cw721_controllers::NftClaimsResponse;
+use cw_controllers::ClaimsResponse;
+use cw_utils::Duration;
+use std::collections::HashSet;
+use std::convert::{From, TryFrom};
+
+lazy_static! {
+    static ref DEFAULT_ADDR: Addr = Addr::unchecked("123");
```
ezekiiel, line 28: Why do we need this? Is there ever a situation where a user actually wants their claims to go to a burner address? This being an invalid address also seems liable to cause trouble.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+    _env: Env,
+    info: MessageInfo,
+) -> Result<Response, ContractError> {
+    let nfts = NFT_CLAIMS.claim_nfts(deps.storage, &info.sender, &_env.block)?;
+    if nfts.is_empty() {
+        return Err(ContractError::NothingToClaim {});
+    }
+
+    let config = CONFIG.load(deps.storage)?;
+    let mut msgs = Vec::new();
```
ezekiiel, line 350: Instead of making a mutable vector and pushing to it repeatedly how about doing this with iterators and `collect`? If we do it like that the compiler can do one allocation at the end instead of having to repeatedly resize the vector when we push.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+    }
+
+    let config = CONFIG.load(deps.storage)?;
+    let cw_send_msg = cw20::Cw20ExecuteMsg::Transfer {
+        recipient: info.sender.to_string(),
+        amount: reward_amount,
+    };
+
+    let wasm_msg = cosmwasm_std::WasmMsg::Execute {
+        contract_addr: config.reward_token_address.unwrap().to_string(),
```
ezekiiel, line 389: This unwrap here seems liable to cause trouble. Could we return an error if this is `None` instead? Understand that it may be the case currently that we will never hit this condition if there is no reward token.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+
+pub fn execute_update_config(
+    info: MessageInfo,
+    deps: DepsMut,
+    new_owner: Option<String>,
+    new_manager: Option<String>,
+    duration: Option<Duration>,
+) -> Result<Response, ContractError> {
+    let new_owner = new_owner
+        .map(|new_owner| deps.api.addr_validate(&*new_owner))
```
ezekiiel, line 409: What is the `&*` doing here? Possible we really do need this and my Rust reading just isn't good enough.

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+        height,
+    })
+}
+
+pub fn query_total_staked_at_height(
+    deps: Deps,
+    _env: Env,
+    height: Option<u64>,
+) -> StdResult<TotalStakedAtHeightResponse> {
+    let height = height.unwrap_or(_env.block.height);
```
ezekiiel, line 526:
```suggestion
    env: Env,
    height: Option<u64>,
) -> StdResult<TotalStakedAtHeightResponse> {
    let height = height.unwrap_or(env.block.height);
```

contracts/stake-cw721/src/state.rs
```diff
@@ -0,0 +1,45 @@
...
+pub const CONFIG: Item<Config> = Item::new("config");
+
+pub const STAKED_NFTS_PER_OWNER: SnapshotMap<&Addr, HashSet<String>> = SnapshotMap::new(
+    "staked_nfts_per_owner",
+    "staked_nfts_per_owner__checkpoints",
+    "staked_nfts_per_owner__changelog",
+    Strategy::EveryBlock,
+);
+
+pub const TOTAL_STAKED_NFTS: SnapshotItem<HashSet<String>> = SnapshotItem::new(
```
ezekiiel, line 28: From my read it seems like we only ever use the length of this `HashSet`. How about just storing a `u64` here?

contracts/stake-cw721/src/contract.rs
```diff
@@ -0,0 +1,1844 @@
...
+
+pub fn execute_fund(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    wrapper: Cw20ReceiveMsg,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if let Some(reward_token_address) = config.reward_token_address {
+        let reward_token_address = deps.api.addr_validate(reward_token_address.as_ref())?;
```
ezekiiel, line 103: Shouldn't need to validate here. Validated before storing in config.

ezekiiel (commented):
This is looking really great. Thanks for putting this up! Made a pass and added some comments. Will make another one once those are discussed.

contracts/stake-cw721/src/contract.rs
cypherape: Makes sense

contracts/stake-cw721/src/contract.rs
cypherape: Done.

contracts/stake-cw721/src/contract.rs
cypherape: Done.

contracts/stake-cw721/src/contract.rs
cypherape: Done.

contracts/stake-cw721/src/contract.rs
cypherape: Done.

contracts/stake-cw721/src/contract.rs
cypherape: Cool.

contracts/stake-cw721/src/contract.rs
cypherape: Cool.

contracts/stake-cw721/src/contract.rs
cypherape: Cool.

contracts/stake-cw721/src/contract.rs
cypherape: Cool.

contracts/stake-cw721/src/contract.rs
cypherape: Cool.

contracts/stake-cw721/src/msg.rs
cypherape: Done.

contracts/stake-cw721/README.md
cypherape: Done.

contracts/stake-cw721/src/contract.rs
cypherape: No, actually we do not need &*. Updated it!

contracts/stake-cw721/src/contract.rs
cypherape: How do we handle the possible error from to_binary(&cw_transfer_msg)?

contracts/stake-cw721/src/state.rs
cypherape: Agreed! How about adding a new query to fetch all the staked tokens? I think this would be a good query function to have. What's your take on this?

contracts/stake-cw721/src/hooks.rs
```diff
@@ -0,0 +1,55 @@
+use crate::state::HOOKS;
+use cosmwasm_std::{to_binary, Addr, StdResult, Storage, SubMsg, WasmMsg};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+// This is just a helper to properly serialize the above message
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum StakeChangedHookMsg {
```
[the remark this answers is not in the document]
cypherape: Thanks👍

contracts/stake-cw721/src/state.rs
JakeHartnell: > Agreed! How about adding a new query to fetch all the staked tokens? I think this would be a good query function to have. What's your take on this?

Yeah, I like this idea! Makes sense for NFTs. Should have pagination.

contracts/cw721-stake/src/msg.rs
```diff
@@ -0,0 +1,110 @@
...
+    Unstake {
+        token_id: String,
+        reward_wallet_address: Option<String>,
+    },
+    ClaimNfts {},
+    ClaimRewards {},
+    UpdateConfig {
+        owner: Option<String>,
+        manager: Option<String>,
+        duration: Option<Duration>,
```
ezekiiel, line 36: nit:
```suggestion
        unstaking_duration: Option<Duration>,
```

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1821 @@
...
+        .transpose()?;
+    let sender = info.sender.clone();
+    let config = CONFIG.load(deps.storage)?;
+    let sender_staked_nft_collection = STAKED_NFTS_PER_OWNER
+        .may_load(deps.storage, &sender)?
+        .unwrap_or_default();
+    if !sender_staked_nft_collection.contains(&token_id) {
+        return Err(ContractError::Unauthorized {});
+    }
+    let balance = REWARD_BALANCE.load(deps.storage).unwrap_or_default();
```
ezekiiel, line 177: Because we initialize this to zero in instantiate might be best to error here if `load` fails. This hitting the default case here suggests to me that something has gone wrong.

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1821 @@
...
+    let total_staked_nfts = TOTAL_STAKED_NFTS.load(deps.storage)?;
+    let amount_to_claim = balance
+        .checked_div(Uint128::from(
+            u128::try_from(total_staked_nfts.len()).unwrap(),
```
ezekiiel, line 181: Could we add a comment somewhere (likely in `state.rs`) about this `len` call being constant time? Something to the effect of:

1. HashSet uses `hashbrown::HashMap`.
2. `hashbrown::HashMap` uses `hashbrown::RawTable`.
3. `hashbrown::RawTable` len calls are constant time: https://docs.rs/hashbrown/latest/src/hashbrown/raw/mod.rs.html#921-923

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1821 @@
...
+        ))
+        .map_err(StdError::divide_by_zero)?;
```
ezekiiel, line 183: With these calculations I can drain the rewards pool.

Say there are 100 reward tokens and 10 NFTs. I have one staked.

1. I unstake my NFT and receive 10 rewards tokens => 90 rewards tokens, 9 NFTs.
2. I stake my NFT => 90 rewards tokens, 10 NFTs.
3. I unstake my NFT and receive 9 rewards tokens => 81 rewards tokens, 9 NFTs.
4. ... repeat

The way that the cw20-stake contract gets around this is that when you stake your balance increases by the proportional ownership and not the amount of tokens you have staked. From `execute_stake`:

```rust
    let amount_to_stake = if staked_total == Uint128::zero() || balance == Uint128::zero() {
        amount
    } else {
        staked_total
            .checked_mul(amount)
            .map_err(StdError::overflow)?
            .checked_div(balance)
            .map_err(StdError::divide_by_zero)?
    };
```

I _think_ this is possible to do with NFT staking if we store `STAKED_BALANCE` and also NFT information and then use NFT information only for sending the correct NFT back and not for voting power calculations. This is tricky stuff.

Your choice on how to move forward here. I would be fine just removing rewards from this contract entirely though. We can start simple and then add them in later.

If we do choose to go the remove rewards route though it would be great to keep the `StakedValue` query around. The frontend uses this query for determining voting power and it would make our lives a little simpler over there. In this case it would just return the number of NFTs staked.

ezekiiel (changes requested):
Rewards are hard and I think this has a bug where one can drain rewards from the contract. I missed this in the first review and an older version of the staking contract actually had the same problem. Many have struggled with this one. :)

My recommendation would be that we just rip rewards out of this contract for now. The nice thing about these contracts is that they can always be upgraded later so we can start simple. :)

contracts/cw721-stake/src/contract.rs
ezekiiel: After writing this review I realized my flash loan program had the same issue aha https://github.com/ezekiiel/cw-flash-loan/commit/3b77e6bc2c1c02f359c3430329c77917e3b9b3fc. You are not alone in running into this sort of thing.

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -72,6 +72,24 @@ fn cw20_staked_balances_voting() -> Box<dyn Contract<Empty>> {
     Box::new(contract)
 }
 
+fn cw721_base() -> Box<dyn Contract<Empty>> {
```
Callum-A, line 75: nit: i see a future where we want to have a `tests` dir and seperate via different voting power modules, e.g. cw20, cw4, cw721

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1536 @@
...
+        .transpose()?;
+
+    let config = Config {
+        owner: owner.clone(),
+        manager,
+        nft_address: deps.api.addr_validate(&msg.nft_address)?,
+        unstaking_duration: msg.unstaking_duration,
+    };
+    CONFIG.save(deps.storage, &config)?;
+    TOTAL_STAKED_NFTS.save(deps.storage, &Uint128::zero(), env.block.height)?;
```
Callum-A, line 53: nit: is this strictly necessary, as long as we use `may_load_at_height`

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1536 @@
...
+            Ok(updated_nft_collection)
+        },
+    )?;
+
+    TOTAL_STAKED_NFTS.update(
+        deps.storage,
+        env.block.height,
+        |total_staked| -> StdResult<_> {
+            total_staked
+                .unwrap()
```
Callum-A, line 120: nit: with my above change this could simply be `unwrap_or_default` and then we can remove the setting to 0

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1536 @@
...
+            }
+        },
+    )?;
+
+    TOTAL_STAKED_NFTS.update(
+        deps.storage,
+        env.block.height,
+        |total_staked| -> StdResult<_> {
+            total_staked
+                .unwrap()
```
Callum-A, line 176: as above

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1536 @@
...
+            Ok(Response::default()
+                .add_messages(return_messages)
+                .add_submessages(hook_msgs)
+                .add_attribute("action", "unstake")
+                .add_attribute("from", info.sender)
+                .add_attribute("claim_duration", "None"))
+        }
+
+        Some(duration) => {
+            let outstanding_claims = NFT_CLAIMS
```
Callum-A, line 208: top work with the NFT claims

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,1536 @@
...
+    })
+}
+
+pub fn query_info(deps: Deps) -> StdResult<Binary> {
+    let info = cw2::get_contract_version(deps.storage)?;
+    to_binary(&cw_core_interface::voting::InfoResponse { info })
+}
+
+#[cfg(test)]
+mod tests {
```
Callum-A, line 456: nit: move to own file?

Callum-A (approved):
LGTM, some small nits

contracts/cw721-stake/src/contract.rs
ezekiiel: Done. :)

contracts/cw-proposal-single/src/tests.rs
Callum-A: should of been more clear here: this is probably future work

contracts/cw721-stake/src/contract.rs
ezekiiel: I've been reflecting on this a little bit and I'm leaning towards the initialize to zero and unwrap approach. My thought is that this more clearly delineates when the contract is behaving as expected. 

For example: say an attacker is somehow able to remove the `TOTAL_STAKED_NFTS` storage data - if we use `unwrap_or_default` this is invisible as we'll just start back at zero. If we initialize to zero and then unwrap post-attack the operations will fail and the data invalidation will be caught.

What do you think about this? My thinking here is dependent on the idea what we'd rather have methods fail than unwrap and continue forward.

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+    entry_point, to_binary, Binary, CosmosMsg, Deps, DepsMut, Empty, Env, MessageInfo, Response,
+    StdError, StdResult, Uint128, WasmMsg,
+};
+use cw2::set_contract_version;
+use cw721::Cw721ReceiveMsg;
+use cw_utils::Duration;
+use std::collections::HashSet;
+use std::convert::{From, TryFrom};
+
+pub(crate) const CONTRACT_NAME: &str = "crates.io:stake_cw721";
```
janitachalam, line 21: nit: cw721-stake here

packages/cw721-controllers/src/nft_claim.rs
```diff
@@ -0,0 +1,422 @@
...
+            });
+            Ok(nft_claims)
+        })?;
+        Ok(())
+    }
+
+    /// Creates a number of NFT claims simeltaniously for a given
+    /// address. Similar to create_nft_claim but does only one storage
+    /// read and write for multiple claims.
+    pub fn create_nft_claims(
```
janitachalam, line 58: can we just always use this function instead of create_nft_claim, and in the case of a single token_id, have a vector with a single element?

packages/cw721-controllers/src/nft_claim.rs
```diff
@@ -0,0 +1,422 @@
...
+        block: &BlockInfo,
+    ) -> StdResult<Vec<String>> {
+        let mut to_send = vec![];
+        self.0.update(storage, addr, |nft_claim| -> StdResult<_> {
+            let (_send, waiting): (Vec<_>, _) =
+                nft_claim.unwrap_or_default().into_iter().partition(|c| {
+                    // if mature and we can pay fully, then include in _send
+                    if c.release_at.is_expired(block) {
+                        // TODO: handle partial paying claims?
+                        to_send.push(c.token_id.clone());
```
janitachalam, line 93: can we use the send vec that is returned from `partition`?

packages/cw721-controllers/src/nft_claim.rs
```diff
@@ -0,0 +1,422 @@
+use cosmwasm_std::{Addr, BlockInfo, CustomQuery, Deps, StdResult, Storage};
+use cw_storage_plus::Map;
+use cw_utils::Expiration;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
```
janitachalam, line 7: does this file need to be in packages because it will also be used by another contract other than `stake-cw721`?

packages/cw721-controllers/src/nft_claim.rs
ezekiiel: Great idea. We don't even use the non vec version at the moment.

packages/cw721-controllers/src/nft_claim.rs
ezekiiel: That's the idea. 🤷‍♂️ In theory this has a really similar setup to cw-controllers in cw-plus which is used for a variety of things.

contracts/cw721-stake/src/state.rs
```diff
@@ -0,0 +1,48 @@
...
+
+/// Maps addresses to the set of NFTs they have staked with this
+/// contract at a given height.
+///
+/// NOTE: we can get the length of this set in constant time as:
+/// - HashSet uses hashbrown::HashMap.
+/// - hashbrown::HashMap uses hashbrown::RawTable.
+/// - hashbrown::RawTable len calls are
+///   [constant time](https://docs.rs/hashbrown/latest/src/hashbrown/raw/mod.rs.html#921-923).
+pub const STAKED_NFTS_PER_OWNER: SnapshotMap<&Addr, HashSet<String>> = SnapshotMap::new(
```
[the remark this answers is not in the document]
ezekiiel: Apparently it does lol. Was able to compile and store this fine. Great question.

contracts/cw721-stake/src/msg.rs
```diff
@@ -0,0 +1,105 @@
...
+use cw_core_macros::voting_query;
+use cw_utils::Duration;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+pub use cw721_controllers::NftClaimsResponse;
+
+#[derive(Serialize, Deserialize, Clone, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum Owner {
```
Callum-A, line 12: is this strictly necessary considering i can just paste the contract address into the field?

contracts/cw721-stake/src/msg.rs
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Serialize, Deserialize, Clone, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub struct TotalStakedAtHeightResponse {
+    pub total: Uint128,
+    pub height: u64,
+}
+
+#[derive(Serialize, Deserialize, Clone, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct GetHooksResponse {
```
Callum-A, line 103: this seems p generic across our staking-hooks should probably move it there eventually (future work)

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+                // filter. Remove in a loop clocks in at ~2x the speed
+                // of difference.
+                for token_id in token_ids.iter() {
+                    // This will implicitly check for duplicates in
+                    // the input vector as removing twice will fail
+                    // the second time around.
+                    let was_present = nft_collection.remove(token_id);
+                    if !was_present {
+                        // Can't unstake that which you do not own.
+                        return Err(ContractError::NotStaked {});
```
janitachalam, line 168: if there's a token ID which they do not own in there, should we still unstake the valid ones, or should we fail the whole operation as we do now?  

I think it's probably fine as is but just thought I'd call it out

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+}
+
+pub fn execute_update_config(
+    info: MessageInfo,
+    deps: DepsMut,
+    new_owner: Option<String>,
+    new_manager: Option<String>,
+    duration: Option<Duration>,
+) -> Result<Response, ContractError> {
+    let new_owner = new_owner
```
janitachalam, line 289: I know the other staking contract follows this owner / manager pattern, but since AFAIK none of our other modules do that, what's the reasoning to add it to the staking contracts in particular?

contracts/cw721-stake/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+    let manager = msg
+        .manager
+        .as_ref()
+        .map(|h| deps.api.addr_validate(h))
+        .transpose()?;
+
+    let config = Config {
+        owner: owner.clone(),
+        manager,
+        nft_address: deps.api.addr_validate(&msg.nft_address)?,
```
janitachalam, line 48: so basic question, why do we only allow staking one kind of NFT?

contracts/cw721-stake/src/msg.rs
ezekiiel: Intention here is to allow a DAO to instantiate this with itself as the owner during its own instantiation. If we just make this field an address you won't be able to put the address of the DAO that will be instantiated in the `ModuleInstantiateInfo`. Our current voting modules don't need this because the wrapper contract can handle the admin transfer.

contracts/cw721-stake/src/msg.rs
ezekiiel: 100%.

contracts/cw721-stake/src/contract.rs
ezekiiel: Good point. I'd probably prefer failing the whole operation here. If you're attempting to unstake a NFT you don't own something is likely amis. Think it also fits into the general "transaction either completely succeeds or completely fails" model that blockchains tend to use.

contracts/cw721-stake/src/contract.rs
ezekiiel: Good question. The original intent was to allow a DAO to delegate the ability to change the unstaking duration to a subDAO. At this point I think it's just become a pattern I do and don't think terribly critically about. What do you think? It is worth keeping around?

contracts/cw721-stake/src/contract.rs
ezekiiel: Such a good question! We could without too much trouble. Would just worry that this would require more thinking about how to assign weights between the two collections and etc. which might be left for later. @Callum-A's voting module weighted combiner will give this to us for free as well.

ezekiiel (approved):

janitachalam (approved):

contracts/cw721-stake/src/contract.rs
janitachalam: I understand that use case, but I think I would err on the side of removing it for now since it seems a bit like we are over-engineering for future scenarios. My main concern was just that the admin story seems to be getting more muddled over time with several different types of admins (the wasm admin, the subdao admin, owners, managers) -- I think we could take a step back and think about the different types of roles / authorizations that we want to support, taking into consideration the authorization middleware as well.

contracts/cw721-stake/src/contract.rs
JakeHartnell: > Such a good question! We could without too much trouble. Would just worry that this would require more thinking about how to assign weights between the two collections and etc. which might be left for later. @Callum-A's voting module weighted combiner will give this to us for free as well.

Exactly!

contracts/cw721-stake/src/contract.rs
JakeHartnell: > I understand that use case, but I think I would err on the side of removing it for now since it seems a bit like we are over-engineering for future scenarios.

Can do this as a separate issue / ticket. We should make whatever we decide consistent everywhere.

PR #313, opened 2022-05-28
Start native voting impl

Callum-A:
Need feedback on how we should handle the `TotalPowerAtHeight` route as this is important for determining thresholds.


Callum-A:
Closing for now after speaking with Zeke on some details

PR #316, opened 2022-05-30
rust-optimizer -> workspace-optimizer

bmorphism:
`rust-optimizer` -> `workspace-optimizer`, uses version `0.12.6`, avoids:

```
... Error during static Wasm validation: Wasm contract doesn't have required export: "instantiate". Exports required by VM: ["allocate", "deallocate", "instantiate"].: create wasm contract failed: invalid request
```


ezekiiel (approved):
Thanks!

janitachalam (approved):

PR #342, opened 2022-06-07
Address issue 3 of the audit report

### LINKED ISSUE #325: 3. Consider checking whether the item key exists before removing
[2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf](https://github.com/DA0-DA0/dao-contracts/files/8848866/2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf)

Check issue 3 of the attached audit.

borgbob:
@Callum-A / @ezekiiel  I was thinking of calling `add_attribute("key_removed", key_removed)` in the response to indicate if the key was actually removed, but it was not clear if it would be useful information

Close #325

### DIFF
--- a/contracts/cw-core/src/contract.rs
+++ b/contracts/cw-core/src/contract.rs
@@ -379,10 +379,14 @@ pub fn execute_remove_item(
         return Err(ContractError::Unauthorized {});
     }
 
-    ITEMS.remove(deps.storage, key.clone());
-    Ok(Response::default()
-        .add_attribute("action", "execute_remove_item")
-        .add_attribute("key", key))
+    if ITEMS.has(deps.storage, key.clone()) {
+        ITEMS.remove(deps.storage, key.clone());
+        Ok(Response::default()
+            .add_attribute("action", "execute_remove_item")
+            .add_attribute("key", key))
+    } else {
+        Err(ContractError::KeyMissing {})
+    }
 }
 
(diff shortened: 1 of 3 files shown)


contracts/cw-core/src/contract.rs
```diff
@@ -379,7 +379,10 @@ pub fn execute_remove_item(
         return Err(ContractError::Unauthorized {});
     }
 
-    ITEMS.remove(deps.storage, key.clone());
+    if ITEMS.has(deps.storage, key.clone()) {
```
borgbob, line 382: I'm new to Rust, so please advise if there's a better way to do this. All this cloning is really strange just to satisfy the compiler, it seems that both `ITEMS.has` and `ITEMS.remove` only need read access to `key`, so why can't we pass it by reference?

ezekiiel: I think unfortunately these clones are needed. The reason we need to clone is because the remove function takes an item of type `K` where `K` is the key being used in the map. If it took `&K` we could pass by reference, but the API tells us that the function wants to take ownership of the key. Under the hood I believe it does this because it serializes the input key into a vec.

borgbob:
@ezekiiel Can you review again?

I think the existing `add_remove_get` test was broken since it was trying to remove a missing key. I've fixed the test and added a new test to verify the error being returned.

ezekiiel (approved):
LG2M. Thanks!

Callum-A (approved):

PR #344, opened 2022-06-07
Update cw-core admin transfer process.

ezekiiel:
The new process is as follows:

1. The current admin nominates a new admin.
2. For the nomination to take effect the nominated admin must execute
   the `AcceptAdminNomination` method on the core contract.
3. At any point the current admin may withdraw their nomination.
4. The current admin may remove the contract's admin entirely without
   any nomination process.

This was suggested by the Oak Security folks and prevents an admin
from accidentally setting the admin of the contract to an invalid
address.

### DIFF
--- a/contracts/cw-core/src/contract.rs
+++ b/contracts/cw-core/src/contract.rs
@@ -53,7 +53,9 @@ pub fn instantiate(
     let admin = msg
         .admin
         .map(|human| deps.api.addr_validate(&human))
-        .transpose()?;
+        .transpose()?
+        // If no admin is specified, the contract is its own admin.
+        .unwrap_or_else(|| env.contract.address.clone());
     ADMIN.save(deps.storage, &admin)?;
 
     let vote_module_msg = msg
@@ -124,7 +126,9 @@ pub fn execute(
         ExecuteMsg::UpdateProposalModules { to_add, to_remove } => {
             execute_update_proposal_modules(deps, env, info.sender, to_add, to_remove)
         }
-        ExecuteMsg::NominateAdmin { admin } => execute_nominate_admin(deps, info.sender, admin),
+        ExecuteMsg::NominateAdmin { admin } => {
+            execute_nominate_admin(deps, env, info.sender, admin)
+        }
         ExecuteMsg::AcceptAdminNomination {} => execute_accept_admin_nomination(deps, info.sender),
         ExecuteMsg::WithdrawAdminNomination {} => {
             execute_withdraw_admin_nomination(deps, info.sender)
@@ -160,19 +164,14 @@ pub fn execute_admin_msgs(
 ) -> Result<Response, ContractError> {
     let admin = ADMIN.load(deps.storage)?;
 
-    match admin {
-        Some(admin) => {
-            // Check if the sender is the DAO Admin
-            if sender != admin {
-                return Err(ContractError::Unauthorized {});
-            }
-
-            Ok(Response::default()
-                .add_attribute("action", "execute_admin_msgs")
-                .add_messages(msgs))
-        }
-        None => Err(ContractError::NoAdmin {}),
+    // Check if the sender is the DAO Admin
+    if sender != admin {
+        return Err(ContractError::Unauthorized {});
     }
+
+    Ok(Response::default()
+        .add_attribute("action", "execute_admin_msgs")
+        .add_messages(msgs))
 }
 
 pub fn execute_proposal_hook(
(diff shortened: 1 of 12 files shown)


Callum-A (approved):
this now feels like turning the keys in sync on a nuclear submarine. i love it

contracts/cw-core/src/msg.rs
```diff
@@ -100,9 +100,26 @@ pub enum ExecuteMsg {
...
-    UpdateAdmin { admin: Option<Addr> },
+    /// Callable by the admin of the contract. If ADMIN is None
+    /// removes the admin from the contract. If ADMIN is Some a new
+    /// admin is proposed and that new admin may become the admin by
+    /// executing the `AcceptAdminNomination` message.
+    ///
+    /// If there is already a pending admin nomination the
+    /// `WithdrawAdminNomination` message must be executed before a
+    /// new admin may be nominated.
+    NominateAdmin { admin: Option<String> },
```
de-husk, line 111: Just wondering why we are switching from an `Addr` to a `String` for the admin address?

ezekiiel: Great question! CosmWasm doesn't actually validate addresses while deserializing them. The result is (before) something like `UpdateAdmin { admin: "foo" }` would actually be a valid message to the contract. This is sub-optimal so we parse into a string so that the type system will make us validate the addresses before we can use them as addresses.

contracts/cw-core/src/contract.rs
```diff
@@ -182,37 +190,76 @@ pub fn execute_proposal_hook(
...
+    let current_nomination = NOMINATED_ADMIN.may_load(deps.storage)?;
+    if current_nomination.is_some() {
+        return Err(ContractError::PendingNomination {});
+    }
 
-            // Save the new DAO Admin (which may be set to None)
-            ADMIN.save(deps.storage, &admin)?;
+    match &nomination {
+        Some(nomination) => NOMINATED_ADMIN.save(deps.storage, nomination)?,
+        None => ADMIN.save(deps.storage, &None)?,
```
de-husk, line 212: When thinking about the non-dao-dao ui usecase, Im a little worried that people could accidentally forget to fill in the `nomination` string with the address and then inadvertently just remove the admin from the dao with no recourse.

I think being explicit about removing the admin with a bool flag or something would be safer.

janitachalam: agreed, we should make it difficult to footgun by removing the admin

de-husk (changes requested):
LGTM

Just one recommendation to make it a bit safer: https://github.com/DA0-DA0/dao-contracts/pull/344/files#r900468002

UPDATE: Also, totally fine with keeping this `Option<String>` interface if we go with bluenote's suggestion about just defaulting the admin to itself when removing, so that we can recover from this mistake

janitachalam (approved):

contracts/cw-core/src/contract.rs
```diff
@@ -160,19 +164,14 @@ pub fn execute_admin_msgs(
...
-                return Err(ContractError::Unauthorized {});
-            }
-
-            Ok(Response::default()
-                .add_attribute("action", "execute_admin_msgs")
-                .add_messages(msgs))
-        }
-        None => Err(ContractError::NoAdmin {}),
+    // Check if the sender is the DAO Admin
+    if sender != admin {
```
de-husk, line 168: Does raw dao have an admin? Im just making sure we are leaving a migration path here.

de-husk: Well, nm, this is the execute_admin msg path, so if we break exectuting admin messages for raw dao that doesnt have admin in the first place, that shouldnt matter.

de-husk: But actually looking down further, this could break nominating an admin: https://github.com/DA0-DA0/dao-contracts/pull/344/commits/01ca0d5cc7db96d6e1c95d0381d7180f819527c5#diff-cc5c48658f12e554fb961f3f264fb1efd28ebd41512da1f65fa27a0339d7df9aR201

de-husk (approved):

contracts/cw-core/src/contract.rs
ezekiiel: Can do something like this if RAW DAO wants to update to the latest contracts: https://github.com/DA0-DA0/dao-contracts/blob/zeke/cw20-stake-migration/contracts/cw20-stake/src/contract.rs#L451-L467

PR #346, opened 2022-06-07
Audit issue 18

borgbob:
I might be pushing my luck here, but it seems  skipping saving when `weight == 0` would work because `query_voting_power_at_height` will return 0 even if the address is missing from storage.

### DIFF
--- a/contracts/cw4-voting/src/contract.rs
+++ b/contracts/cw4-voting/src/contract.rs
@@ -40,9 +40,13 @@ pub fn instantiate(
     let mut total_weight = Uint128::zero();
     for member in initial_members.iter() {
         let member_addr = deps.api.addr_validate(&member.addr)?;
-        let weight = Uint128::from(member.weight);
-        USER_WEIGHTS.save(deps.storage, &member_addr, &weight, env.block.height)?;
-        total_weight += weight;
+        if member.weight > 0 {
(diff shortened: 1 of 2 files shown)


contracts/cw4-voting/src/contract.rs
```diff
@@ -111,12 +113,16 @@ pub fn execute_member_changed_hook(
         } else {
             negative_difference += Uint128::from(old - weight);
         }
-        USER_WEIGHTS.save(
-            deps.storage,
-            &user_address,
-            &Uint128::from(weight),
-            env.block.height,
-        )?;
+        if weight == 0 {
```
ezekiiel, line 116: Could we add an explanatory comment here about why this works? Might be nice to add a similar one to `query_voting_power_at_height` just to make sure someone doesn't unknowingly remove the `unwrap_or_default` there.

ezekiiel (commented):
This is great! Thanks. :)

Could we add a test for instantiation with zero weights and for updates that cause zero weights to be sent to the hook before we land this?

ezekiiel (approved):
Great work! Thanks for putting together the test.

@Callum-A would you mind giving the final stamp / merge here when you have a chance. Know you're more familiar with this code than I am.

Callum-A (approved):

PR #348, opened 2022-06-07
Audit issue 4

### LINKED ISSUE #326: 4. Admin withdrawals will affect the reward distribution
[2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf](https://github.com/DA0-DA0/dao-contracts/files/8848869/2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf)

Check issue 4 of the attached audit.

chinhnotional:
@Callum-A @ezekiiel 
As in the audit written, admin can only withdraw the entire balance, not a partial withdrawal. I solved it by adding an `amount` field in `ExecuteMsg::Withdraw` that allows withdrawal of an amount token (0 < amount <= contract balance). Anh I also add test for invalid amount. Thank you guys for review.
Close #326

### DIFF
--- a/contracts/stake-cw20-reward-distributor/schema/execute_msg.json
+++ b/contracts/stake-cw20-reward-distributor/schema/execute_msg.json
@@ -53,7 +53,15 @@
       ],
       "properties": {
         "withdraw": {
-          "type": "object"
+          "type": "object",
+          "required": [
+            "amount"
+          ],
+          "properties": {
+            "amount": {
+              "$ref": "#/definitions/Uint128"
+            }
+          }
         }
       },
       "additionalProperties": false
--- a/contracts/stake-cw20-reward-distributor/src/contract.rs
+++ b/contracts/stake-cw20-reward-distributor/src/contract.rs
@@ -205,7 +205,7 @@ pub fn execute_withdraw(
 
     let msg = to_binary(&cw20::Cw20ExecuteMsg::Transfer {
         recipient: config.owner.clone().into(),
-        amount: amount,
+        amount,
     })?;
     let send_msg: CosmosMsg = WasmMsg::Execute {
         contract_addr: config.reward_token.into(),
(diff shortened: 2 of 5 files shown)


chinhnotional:
This is my first commit for Dao contract, hope it helps.

ezekiiel:
Hey @chinhnotional ! Thanks so much for the PR. This looks great.

One thing that would be nice to test is what the behavior of the contract is after a partial withdrawal takes place. What we'd expect (I think) is that rewards continue to pay out until the contract runs out of funds. If we could write a test for that it would be great.

chinhnotional:
Yes, i will do it

Callum-A:
Just got to regenerate the schema can do this by running: `./scripts/schema.sh`

Callum-A:
> Sorry for my carelessness, got lint error a few times, I just updated the schema.

All good! WFs require additional approval for new contributors, but we got there! Thanks for your help

Callum-A (approved):

borgbob:
> @ben2x4 would you mind taking a look at this? Code is fine, but just want to make sure we're on the same page about the need for this functionality:
> 
> image
> 
> I have mixed feelings about if we want this functionality. It seems like the only situation where you'd want something like this is if the admin inadvertently sends too much funds to the contract? Would love your thoughts.

Note that this sentence: 

```
After the use of an admin withdrawal min(balance_info.balance, pending_rewards) in contracts/stake-cw20-reward-distributor/src/contract.rs:156 will always result in zero rewards being paid out
```

Is no longer true after #352 is merged with the improvements from #354. With both #352 and #354, zero rewards will result in an error being returned.

So unless you really need to allow partial withdrawals, this PR is overlapping with #352

ezekiiel:
> @borgbob Is no longer true after https://github.com/DA0-DA0/dao-contracts/pull/352 is merged with the improvements from https://github.com/DA0-DA0/dao-contracts/pull/354. With both https://github.com/DA0-DA0/dao-contracts/pull/352 and https://github.com/DA0-DA0/dao-contracts/pull/354, zero rewards will result in an error being returned.

I think this is a great point. I'll send this over to the auditors and see what they think. I suspect that you're right here.

ezekiiel:
Sent this information over to our auditors. Will close this for now pending a response.

Thanks for the PR at @chinhnotional. Sorry we can't land it right now. Hope to see more code / reviews from you in the future!

PR #349, opened 2022-06-07
Don't allow zero reward durations.

### LINKED ISSUE #327: 5. Reward duration should be validated as non-zero value
[2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf](https://github.com/DA0-DA0/dao-contracts/files/8848869/2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf)

Check issue 5 of the attached audit.

ezekiiel:
Resolves #327.

Previously a rewards duration of zero was allowed. This is suboptimal because should the admin set a zero rewards duration there are a few places where we'd run into divide by zero errors. This adds the needed checks to be sure a zero reward duration is not set.

### DIFF
--- a/contracts/cw20-stake-external-rewards/src/contract.rs
+++ b/contracts/cw20-stake-external-rewards/src/contract.rs
@@ -62,6 +62,10 @@ pub fn instantiate(
     };
     CONFIG.save(deps.storage, &config)?;
 
+    if msg.reward_duration == 0 {
+        return Err(ContractError::ZeroRewardDuration {});
+    }
+
     let reward_config = RewardConfig {
         period_finish: 0,
         reward_rate: Uint128::zero(),
@@ -179,6 +183,9 @@ pub fn execute_fund(
         reward_rate: amount
             .checked_div(Uint128::from(reward_config.reward_duration))
             .map_err(StdError::divide_by_zero)?,
+        // As we're not changing the value and changing the value
+        // validates that the duration is non-zero we don't need to
+        // check here.
         reward_duration: reward_config.reward_duration,
     };
 
@@ -374,6 +381,11 @@ pub fn execute_update_reward_duration(
     if reward_config.period_finish > env.block.height {
         return Err(ContractError::RewardPeriodNotFinished {});
     };
+
+    if new_duration == 0 {
+        return Err(ContractError::ZeroRewardDuration {});
+    }
+
     let old_duration = reward_config.reward_duration;
     reward_config.reward_duration = new_duration;
     REWARD_CONFIG.save(deps.storage, &reward_config)?;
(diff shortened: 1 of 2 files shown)


Callum-A (approved):

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -62,6 +62,10 @@ pub fn instantiate(
     };
     CONFIG.save(deps.storage, &config)?;
 
+    if msg.reward_duration == 0 {
```
borgbob, line 65: Maybe this validation could be moved to the top of the function? It would allow avoiding doing a bunch of unnecessary setup when 0 is passed.

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -716,6 +728,55 @@ mod tests {
...
+            },
+            Cw20Coin {
+                address: ADDR2.to_string(),
+                amount: Uint128::new(50),
+            },
+            Cw20Coin {
+                address: ADDR3.to_string(),
+                amount: Uint128::new(50),
+            },
+        ];
```
borgbob, line 749: Just a small nitpick, to reduce a bit the verbosity of the test (we don't need any addresses to verify the instantiate validation), we could use an empty vector here, especially if [this](https://github.com/DA0-DA0/dao-contracts/pull/349/files#r894816126) is done.

I had sent a similar PR for this issue by mistake, here's how I've written the test: https://github.com/DA0-DA0/dao-contracts/pull/355/files

ezekiiel: Wow. Thanks for such a close review. Great point!

contracts/stake-cw20-external-rewards/src/contract.rs
ezekiiel: This is a really interesting point. I think I'm inclined to keep it the way it is for two reasons:

1. Anything could fail first. Changing the order likely only makes sense if we have some heuristic for deciding which error condition is going to fail more frequently.
2. IIRC (@de-husk knows the details on this) in CosmWasm your gas payment is based on gas wanted and not gas used. So, if I budget N gas for this instantiation and it fails early because we reorganize these and uses M gas, I sill pay N gas for the execution.

Appreciate you taking a close look at this and the good feedback! Happy to be convinced otherwise on either of these points.

de-husk: Totally correct about paying for GasWanted, not GasUsed here. 

However, in practice `GasWanted` is usually set automatically by calling the `/simulate` cosmos abci query, and I need to learn more about how cosmwasm gas simulation is done, but it runs a read-only (single node / non-consensus) dry-run of the code using stale data and returns the gas used. So I think you should actually be lowering the gas threshold by exiting early for most people.

JakeHartnell (approved):

PR #350, opened 2022-06-08
Replace TODO comment with explanation of why things are the way they are.

### LINKED ISSUE #336: 16. Outstanding TODO comments are present in the codebase
[2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf](https://github.com/DA0-DA0/dao-contracts/files/8848890/2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf)

Check issue 16 of the attached audit.

ezekiiel:
Replaces TODO comment with an explanation of why the numbers are they
way they are.

### DIFF
--- a/contracts/stake-cw20-external-rewards/src/contract.rs
+++ b/contracts/stake-cw20-external-rewards/src/contract.rs
@@ -829,8 +829,16 @@ mod tests {
         assert_pending_rewards(&mut app, &reward_addr, ADDR2, 2500);
         assert_pending_rewards(&mut app, &reward_addr, ADDR3, 6000);
 
+        // Current height is 1034. ADDR1 is receiving 500 tokens/block
+        // and ADDR2 / ADDR3 are receiving 250.
+        //
+        // At height 101000 99966 additional blocks have passed. So we
+        // expect:
+        //
+        // ADDR1: 5000 + 99966 * 500 = 49,998,000
+        // ADDR2: 2500 + 99966 * 250 = 24,994,000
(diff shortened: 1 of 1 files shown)


Callum-A (approved):

contracts/stake-cw20-external-rewards/src/contract.rs
```diff
@@ -829,8 +829,16 @@ mod tests {
...
 
+        // Current height is 1034. ADDR1 is receiving 500 tokens/block
+        // and ADDR2 / ADDR3 are receiving 250.
+        //
+        // At height 101000 99966 additional blocks have passed. So we
+        // expect:
+        //
+        // ADDR1: 5000 + 99966 * 500 = 49,998,000
+        // ADDR2: 2500 + 99966 * 250 = 24,994,000
+        // ADDR3: 6000 + 99966 * 250 = 24,997,500
```
Samangan, line 840: I think this comment is totally fine and better than not having one. But ideally we wouldn't need a comment and the actual test code would be more descriptive. 

Something like this:
```
blocks_to_add = 99966
height += blocks_to_add
app.borrow_mut().update_block(|b| b.height = height);

assert_pending_rewards(&mut app, &reward_addr, ADDR1, addr1_total_rewards + blocks_to_add * addr1_per_block_rewards);
...
```

ezekiiel: Great point. Let's do this as a followup so we can land this.

Issue #319, opened 2022-06-06
Update deploy local


Callum-A:
Deploy local points to the legacy contracts still, need to update it and maybe rename the existing to `deploy_legacy.sh`

PR #351, opened 2022-06-08
Fix optimized cw_core.wasm storage bug + Fix local dev script for v1 contracts

Samangan:
Closes: https://github.com/DA0-DA0/dao-contracts/issues/319

There is a bug storing the optimized cw-core.wasm. 

When I try to upload cw-core.wasm, I keep getting this error:
```
failed to execute message; message index: 0: Error calling the VM: Error during static Wasm validation: Wasm contract doesn't have required export: "instantiate". Exports required by VM: ["allocate", "deallocate", "instantiate"].: create wasm contract failed: invalid request
```

I tried uploading the unoptimized wasm binary, but it's too large to store on juno with current configurations.

Im going to do more digging into fixing this before opening this PR

UPDATE:  We fixed the cw-core.wasm store bug by removing the dependency cycle in: `cw-core -> debug/cw-proposal-sudo -> cw-core`.


Samangan:
> You are an absolute legend for putting this together. Thank you!
> 
> I've also run into this `cw-core` issue many times. My suspicion is that it is caused by the library version of the wasm file overriding the non-library one. Removing all of the contracts that depend on `cw-core` as a dependency fix this issue and let you build / compile it. It's a bit of a pain though.

Thanks for this info!

Going down this dependency pruning path, I was able to get this working by just removing these 2 dev dependences in `cw-core`: https://github.com/DA0-DA0/dao-contracts/blob/main/contracts/cw-core/Cargo.toml#L52-L53

and then running the workspace-optimizer from root as normal.

Samangan:
>Not sure why these 2 deps break the rust optimization.

Digging deeper it works when only commenting out this dep: 
```
cw-proposal-sudo = { version = "0.1.0", path = "../../debug/cw-proposal-sudo"}
```
I noticed unlike the other one, this one has a cyclic dependency back to `cw-core` as a library, which is maybe why its breaking.

Samangan:
zeke and I paired on removing the dependency cycle, and everything is working now.

bmorphism:
yay! 👀 this was the last missing piece for testing https://github.com/DA0-DA0/dao-contracts/pull/321/files#diff-d0024dcfce8263455f76f35323f55051631264ba689777e927397480a6914ff4R54 -- the deployment / uploading with GitHub Actions using the service wallet pattern should be ready after that

my feedback would be to update the `metadata` in `Cargo.toml` everywhere old optimizer image is referenced to reflect the new build / optimize image being used `workspace-optimizer:0.12.6` e.g. https://github.com/DA0-DA0/dao-contracts/blob/7fb7ce4688e9454d9c06911cacc4d8b5bee4b3c1/contracts/cw-core/Cargo.toml#L24

also, in researching this a bit and trying to eliminate cache, I wanted to point out that `rust-optimizer:0.12.6` also exists, but the key difference is that `workspace-optimizer` uses deterministic order in which to compile the contracts (alphabetic)

we love determinism, so I think it' a good change

bmorphism:
https://github.com/CosmWasm/rust-optimizer#contracts-as-workspace-members

Callum-A (approved):

ezekiiel (approved):

PR #352, opened 2022-06-08
Address audit issue 17: messages with zero rewards are inefficient 

### LINKED ISSUE #337: 17. Messages with zero rewards are inefficient
[2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf](https://github.com/DA0-DA0/dao-contracts/files/8848896/2022-06-06_Audit_Report_-_DAO_DAO_v0.1.pdf)

Check issue 17 of the attached audit.

janitachalam:
Closes #337 -- the issue claims that in contracts/stake-cw20-reward-distributor/src/contract.rs:146, block_diff can be 0 and thus a message will zero rewards will be fired. This actually appears to be handled on line contracts/stake-cw20-reward-distributor/src/contract.rs:162, because if the amount is zero, None will be returned as the msg, but this PR adds a check for block diff <= 0 to circumvent the query on line 153.

### DIFF
--- a/contracts/stake-cw20-reward-distributor/src/contract.rs
+++ b/contracts/stake-cw20-reward-distributor/src/contract.rs
@@ -143,10 +143,10 @@ pub fn validate_staking(deps: Deps, staking_addr: Addr) -> bool {
 fn get_distribution_msg(deps: Deps, env: &Env) -> Result<Option<CosmosMsg>, ContractError> {
     let config = CONFIG.load(deps.storage)?;
     let last_payment_block = LAST_PAYMENT_BLOCK.load(deps.storage)?;
-    let block_diff = env.block.height - last_payment_block;
-    if block_diff == 0 {
+    if last_payment_block >= env.block.height {
         return Ok(None);
     }
+    let block_diff = env.block.height - last_payment_block;
 
     let pending_rewards: Uint128 = config.reward_rate * Uint128::new(block_diff.into());
 
--- a/contracts/stake-cw20-reward-distributor/src/contract.rs
+++ b/contracts/stake-cw20-reward-distributor/src/contract.rs
@@ -144,6 +144,9 @@ fn get_distribution_msg(deps: Deps, env: &Env) -> Result<Option<CosmosMsg>, Cont
     let config = CONFIG.load(deps.storage)?;
     let last_payment_block = LAST_PAYMENT_BLOCK.load(deps.storage)?;
     let block_diff = env.block.height - last_payment_block;
+    if block_diff <= 0 {
+        return Err(ContractError::ZeroRewards {});
+    }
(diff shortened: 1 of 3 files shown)


contracts/stake-cw20-reward-distributor/src/contract.rs
```diff
@@ -144,6 +144,10 @@ fn get_distribution_msg(deps: Deps, env: &Env) -> Result<Option<CosmosMsg>, Cont
     let config = CONFIG.load(deps.storage)?;
     let last_payment_block = LAST_PAYMENT_BLOCK.load(deps.storage)?;
     let block_diff = env.block.height - last_payment_block;
+    if block_diff <= 0 {
+        return Ok(None);
```
borgbob, line 148: The audit report recommendation: `We recommend reverting with an error when block_diff is zero.`

Shouldn't we return an error here?

contracts/stake-cw20-reward-distributor/src/contract.rs
```diff
@@ -240,3 +244,34 @@ fn query_info(deps: Deps, env: Env) -> StdResult<InfoResponse> {
         balance: balance_info.balance,
     })
 }
+
+mod tests {
```
borgbob, line 248: Missing `#[cfg(test)]` above `mod tests`  (this is why there are a bunch of lint warnings about unused imports)

borgbob: It seems there's already a separate file for tests, so you can probably move the new test to `src/tests.rs`

contracts/stake-cw20-reward-distributor/src/contract.rs
```diff
@@ -144,7 +144,7 @@ fn get_distribution_msg(deps: Deps, env: &Env) -> Result<Option<CosmosMsg>, Cont
     let config = CONFIG.load(deps.storage)?;
     let last_payment_block = LAST_PAYMENT_BLOCK.load(deps.storage)?;
     let block_diff = env.block.height - last_payment_block;
-    if block_diff <= 0 {
+    if block_diff == 0 {
```
borgbob, line 147: Isn't it possible that `last_payment_block > env.block.height`? Check [this](https://github.com/DA0-DA0/dao-contracts/pull/352#issuecomment-1151090605) for a suggested fix

contracts/stake-cw20-reward-distributor/src/contract.rs
Callum-A: I'm unsure on erroring here due to the conflicts it could cause when handling config updates

ezekiiel: An error here actually seems like it would be nice. Something along the lines of "rewards have already been distributed for this block".

PR #354, opened 2022-06-10
(Audit issue 17) Zero rewards improvements

borgbob:
- Remove distribution_msg from `execute_update_config`. It doesn't seem   necessary to return that information from `execute_update_config`. (at least I don't see any `execute_update_config` test that depend on it)
- Change get_distribution_msg signature to return   Result<CosmosMsg, ContractError>. Option is no longer required since
  in all cases where we would return None, we are now returning an  Error.
- Remove the direct `get_distribution_msg` test. Instead enhance the  existing public contract API test to verify the new
  RewardsDistributedForBlock scenario.

cc @ezekiiel 

ref: #352

### DIFF
--- a/contracts/stake-cw20-reward-distributor/src/contract.rs
+++ b/contracts/stake-cw20-reward-distributor/src/contract.rs
@@ -93,7 +93,6 @@ pub fn execute_update_config(
         return Err(ContractError::Unauthorized {});
     }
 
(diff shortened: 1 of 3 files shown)


ezekiiel (approved):
Great work! I like this change.

Callum-A (approved):
very nice, cheers for picking up!

Issue #356, opened 2022-06-13
Implement minimum proposal duration in cw-proposal-single.


ezekiiel:
The auditors included this as an extra point this week:

<img width="679" alt="image" src="https://user-images.githubusercontent.com/30676292/173435396-eb112352-5d89-488b-9399-9eca82ede02d.png">

ezekiiel:
I have mixed feelings about this one. Sent the following to the auditors to see if they have some more details:

```
a question re issue (3) from the updated audit report:

The audit suggests that we implement a min voting period for proposals
which is the amount of time that they must be open before they can be
passed. The idea being that this will prevent an attacker from quickly
creating and executing a proposal before DAO members have time to
react. Thinking about this a little, I'm not sure there is actually a
situation where, simultaneously, an attacker has enough voting to pass
a proposal early and the rest of the DAO has enough tokens to do
anything about it.

This is because in DAO DAO a proposal will only ever pass early if it
is guaranteed to pass. i.e. no sequence of votes could make the
proposal fail. For example, with each of the voting thresholds we
support:

- AbsoluteCount: if the attacker has the needed count of tokens,
nobody can stop them.

- AbsoluteThreshold: if the attacker has the needed amount of tokens,
nobody can stop them.

- ThresholdQuorum: an attacker could react quorum alone, but the
proposal will only pass early if the number of tokens put up mean that
no voting could stop it. For example, with a 10% quorum and majority
threshold - if an attacker puts up 10% of tokens for yes the proposal
will not pass either until the majority of voting power has voted yes
or the proposal voting period has expired. If the attacker does not
control the majority of tokens in this situation the voting body will
still have the regular amount of time to mount a response.

This all makes me think that we _could_ implement this min voting
period, but the only effect would be to have a period of time wherein
DAO voters know the DAO will be taken over, but can not actually stop
it. Maybe they could withdraw liquidity from pools during this time to
decrease the amount of damage that is done, but I can't really think
of many other moves that the DAO members could make.

Does this align with what you folks were thinking of for that point?
I'm happy to implement this in the contracts, but want to be sure
we're not adding complexity to the code where it is not needed.
```

PR #357, opened 2022-06-13
Add min_voting_period to cw-proposal-single.

ezekiiel:
This adds a minimum voting period. Having a minimum voting period
allows DAO members some time to respond in the event of a DAO hostile
takeover.

For example, say a DAO exists with some tokens in a LP pool. An
attacker may purchase a large number of those tokens and stake them to
take over the DAO and give the DAO's treasury to themselves. If a
minimum voting period is configured DAO members will have some time to
see this occuring and remove liquidity from the pool to reduce the
amount of damage the attacker might do.

See this issue for some context: https://github.com/DA0-DA0/dao-contracts/issues/356#issuecomment-1154476150

IMO if folks are against this change I'd be happy to just close this PR and
move on. I do think that, on balance, this is probably worth the added
complexity.

### DIFF
--- a/contracts/cw-proposal-single/src/proposal.rs
+++ b/contracts/cw-proposal-single/src/proposal.rs
@@ -13,14 +13,25 @@ use crate::{
 pub struct Proposal {
     pub title: String,
     pub description: String,
+    /// The address that created this proposal.
     pub proposer: Addr,
-
+    /// The block height at which this proposal was created. Voting
+    /// power queries should query for voting power at this block
+    /// height.
     pub start_height: u64,
+    /// The minimum amount of time this proposal must remain open for
+    /// voting. The proposal may not pass unless this is expired or
+    /// None.
+    pub min_voting_period: Option<Expiration>,
+    /// The the time at which this proposal will expire and close for
+    /// additional votes.
     pub expiration: Expiration,
-
+    /// The threshold at which this proposal will pass.
     pub threshold: Threshold,
+    /// The total amount of voting power at the time of this
+    /// proposal's creation.
     pub total_power: Uint128,
-
+    /// The messages that will be executed should this proposal pass.
     pub msgs: Vec<CosmosMsg<Empty>>,
 
     pub status: Status,
@@ -116,6 +127,18 @@ impl Proposal {
         // If re-voting is allowed nothing is known until the proposal
         // has expired.
         if self.allow_revoting && !self.expiration.is_expired(block) {
+            return false;
+        }
+        // If the min voting period is set and not expired the
+        // proposal can not yet be passed. This gives DAO members some
+        // time to remove liquidity / scheme on a recovery plan if a
+        // single actor accumulates enough tokens to unilaterally pass
+        // proposals.
+        if !self
+            .min_voting_period
+            .map(|min| min.is_expired(block))
+            .unwrap_or(true)
+        {
             return false;
         }
 
@@ -269,6 +292,7 @@ mod test {
         votes: Votes,
         total_power: Uint128,
         is_expired: bool,
+        min_voting_period_elapsed: bool,
         allow_revoting: bool,
     ) -> (Proposal, BlockInfo) {
         let block = mock_env().block;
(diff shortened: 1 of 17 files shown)


Callum-A (approved):
Super work

contracts/cw-proposal-single/src/proposal.rs
```diff
@@ -118,6 +129,18 @@ impl Proposal {
...
+        // time to remove liquidity / scheme on a recovery plan if a
+        // single actor accumulates enough tokens to unilaterally pass
+        // proposals.
+        if !self
+            .min_voting_period
+            .map(|min| min.is_expired(block))
+            .unwrap_or(true)
+        {
+            return false;
+        }
```
borgbob, line 143: Not sure if it is because I'm new to Rust and not used to the idioms of testing conditions wrapped into `Option`, but I found this somewhat difficult to read. IMHO, here's an alternative that looks more explicit/easier to understand:

```rust
if let Some(min) = self.min_voting_period {
    if !min.is_expired(block) {
        return false;
    }
}
```

contracts/cw-proposal-single/src/utils.rs
```diff
@@ -24,3 +27,27 @@ pub fn get_total_power(deps: Deps, dao: Addr, height: Option<u64>) -> StdResult<
         .query_wasm_smart(dao, &voting::Query::TotalPowerAtHeight { height })?;
     Ok(response.power)
 }
+
+/// Validates that the min voting period is less than the max voting
+/// period. Passes arguments through the function.
```
borgbob, line 32: I wonder if it would make sense to also have a "maximum minimum voting period" :smile: . Let me explain:

Imagine someone creating a DAO accidentally sets a minimum voting period of years and ends up locking funds in the treasury. Would there be another way to unlock the funds without voting and waiting the minimum period?

I'm far from being an expert, but it is hard to imagine a situation where the minimum voting period must be higher than 1 month (seems to be more than enough time for anyone to exit in case of DAO takeover). So maybe setting `min` to have a maximum value of 1 month would be a good idea to protect against user mistakes.

borgbob (approved):

borgbob:
I like the idea.

I'm still not an expert in the DAODAO contract architecture, but it seems this parameter is added to the `cw-proposal-single` contract which represents a single proposal. (At least that's what I understood from https://github.com/DA0-DA0/dao-contracts/wiki/DAO-DAO-v1-Contracts-Design).

Since new proposal contracts can be added dynamically, I'm wondering if the following scenario is possible:

- Someone buys enough DAO tokens to pass any proposal
- They create a proposal to withdraw funds that doesn't have the minimum voting period set (Assuming the attacker can create proposals with arbitrary parameters)
- Immediately the voting finishes and funds are stolen

So, to make this a security measure against DAO takeover, shouldn't the min voting period be a parameter to the DAO core that is enforced when creating new proposal contracts?

contracts/cw-proposal-single/src/utils.rs
ezekiiel: This is great feedback! I tend to think we should leave this unvalidated. IMO limits like these are often best set in the frontend. Ideally, we want the contracts to be as flexible as possible and the frontend to set sane limits. For example, we currently (in the contracts) allow for passing thresholds that are smaller than 50% - normally, nobody should ever set these and in the frontend we make a fuss about it.

ezekiiel:
@borgbob thanks so much for the comments!

> At least that's what I understood from https://github.com/DA0-DA0/dao-contracts/wiki/DAO-DAO-v1-Contracts-Design

This is a bug in the docs. :) I'll update them! A proposal module can handle more than one proposal. In this case the word "single" refers to the fact that these are single-choice proposals. Soon, there will be proposal modules called things like `cw-proposal-multiple-choice` and things.

contracts/cw-proposal-single/src/proposal.rs
ezekiiel: Great point. :) Was being a little fancy here. Will update. Your suggestion is much more clean.

JakeHartnell (approved):

PR #358, opened 2022-06-14
Authorizations

nicolaslara:
This PR should replace #305 .

The approach here is much simpler with the authorizations integrated into the proposals. We can discuss if this is the right approach, but the middleware approach added so much complexity that I don't think it's worth going that route. 

The most interesting part of this PR should be the `message-filter` contract implementation and its tests. Additionally the proposal tests should give an idea of how this would be used in a more realistic scenario.

The PR is not production-ready, but I'd like to get some feedback before moving forward with the implementation.


authorizations/message-filter/src/contract.rs
```diff
@@ -0,0 +1,340 @@
...
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+#[serde(untagged)]
+pub enum Value {
+    Null,
+    Bool(bool),
+    U128(Uint128),
+    Number(u64), // Why doesn't u128 work here?
```
Callum-A, line 23: u128 doesn't serialise directly thus Uint128, believe this is fixed in some version however

authorizations/message-filter/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
```
Callum-A, line 2: can delete this file

authorizations/message-filter/.gitignore
```diff
@@ -0,0 +1,15 @@
+# Build results
```
Callum-A, line 1: can delete

authorizations/message-filter/schema/execute_msg.json
```diff
@@ -0,0 +1,32 @@
+{
```
Callum-A, line 1: regenerate schema

authorizations/message-filter/src/contract.rs
nicolaslara: I think I tried with `Uint128` but then *all* numbers became strings. Not sure if the solution I have now (numbers behave like numbers but large enough numbers require strings) is nice and user friendly or will just be confusing when it behaves differently for larger ints.

contracts/cw-auth-middleware/src/contract.rs
```diff
@@ -0,0 +1,188 @@
...
+
+fn execute_authorize(
+    deps: DepsMut,
+    msgs: Vec<CosmosMsg<Empty>>,
+    sender: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let auths = AUTHORIZATIONS.load(deps.storage, &config.dao)?;
+
+    if auths.is_empty() {
```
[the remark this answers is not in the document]
nicolaslara: We could remove this. My thought was that we would want the auth contract to be there so that it's easy to configure it by adding and removing rules but that if it isn't configured it would just behave as if it wasn't there, but I understand how it can be confusing.

I actually just realized that removing this would actually leat to the same outcome, as no sub_msg would get added and the `Ok()` would be returned by default. So the solution would be to still check `is_empy()` but return an error. Just slightly different semantics. Happy to change it either way.

authorizations/message-filter/src/contract.rs
```diff
@@ -0,0 +1,340 @@
+use cosmwasm_std::{entry_point, Uint128};
```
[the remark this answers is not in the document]
nicolaslara: Added a basic readme. But I'll add more info there soon

authorizations/message-filter/src/contract.rs
```diff
@@ -0,0 +1,374 @@
...
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn execute(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
+        ExecuteMsg::AddAuthorization { addr, msg } => {
+            let config = CONFIG.load(deps.storage)?;
```
janitachalam, line 108: can we make these blocks into their own functions?

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -252,15 +256,39 @@ pub fn execute_execute(
...
+    //  * Turn this into a query
+    //  * Return unauthorized if the query fails
+    //  * Remove the authorization from the "update" on the sub authorizations
+    //  * Add the update messages anyway but ignore the result
+    //  * Update all the tests to reflect the query changes
+    //  * Figure out a better way to add the auth to this proposal instead of using ReplaceOwner
+    //  * Write proper tests for the auth, and more realistic tests for this proposal where we don't fake the sender
+    //
+    let response = Response::default();
+    let response = if let Some(auths) = AUTHORIZATION_MODULE.may_load(deps.storage)? {
```
janitachalam, line 270: are the performance implications of adding this query to every message negligible?

authorizations/message-filter/src/contract.rs
nicolaslara: done!

nicolaslara:
This is now "feature complete". It currently supports:

 * Ability to restrict the proposals that can be executed via proposal-single (based on the sender and the messages)
 * Adding a cw-auth-middleware contract to cw-core as a proposal. This means we can call `Execute` on cw-auth-middleware and, if the auth passes, have the dao execute the messages in the same block (without the need for voting) 
 * Adding cw-4 groups to cw-auth-middleware so that any member of the group shares the same permissions
 * Three authorization contracts:
     * `message-filter`: The most powerful auth, that accepts a custom message matching pattern and allows or rejects any message that matches the pattern
     * `and`: Authorizes a message if all its children authorize it
     * `whitelist`: Authorizes any message by a whitelisted sender

There are a lot of changes and improvements that I would like to improve here, but I think I should focus on the minimal cleanup needed to be able to ship this and keep adding the improvements in future versions.

For reference, these are some of the things I would like to improve (in no particular order):

 * Rename cw-auth-middlware to cw-auth-manager. It's not a middleware!
 * Document and define the Authorization spec
 * Simplify code reuse. This could be done by:
      * Implementing an Authorization trait that contains default implementations
      * Either reorganize the code so the contract one object (like ng-contracts or Sylvia (https://github.com/CosmWasm/sylvia)) 
      * Or by adding generic functions that can be imported by the different auth implementation
      * Messages in the auth spec may need to be derived in sub auth contracts
 * Split group management into its own contract. This would simplify the implementation of the auth manager. 
 * Add logical auth contracts for composition (any, all, not). This simplifies the construction of complex auths and makes the auth manager cleaner
 * Make it easy to create complex auths via nested instantiate messages (similar to cw-core)
 * Allow the auth manager to specify the default Allow/Reject behaviour (similar to message filter) 
 * Tooling that makes it easy to use auths. Can be done via:
      * Proxy-accounts for keplr
      * JS code that wraps a transaction in the necessary auth boilerplate so that apps can implement the calls themselves
 * Document everything

authorizations/and/src/contract.rs
```diff
@@ -0,0 +1,128 @@
...
+                    },
+                )
+        }
+    }
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::Authorize { msgs, sender } => authorize_messages(deps, env, msgs, sender),
```
JakeHartnell, line 91: Rename to `QueryMsg::IsAuthorized`?

`authorize_messages` sounds like an execute method when it's actually a query handler.

authorizations/message-filter/src/contract.rs
```diff
@@ -0,0 +1,408 @@
...
+#[serde(rename_all = "snake_case")]
+#[serde(untagged)]
+pub enum Value {
+    Null,
+    Bool(bool),
+    U128(Uint128),
+    Number(u64), // Why doesn't u128 work here?
+    String(String),
+    Array(Vec<Value>),
+    Object(Map<String, Value>),
```
ethanfrey, line 27: I am quite worried about using Map in any serde serialisable struct.
Also, very odd to see this come from schemars.

This *may* be safe, but needs some very careful checking.
Minimally, check the json encoding "manually" in test cases.
Also ensure that when this compiles to wasm, it passes `check_contract`
([see this example](https://github.com/CosmWasm/cw-plus/blob/main/.circleci/config.yml#L602-L618))

authorizations/message-filter/src/contract.rs
```diff
@@ -0,0 +1,408 @@
...
+    str_to_value(&serialized)
+}
+
+fn str_to_value(msg: &str) -> Result<Value, ContractError> {
+    from_str(msg).map_err(|_| ContractError::CustomError {
+        val: "invalid str".to_string(),
+    })
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
```
ethanfrey, line 44: I don't think this need to be a contract but a library imported directly by thing using it.

ethanfrey (commented):
This needs more design discussion on a few levels

authorizations/message-filter/src/contract.rs
nicolaslara: Ran the checks and they seem to pass. I can run them on the optimized contract later.
__devel_dao-contracts

I've tested this manually and it works on the chain, and schemars uses https://docs.rs/indexmap/latest/indexmap/ internally to preserve ordering. I can add some more tests to the serialization/deserialization specifically focusing on ordering (is non-determinism your main concern?)

authorizations/and/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "and"
```
JakeHartnell, line 2: Can we think of a better name for this contract? :joy:

nicolaslara: `logical-and`?  Thought maybe it should be something auth specific, like `must-pass-all` or `authorization-joiner-and` 🤷  open to suggestions

authorizations/message-filter/src/contract.rs
nicolaslara: It makes sense to refactor the message parsing and matching into a library. We would still need a message filter contract if we want DAOs to add auths in a generic way (for example: authorize an account to mint NFTs from the DAO account to give to members)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -263,15 +272,55 @@ pub fn execute_execute(
         None => vec![],
     };
 
+    // Check that the proposal is authorized if authorizations are configured
+    let response = Response::default();
+    let response = if let Some(auths) = AUTHORIZATION_MODULE.may_load(deps.storage)? {
```
JakeHartnell, line 277: > That said, I'm not sure how useful it is to be able to add auths to proposal-single. I added that cause it was the easiest way to test initially, but then figured out that making the auth manager behave like a proposal allows for more flexibility.

Ok, getting caught up on this... I think allowing the auth manager to behave like a proposal would be best.

Here's what I envision:
```
SubDAO -------> AuthorizationsContract --------> DAO
```

So the parent DAO would add the Authorizations module as a proposal module. The authorizations contract would then be able to call `ExecuteProposalHook` on the main DAO. The SubDAO would simply vote to call the authorizations contract with the message they intend to execute.

contracts/cw-auth-middleware/src/contract.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    Ok(Response::default()
+        .add_attribute("action", "remove_group")
+        .add_attribute("name", name))
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::Authorize { msgs, sender } => query_authorizations(deps, msgs, sender),
```
JakeHartnell, line 328: Maybe rename to `IsAuthorized`?

contracts/cw-proposal-single/src/contract.rs
nicolaslara: Yeah, I like this approach too.

Just to be clear, this is currently possible (see the tests on cw-auth-middleware), the questions is if this should be the only way of doing it or we should keep the changes to proposal single aswell.

JakeHartnell:
> This is now "feature complete". It currently supports:
> 
> * Ability to restrict the proposals that can be executed via proposal-single (based on the sender and the messages)
> * Adding a cw-auth-middleware contract to cw-core as a proposal. This means we can call `Execute` on cw-auth-middleware and, if the auth passes, have the dao execute the messages in the same block (without the need for voting)
> * Adding cw-4 groups to cw-auth-middleware so that any member of the group shares the same permissions
> * Three authorization contracts:
>   
>   * `message-filter`: The most powerful auth, that accepts a custom message matching pattern and allows or rejects any message that matches the pattern
>   * `and`: Authorizes a message if all its children authorize it
>   * `whitelist`: Authorizes any message by a whitelisted sender

Wow, this is a lot! Very powerful, I like the general direction.

_Less_ might be _more_ here as we will need to get these contracts documented and audited.

Would be nice to simplify things a bit. Think MVP. What things can we remove? IMO the main contracts are the `cw-auth-middleware` and `message-filter`.

> * Split group management into its own contract. This would simplify the implementation of the auth manager.

I would argue groups should not be part of `cw-auth-middleware` because we can achieve the same functionality using `cw1-whitelist`! So it's already done! `cw1-whitelist` can be managed by a group or a DAO.

>   * Implementing an Authorization trait that contains default implementations

This is a great idea!

>   * Messages in the auth spec may need to be derived in sub auth contracts

"Sub Auth contracts seem like something we could put off for a V2.

> * Add logical auth contracts for composition (any, all, not). This simplifies the construction of complex auths and makes the auth manager cleaner
> * Make it easy to create complex auths via nested instantiate messages (similar to cw-core)
> * Allow the auth manager to specify the default Allow/Reject behaviour (similar to message filter)

More things that I would suggest should be out of scope until a V2. Let's build something simple first and do the hard stuff like message filtering well and _then_ think about about how such things can be composed.

> * Tooling that makes it easy to use auths.

Integrating proposal templates and UI for this would be a great start... but again MVP!

This is a really great start though. Happy to setup a call to talk through this stuff. I was initially skeptical about implementing message filter as a separate contract, but I think I'm convinced now.

nicolaslara:
Thanks for the feedback @JakeHartnell! I'm a 100% onboard with simplifying this. 

> I would argue groups should not be part of cw-auth-middleware because we can achieve the same functionality using cw1-whitelist! So it's already done! cw1-whitelist can be managed by a group or a DAO.

I agree they should not be part of cw-auth-middleware. It's also one of the main sources of complexity, so removing it would help. 

The only issue I see with using cw1-whitelist for group auth, is that you have to interact the dao in a different way if you're the one authorized vs if the cw1-whitelist contract is the one authorized, but that may be a tradeoff that is worth it for the simplicity gains. 

Let's discuss this on a call and make a plan for shipping it!

JakeHartnell:
> The only issue I see with using cw1-whitelist for group auth, is that you have to interact the dao in a different way if you're the one authorized vs if the cw1-whitelist contract is the one authorized, but that may be a tradeoff that is worth it for the simplicity gains.

Worth the tradeoff! Getting people to use it will inform next steps. Simple is good. 

> Let's discuss this on a call and make a plan for shipping it!
Let's do it!

nicolaslara:
Started some of the work of removing bloat and generalizing auths as an interface here: https://github.com/nicolaslara/dao-contracts/tree/authorizations_as_interface. Still a WIP, but it should already be closer to an MVP. I want to cleanup the interface and document how to use it. Maybe the easiest way is to just create a template for a minimal auth that can be created via cargo generate to avoid adding more complexity to the Trait

nicolaslara:
another short update here: I've been trying to move everything that is not specific to daodao to its own repo: https://github.com/nicolaslara/authorizations. Most of it is there and the interface is relatively clean/stable. Now I just need to modify the daodao parts to use that interface. I'll try to have this ready by cosmoverse

nicolaslara:
a much simple implementation of this is now at https://github.com/DA0-DA0/dao-contracts/compare/main...nicolaslara:dao-contracts:nico/authorizations?expand=1

This is now only an Authorizations based proposal for dao dao, all the implementations of auths have been moved to a different repo. 

@JakeHartnell @ezekiiel we should discuss this at cosmoverse

0xekez:
i'm going to close this for now as i feel our focus is elsewhere. the work here is very good. lets reopen this when we return to authorizations and have some more coherency about what our design requirements are.

PR #362, opened 2022-06-18
Add migrate method to cw20-stake to allow migration from beta.

ezekiiel:
This adds a migrate method to `cw20-stake` which allows migration from a beta staking contract to a non-beta staking contract. I have tested this migration using a testnet DAO which can be found here: https://testnet.daodao.zone/dao/juno1m4wsr4wj0hz9uz4kavh0yt8v3ghsf347etqzt66rkqprfn5myajqhw3vpd

### DIFF
--- a/contracts/cw20-stake/src/contract.rs
+++ b/contracts/cw20-stake/src/contract.rs
@@ -2,16 +2,16 @@
 use cosmwasm_std::entry_point;
 
 use cosmwasm_std::{
-    from_binary, to_binary, Addr, Binary, Deps, DepsMut, Empty, Env, MessageInfo, Response,
-    StdError, StdResult, Uint128,
+    from_binary, from_slice, to_binary, to_vec, Addr, Binary, Deps, DepsMut, Empty, Env,
+    MessageInfo, Response, StdError, StdResult, Uint128,
 };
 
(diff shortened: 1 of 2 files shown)


Callum-A (approved):

de-husk (approved):

contracts/cw20-stake/src/contract.rs
```diff
@@ -436,6 +436,39 @@ pub fn query_hooks(deps: Deps) -> StdResult<GetHooksResponse> {
     })
 }
 
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
```
JakeHartnell, line 440: This is great! A test would be nice? Though we will do some manual testing.

JakeHartnell (approved):

contracts/cw20-stake/src/contract.rs
ezekiiel: Agree about the test. It's a bit of an operation to test properly in multi-test because we'd need to check in a beta version of the staking contract. Might be worthwhile though to do that and put it in `legacy/` then write some tests going through the whole migration process in cw-core.

PR #363, opened 2022-06-19
Chore: Update @daodao/types to match current contract interfaces

whalelephant:


JakeHartnell:
We need to remove these and switch to @pyramation's code gen work... will make a ticket for that.

JakeHartnell:
Made an issue: https://github.com/DA0-DA0/dao-contracts/issues/364

JakeHartnell:
Thanks for this, but I think we're going to take a different approach than what we've used in the past.

PR #366, opened 2022-06-21
Add cw-token-swap contract.

ezekiiel:
This is an escrow contract for swapping between native and cw20
tokens. The contract is instantiated with two counterparties and their
promised funds. Promised funds may either be native tokens or cw20
tokens. Upon both counterparties providing the promised funds the
transaction is completed and both sides receive their tokens.

At any time before the other counterparty has provided funds a
counterparty may withdraw their funds.

We will use this contract for token swaps between DAO DAO and other
DAOs.


contracts/cw-escrow/src/msg.rs
```diff
@@ -0,0 +1,60 @@
...
+    Cw20 {
+        contract_addr: String,
+        amount: Uint128,
+    },
+}
+
+/// Information about a counterparty in this escrow transaction and
+/// their promised funds.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+pub struct Counterparty {
```
ben2x4, line 23: stylistic point, but I prefer instead of custom TokenInfo using an amount field and the a denom field which is a type from the cw20 package.

contracts/cw-escrow/src/contract.rs
```diff
@@ -0,0 +1,225 @@
...
+    } else {
+        Err(ContractError::InvalidFunds {})
+    }
+}
+
+pub fn execute_withdraw(deps: DepsMut, info: MessageInfo) -> Result<Response, ContractError> {
+    let counterparty_one = COUNTERPARTY_ONE.load(deps.storage)?;
+    let counterparty_two = COUNTERPARTY_TWO.load(deps.storage)?;
+
+    let (counterparty, other_counterparty, storage) = if info.sender == counterparty_one.address {
```
ben2x4, line 174: lets make this a function

contracts/cw-escrow/src/contract.rs
```diff
@@ -0,0 +1,225 @@
...
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
+        ExecuteMsg::Receive(msg) => execute_receive(deps, info.sender, msg),
+        ExecuteMsg::Fund {} => execute_fund(deps, info),
+        ExecuteMsg::Withdraw {} => execute_withdraw(deps, info),
+    }
+}
+
+pub fn execute_receive(
```
ben2x4, line 51: Feels like a lot of duplicated code, can we make execute receive funnel into execute fund?

ben2x4 (commented):
Looks good! Mostly just stylistic nits and think we can remove some code duplication here.

ben2x4:
I also feel like the name get a little confusing with https://github.com/CosmWasm/cw-tokens/tree/main/contracts/cw20-escrow. Maybe we can call it cw-token-swap! 

Would love to push this upstream to the cw-tokens repo eventually!

contracts/cw-escrow/src/contract.rs
ezekiiel: Great suggestion. See `get_counterparty` in the most recent commit. :)

contracts/cw-escrow/src/msg.rs
ezekiiel: `cw20::Denom` is a nice type. My only issue with it is that it isn't safe to use in contract message types as the `Cw20` variant holds an `Addr` so either way we've got to roll our own type here for the message section which removes some of the advantage of being "familiar" to callers. of the contract. I also like with these types that you can name them `Checked..` to indicate that this is a verified version of the type.

Happy to swap it over though if that isn't convincing. To quote @elsehow this is a "strong opinion weakly held".

contracts/cw-token-swap/src/contract.rs
```diff
@@ -0,0 +1,247 @@
...
+    storage: Item<CheckedCounterparty>,
+) -> Result<Response, ContractError> {
+    if paid != expected {
+        return Err(ContractError::InvalidAmount {
+            expected,
+            actual: paid,
+        });
+    }
+
+    if counterparty.provided {
```
janitachalam, line 102: nit: this can be the first check

contracts/cw-token-swap/src/contract.rs
```diff
@@ -0,0 +1,247 @@
...
+pub fn instantiate(
+    deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let counterparty_one = msg.counterparty_one.into_checked(deps.as_ref())?;
+    let counterparty_two = msg.counterparty_two.into_checked(deps.as_ref())?;
```
janitachalam, line 29: do we intend to allow the case where the two counterparties are the same address?

JakeHartnell: Probably good to have a check for that.

contracts/cw-token-swap/src/contract.rs
ezekiiel: Great point. Likely the more relevant of the two errors. :)

contracts/cw-token-swap/src/contract.rs
ezekiiel: Done. Great catch!

janitachalam (approved):

contracts/cw-token-swap/src/msg.rs
```diff
@@ -0,0 +1,60 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::state::CheckedCounterparty;
+
+/// Information about the token being used on one side of the escrow.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum TokenInfo {
+    /// A native token.
+    Native { denom: String, amount: Uint128 },
```
Callum-A, line 12: everyday i pray for TokenFactory

Callum-A, line 12: Also isn't this just a duplicate of the same code in `state.rs` why not use that here

Callum-A (approved):

contracts/cw-token-swap/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    msg::{Counterparty, TokenInfo},
+    ContractError,
+};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum CheckedTokenInfo {
```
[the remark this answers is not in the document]
ezekiiel: Yep! All are supported.

contracts/cw-token-swap/src/msg.rs
ezekiiel: lol me too I think I've written this code like 100 times.

ezekiiel: One is a checked and one is not. The checked edition uses `Addr` instead of `String`. Pretty similar, but I like separating them like this. :)

PR #370, opened 2022-06-22
add cw-admin-factory contract

### LINKED ISSUE #359: Write factory contract that can instantiate a contract with the contract as its own admin.
This contract should likely just take a `Binary` message and a code ID and:

1. Instantiate an instance of the contract with itself as the admin.
2. Do a `WasmMsg::UpdateAdmin` to make the contract its own admin.

janitachalam:
closes #359

### DIFF
--- a/contracts/cw-admin-factory/src/msg.rs
+++ b/contracts/cw-admin-factory/src/msg.rs
@@ -1,0 +1,22 @@
+use cosmwasm_std::Binary;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+pub struct InstantiateMsg {}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    /// Instantiates the target contract with the provided instantiate message and code id and
+    /// updates the contract's admin to be itself.
+    InstantiateContractWithSelfAdmin {
+        instantiate_msg: Binary,
+        code_id: u64,
+        label: String,
+    },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {}
--- a/contracts/cw-admin-factory/.cargo/config
+++ b/contracts/cw-admin-factory/.cargo/config
@@ -1,0 +1,4 @@
+[alias]
+wasm = "build --release --target wasm32-unknown-unknown"
+unit-test = "test --lib"
+schema = "run --example schema"
--- a/contracts/cw-admin-factory/.gitignore
+++ b/contracts/cw-admin-factory/.gitignore
@@ -1,0 +1,16 @@
+# Build results
+/target
+
+# Cargo+Git helper file (https://github.com/rust-lang/cargo/blob/0.44.1/src/cargo/sources/git/utils.rs#L320-L327)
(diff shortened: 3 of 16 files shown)


contracts/cw-core-factory/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+        proposal_modules_instantiate_info: vec![ModuleInstantiateInfo {
+            code_id,
+            msg: to_binary(&instantiate).unwrap(),
+            admin: Admin::CoreContract {},
+            label: "voting module".to_string(),
+        }],
+        initial_items: None,
+    };
+
+    // multi-test does not support UpdateAdmin yet :(
```
janitachalam, line 68: This is unfortunate. Unsure what to do about this other than relying on manual testing until the support is added.

contracts/cw-core-factory/src/contract.rs
```diff
@@ -0,0 +1,85 @@
...
+    instantiate_msg: Binary,
+    code_id: u64,
+) -> Result<Response, ContractError> {
+    // Instantiate the cw-core contract with factory as the admin.
+    let instantiate = WasmMsg::Instantiate {
+        admin: Some(env.contract.address.to_string()),
+        code_id,
+        msg: instantiate_msg,
+        funds: vec![],
+        label: "cw-core contract".to_string(),
```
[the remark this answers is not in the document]
janitachalam: I think that's a great idea.

contracts/cw-core-factory/src/tests.rs
ezekiiel: Made an issue in cw-plus: https://github.com/CosmWasm/cw-plus/issues/744

contracts/cw-core-factory/src/tests.rs
```diff
@@ -0,0 +1,113 @@
...
+                code_id: cw_core_code_id,
+            },
+            &[],
+        )
+        .unwrap_err();
+
+    assert!(matches!(
+        err.downcast().unwrap(),
+        cw_multi_test::error::Error::UnsupportedWasmMsg(_)
+    ))
```
ezekiiel, line 112: Until cw-plus supports this it might be nice to test this method on a non integration test way just using `mock_env` and then likes, calling the method directly, and then verifying that the returned response is what we expect. Here's an example of doing that from cw-plus: https://github.com/CosmWasm/cw-plus/blob/14f4e922fac9e2097a8efa99e5b71d04747e340a/contracts/cw20-base/src/contract.rs#L623-L638

contracts/cw-core-factory/src/tests.rs
janitachalam: thank you!

contracts/cw-admin-factory/src/msg.rs
```diff
@@ -0,0 +1,20 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+pub struct InstantiateMsg {}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    InstantiateContract {
```
ezekiiel, line 11: Could we add a docstring to this message type to explain what it does? Eventually, we'll package up these contracts and publish them as cargo crates and the docs will be nice for folks looking at them.

ezekiiel (approved):
This looks great! No feedback on the functionality. Would be great to add a test using the `mock_env` and etc. functions, but not a big deal if you alternatively want to test this on chain and then link to the transaction or something.

ezekiiel:
Need to rebase on main for typescript to work.

Issue #369, opened 2022-06-22
Rename `Vote` query to `GetVote` in cw-proposal-single.


ezekiiel:
Having a query message and execute message variant with the same name causes trouble in the generated typescript code.

This is suboptimal as if the typescript code generation code would need to have a separate execute and query client if it were to support identically named variants.

@Callum-A also is a fan of the `GetVote` name. :)

PR #347, opened 2022-06-07
Switch to using cosmwasm-typescript-gen

de-husk:
I wanted to get the latest ibc interface changes. Looks like there are other changes here as well. 

closes: https://github.com/DA0-DA0/dao-contracts/issues/364

PR #371, opened 2022-06-23
Rename QueryMsg::Vote -> QueryMsg::GetVote since it causes the typescript generation to break

de-husk:
closes: https://github.com/DA0-DA0/dao-contracts/issues/369

I also confirmed that [the generated typescript files](https://github.com/DA0-DA0/dao-contracts/pull/347) are now `yarn build`ing after these changes.

### DIFF
--- a/contracts/cw-proposal-single/examples/schema.rs
+++ b/contracts/cw-proposal-single/examples/schema.rs
@@ -56,4 +56,5 @@ fn main() {
         "ProposalHooksResponse",
     );
     export_schema_with_title(&schema_for!(HooksResponse), &out_dir, "VoteHooksResponse");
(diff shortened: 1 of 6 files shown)


de-husk:
We discussed this with @pyramation in discord, #smart-contracts. We decided to do this for now, but fix it for real later. That work is already tracked here: https://github.com/CosmWasm/cosmwasm-typescript-gen/issues/26

pyramation:
@JakeHartnell we should likely pair and/or jam on ideas on how to do the proper solution on the TS side. I think it would actually require that we separate not only the classes but also the types for execute and query types as completely separate builds. I also agree others will likely run into this, but for now it's been only a few.

ezekiiel (approved):
Thank you!

Callum-A (approved):

PR #378, opened 2022-06-29
Add timestamps to proposals

### LINKED ISSUE #343: support timestamps in proposals
Persist and surface the timestamp of transactions in a proposal (e.g. when it was created, passed, executed) will require some UI work as well

janitachalam:
Adds a timestamp for created and last updated, in unix epoch time

closes #343


contracts/cw-proposal-single/src/proposal.rs
```diff
@@ -118,6 +124,7 @@ impl Proposal {
     /// Sets a proposals status to its current status.
     pub fn update_status(&mut self, block: &BlockInfo) {
         self.status = self.current_status(block);
+        self.last_updated = block.time
```
ezekiiel, line 127: Think we need to do this in `execute_close` and `execute_execute` as well if we want to count those as "updates".

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -277,6 +282,10 @@ pub fn execute_execute(
         old_status.to_string(),
         prop.status.to_string(),
     )?;
+
+    // Update proposal's last updated timestamp.
+    prop.last_updated = env.block.time;
+
```
ezekiiel, line 288:
```suggestion
```

Any particular reason to do this twice in `execute_execute` here?

ezekiiel (commented):
One small thing and then I say :shipit:

contracts/cw-proposal-single/src/contract.rs
janitachalam: oh god, clearly I was writing this under heavy influence of brain fog

PR #383, opened 2022-07-05
Basic Native Token Voting Contract

Callum-A:
Basic native voting contract which assumes the denom provided is not the staking denom on the network. If we want DAOs to use the native staking token another contract will need to be written. This is for tokens such as ION or IBCed tokens e.g. PageDAO's ERC 20 (via Gravity Bridge)


contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,287 @@
...
+    info: MessageInfo,
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    STAKED_BALANCES.update(
+        deps.storage,
+        &info.sender,
+        env.block.height,
+        |balance| -> StdResult<Uint128> { Ok(balance.unwrap_or_default().checked_sub(amount)?) },
```
janitachalam, line 107: maybe map the overflow error to a better user-facing error about unstaking amount

contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,287 @@
...
+    duration: Option<Duration>,
+) -> Result<Response, ContractError> {
+    let new_owner = new_owner
+        .map(|new_owner| deps.api.addr_validate(&new_owner))
+        .transpose()?;
+    let new_manager = new_manager
+        .map(|new_manager| deps.api.addr_validate(&new_manager))
+        .transpose()?;
+    let mut config: Config = CONFIG.load(deps.storage)?;
+    if Some(info.sender.clone()) != config.owner && Some(info.sender.clone()) != config.manager {
```
janitachalam, line 163:
```suggestion
    if Some(info.sender) != config.owner || Some(info.sender) != config.manager {
```

also this check can be moved before the address validation of new owner / manager

contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,287 @@
...
+                .unwrap_or_else(|| "None".to_string()),
+        ))
+}
+
+pub fn execute_claim(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+) -> Result<Response, ContractError> {
+    let release = CLAIMS.claim_tokens(deps.storage, &info.sender, &env.block, None)?;
```
janitachalam, line 199: should we implement the cap via config, or leave it for now?

Callum-A: leave for now, this approach has been fine for cw20-stake on mainnet

janitachalam (approved):

contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,299 @@
...
+    let info = cw2::get_contract_version(deps.storage)?;
+    to_binary(&cw_core_interface::voting::InfoResponse { info })
+}
+
+pub fn query_dao(deps: Deps) -> StdResult<Binary> {
+    let dao = DAO.load(deps.storage)?;
+    to_binary(&dao)
+}
+
+pub fn query_config(deps: Deps) -> StdResult<GetConfigResponse> {
```
ezekiiel, line 281: How would you feel about just returning `Config` from this method? In the past we've had some issues where we've added a field to a config struct and then forgot to add it to the query response. I tend to think just returning the config object mitigates some of that.

contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,299 @@
...
+    let manager = match msg.manager {
+        Some(manager) => Some(deps.api.addr_validate(&manager)?),
+        None => None,
+    };
+
+    let config = Config {
+        owner,
+        manager,
+        denom: msg.denom,
+        unstaking_duration: msg.unstaking_duration,
```
ezekiiel, line 41: How would you feel about making it not allowed to set an unstaking duration of zero (time or blocks)? We currently allow that in cw20-stake which I think was a bit of a mistake TBH.

Will add this to cw721-stake as well if you think it's a reasonable idea. :)

contracts/cw-native-staked-balance-voting/src/msg.rs
```diff
@@ -0,0 +1,53 @@
...
+        duration: Option<Duration>,
+    },
+    Claim {},
+}
+
+#[voting_query]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
+    Dao {},
```
ezekiiel, line 38: This is cool that you have this query. :)

contracts/cw-native-staked-balance-voting/src/tests.rs
```diff
@@ -0,0 +1,806 @@
...
+            owner: Some(DAO_ADDR.to_string()),
+            manager: Some(ADDR1.to_string()),
+            unstaking_duration: Some(Duration::Height(5)),
+            denom: DENOM.to_string(),
+        }
+    )
+}
+
+#[test]
+fn test_voting_power_queries() {
```
ezekiiel, line 703: Awesome job on all of these tests.

contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,311 @@
...
+        Some(owner) => Some(deps.api.addr_validate(&owner)?),
+        None => None,
+    };
+
+    let manager = match msg.manager {
+        Some(manager) => Some(deps.api.addr_validate(&manager)?),
+        None => None,
+    };
+
+    validate_duration(msg.unstaking_duration)?;
```
ezekiiel, line 55: Crazy to me that `Duration` is trivially copyable like this. Gets me every time.

contracts/cw-native-staked-balance-voting/src/contract.rs
```diff
@@ -0,0 +1,311 @@
...
+        owner,
+        manager,
+        denom: msg.denom,
+        unstaking_duration: msg.unstaking_duration,
+    };
+
+    CONFIG.save(deps.storage, &config)?;
+    DAO.save(deps.storage, &info.sender)?;
+
+    Ok(Response::new().add_attribute("action", "instantiate"))
```
ezekiiel, line 67: How would you feel about adding a little info here about how the contract was instantiated? Nice to see this when you query the TX that instantiates.

ezekiiel (approved):
Amazing work as always. :) Just some little thoughts.

PR #385, opened 2022-07-06
Failure to execute a proposal ought to close it

### LINKED ISSUE #232: Failure to execute a proposal ought to close it.
This should be configurable on/off. @ben2x4's example scenario:

1. I make a proposal to burn some tokens.
2. Proposal passes.
3. DAO treasury does not have enough tokens so the proposal may not be executed.
4. Proposal remains open and executable.
5. A long time later: the DAO treasury now has tokens.
6. Someone executes the proposal and everyone is sad. :(

Buckram123:
Closes #232
Not sure if this feature should be configurable per proposal, or the entire contract.


contracts/cw-proposal-single/src/contract.rs
```diff
@@ -715,14 +727,16 @@ pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response,
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractError> {
-    if msg.id % 2 == 0 {
+    if msg.id == FAILED_PROPOSAL_REPLY_ID {
+        Ok(Response::new())
+    } else if msg.id % 2 == 0 {
         // Proposal hook so we can just divide by two for index
-        let idx = msg.id / 2;
+        let idx = msg.id / 2 - 1;
```
janitachalam, line 734: just curious, why did you have to change this?

Buckram123: I had to reserve some reply id where we don't delete hook, decided it can be 0, is there any better way?

Buckram123:
> thanks for the PR! I'm also wondering if it makes sense to just have this be the default behavior vs. making it configurable. Also re: configurability, we usually do this on the module-level rather than the proposal-level

So it's better to make enabled by default, but reconfigurable later?

contracts/cw-proposal-single/src/contract.rs
ezekiiel: Even / odd worked really well when there were only two options.

One thing we might try is some bitmasking. Could make the first couple bits in the ID an identifier for what kind of reply we're dealing with.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -715,14 +727,16 @@ pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response,
...
         PROPOSAL_HOOKS.remove_hook_by_index(deps.storage, idx)?;
         Ok(Response::new())
     } else {
         // Vote hook so we can minus one then divid by two for index
-        let idx = (msg.id - 1) / 2;
+        let idx = (msg.id - 1) / 2 - 1;
```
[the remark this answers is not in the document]
Buckram123: Hey, no substracting, since I changed index to start at 1 here: https://github.com/DA0-DA0/dao-contracts/pull/385/files#diff-a9c3c187cde5e8b406a240b2191e7c0e0cb51d1a52b4e3f78e37a23c6c24a481L38-R39
But you are right, it is redundant, since I only needed to reserve `0`, not `0` and `1`. Thanks for noticing it!

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -715,14 +727,16 @@ pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response,
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractError> {
-    if msg.id % 2 == 0 {
+    if msg.id == FAILED_PROPOSAL_REPLY_ID {
+        Ok(Response::new())
```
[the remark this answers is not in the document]
Buckram123: I like it, should I add some information to Response(or Responses) as well?

packages/voting/src/proposal.rs
```diff
@@ -6,6 +6,14 @@ use crate::{deposit::CheckedDepositInfo, status::Status};
 pub const DEFAULT_LIMIT: u64 = 30;
 pub const MAX_PROPOSAL_SIZE: u64 = 30_000;
 
+/// Masks for reply id
+pub const FAILED_PROPOSAL_EXECUTION_MASK: u64 = 0b00;
+pub const FAILED_PROPOSAL_HOOK_MASK: u64 = 0b01;
+pub const FAILED_VOTE_HOOK_MASK: u64 = 0b11;
+
+pub const BITS_RESERVED_FOR_REPLY_TYPE: u8 = 2;
+pub const REPLY_TYPE_MASK: u64 = (1 << BITS_RESERVED_FOR_REPLY_TYPE) - 1;
```
ezekiiel, line 15: <3

ezekiiel (commented):
This fucking rocks. Awesome job! Truly this is so cool.

Two things before we merge this:

1. How would you feel about moving some of these masking and tagging operations into functions? Might help make the code more readable where they are used because you won't need to remind yourself how bitwise operations work.
2. Being extremely pedantic here but these are smart contracts, so: could we do a check when creating new proposals that it wouldn't make the proposal ID so large that our masking causes bits to get pushed off?

Thanks so much! Sorry for being so slow to get to reviewing this.

contracts/cw-proposal-single/src/contract.rs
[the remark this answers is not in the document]
ezekiiel: Yes! That would be great.

ezekiiel:
Actually, re (2) here no idea what to do about it. Seems extremely unlikely that this will ever happen and if someone gets to that point they're hosed anyway. Ignore that one. :)

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -335,6 +339,7 @@ pub fn execute_execute(
         VoteResult::SingleWinner(winning_choice) => {
             let response = match winning_choice.msgs {
                 Some(msgs) => {
+                    let res = Response::default();
```
ezekiiel, line 342: nit: imo code is a little easier to read if we remove this and just construct a `Response::default()` when returning. Otherwise, the reader has to read up and see what else has been added.

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -696,16 +713,25 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
 pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractError> {
-    if msg.id % 2 == 0 {
-        // Proposal hook so we can just divide by two for index
-        let idx = msg.id / 2;
-        PROPOSAL_HOOKS.remove_hook_by_index(deps.storage, idx)?;
-    } else {
-        // Vote hook so we can minus one then divid by two for index
-        let idx = (msg.id - 1) / 2;
-        VOTE_HOOKS.remove_hook_by_index(deps.storage, idx)?;
+    let repl = TaggedReplyId::new(msg.id);
```
ezekiiel, line 716: LOVE this abstaction. Great work!

contracts/cw-proposal-multiple/src/tests.rs
```diff
@@ -3534,6 +3562,7 @@ fn test_active_threshold_none() {
     let max_voting_period = cw_utils::Duration::Height(6);
     let instantiate = InstantiateMsg {
         min_voting_period: None,
+        close_proposal_on_execution_failure: true,
```
ezekiiel, line 3565: Would be great to add a test for this to this file / module. Know it's pretty much the same code as cw-proposal-single, but it'll help prevent future regressions.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -263,15 +267,24 @@ pub fn execute_execute(
...
-        let execute_message = WasmMsg::Execute {
-            contract_addr: config.dao.to_string(),
-            msg: to_binary(&cw_core::msg::ExecuteMsg::ExecuteProposalHook { msgs: prop.msgs })?,
-            funds: vec![],
-        };
-        Response::<Empty>::default().add_message(execute_message)
-    } else {
-        Response::default()
+    let response = {
+        let res = Response::default();
```
ezekiiel, line 271: Same as earlier nit: my preference is to move this into the `match` statement below. What do you think about this?

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -714,15 +729,23 @@ pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response,
...
-        VOTE_HOOKS.remove_hook_by_index(deps.storage, idx)?;
-        Ok(Response::new())
+    let repl = TaggedReplyId::new(msg.id);
+    match repl {
+        TaggedReplyId::FailedProposalExecution(proposal_id) => {
+            let mut prop = PROPOSALS
+                .may_load(deps.storage, proposal_id)?
+                .ok_or(ContractError::NoSuchProposal { id: proposal_id })?;
+            prop.status = Status::ExecutionFailed;
+            PROPOSALS.save(deps.storage, proposal_id, &prop)?;
```
ezekiiel, line 739: Instead of doing `may_load` followed by `save` `cosmwasm_std::Map` has an `update` method for these situations.

Example of doing something similar: https://github.com/DA0-DA0/dao-contracts/blob/6dc92d9cba7f00d570f97fbceadd334fb61d6ccf/contracts/cw-proposal-single/src/contract.rs#L319-L346

ezekiiel: Can do the same in `cw-proposal-multiple`.

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -4103,3 +4137,238 @@ fn test_timestamp_updated() {
...
+    // Execute proposal
+    app.execute_contract(
+        Addr::unchecked(CREATOR_ADDR),
+        govmod_single.clone(),
+        &ExecuteMsg::Execute { proposal_id: 3 },
+        &[],
+    )
+    .expect_err("Should be sub overflow");
+
+    // Status should have changed to 'Executed'
```
ezekiiel, line 4366:
```suggestion
    // Status should still be passed
```

packages/voting/src/reply.rs
```diff
@@ -0,0 +1,46 @@
...
+const FAILED_PROPOSAL_EXECUTION_MASK: u64 = 0b00;
+const FAILED_PROPOSAL_HOOK_MASK: u64 = 0b01;
+const FAILED_VOTE_HOOK_MASK: u64 = 0b11;
+
+const BITS_RESERVED_FOR_REPLY_TYPE: u8 = 2;
+const REPLY_TYPE_MASK: u64 = (1 << BITS_RESERVED_FOR_REPLY_TYPE) - 1;
+
+/// Since we can only pass `id`, and we need to perform different actions in reply,
+/// we decided to take few bits to identify "Reply Type".
+/// See <https://github.com/DA0-DA0/dao-contracts/pull/385#discussion_r916324843>
```
ezekiiel, line 11: Love this comment.

packages/voting/src/reply.rs
```diff
@@ -0,0 +1,46 @@
...
+    ///
+    /// We know it costs extra pattern match, but cleaner code in `reply` Methods
+    pub fn new(id: u64) -> Self {
+        let reply_type = id & REPLY_TYPE_MASK;
+        let id = id >> BITS_RESERVED_FOR_REPLY_TYPE;
+        match reply_type {
+            FAILED_PROPOSAL_EXECUTION_MASK => TaggedReplyId::FailedProposalExecution(id),
+            FAILED_PROPOSAL_HOOK_MASK => TaggedReplyId::FailedProposalHook(id),
+            FAILED_VOTE_HOOK_MASK => TaggedReplyId::FailedVoteHook(id),
+            _ => unreachable!(),
```
ezekiiel, line 30: How would you feel about returning an error here?

Could make an error struct:

```rust
pub enum TagError {
  #[error("Unknown reply id ({id}).")]
  UnknownReplyId { id: u64 }
}
```

Then, in the contract error for our proposal modules we can add a variant like:

```rust
    #[error("{0}")]
   Tag(#[from] TagError),
```

And things like `?` will work between the two.

This purely a UX thing. Outcome for caller is the same.

ezekiiel (commented):
This is getting so close. Thanks so much for all your follow up work here.

The last thing we need to add is the ability to close proposals that are in the `ExecutionFailed` state. This way folks can still get proposal deposits back.

To do this you'll just have to change this line:

```rust
    if prop.status != Status::Rejected {
        return Err(ContractError::WrongCloseStatus {});
    }
```

in the `execute_close` methods for both modules.

I can push commits to patch up the CI failures happening here before merging so feel free to ignore those and focus on the interesting stuff. :)

contracts/cw-proposal-multiple/src/contract.rs
Buckram123: Thanks, was thinking if I use `#[inline]` attribute for `new()`  it can optimize it into single match? _But i could be wrong here_ .Now it doesn't matter since we return `Result` now 😆.

contracts/cw-proposal-single/src/contract.rs
Buckram123: Yep, forgot about it for some reason)
Thanks.

packages/voting/src/reply.rs
Buckram123: Not sure what we want to return in error, we have three fine options, I guess: 

1. `Reply.id`(pre-tagged)
2. `reply_type`(tag)
3. Both `reply_type` and id after bit shift

contracts/cw-proposal-single/src/tests.rs
Buckram123: Thanks

packages/voting/src/reply.rs
ezekiiel: Great! Thanks!

packages/voting/src/reply.rs
```diff
@@ -0,0 +1,58 @@
...
+/// Since we can only pass `id`, and we need to perform different actions in reply,
+/// we decided to take few bits to identify "Reply Type".
+/// See <https://github.com/DA0-DA0/dao-contracts/pull/385#discussion_r916324843>
+pub enum TaggedReplyId {
+    FailedProposalExecution(u64),
+    FailedProposalHook(u64),
+    FailedVoteHook(u64),
+}
+
+impl TaggedReplyId {
```
janitachalam, line 18: really awesome stuff here

janitachalam (approved):
Really awesome work! I went through it closely and couldn't find any issues. The only nit I would have is to add a mock test which calls reply directly just to verify that logic, [similar to this one](https://github.com/DA0-DA0/dao-contracts/blob/992d201b303d42d568391c51c0646e28a44199b2/contracts/cw-admin-factory/src/tests.rs#L126)

contracts/cw-proposal-multiple/src/msg.rs
```diff
@@ -26,6 +26,9 @@ pub struct InstantiateMsg {
     /// proposal. None if there is no deposit requirement, Some
     /// otherwise.
     pub deposit_info: Option<DepositInfo>,
+    /// Failed proposal executions will be marked failed,
+    /// instead of reverted state, if this feature is enabled
```
ezekiiel, line 30:
```suggestion
    /// If set to true proposals will be closed if their execution
    /// fails. Otherwise, proposals will remain open after execution
    /// failure. For example, with this enabled a proposal to send 5
    /// tokens out of a DAO's treasury with 4 tokens would be closed when
    /// it is executed. With this disabled, that same proposal would
    /// remain open until the DAO's treasury was large enough for it to be
    /// executed.
```

contracts/cw-proposal-multiple/src/msg.rs
```diff
@@ -86,6 +89,9 @@ pub enum ExecuteMsg {
         /// Information about the deposit required to make a
         /// proposal. None if no deposit, Some otherwise.
         deposit_info: Option<DepositInfo>,
+        /// Failed proposal executions will be marked failed,
+        /// instead of reverted state, if this feature is enabled
```
ezekiiel, line 93: Same suggestion as the doc comment on the `InstantiateMsg`.

contracts/cw-proposal-multiple/src/state.rs
```diff
@@ -34,6 +34,9 @@ pub struct Config {
     /// Information about the depost required to create a
     /// proposal. None if no deposit is required, Some otherwise.
     pub deposit_info: Option<CheckedDepositInfo>,
+    /// Failed proposal executions will be marked failed,
+    /// instead of reverted state, if this feature is enabled
```
ezekiiel, line 38: Doc comment suggestion here as well if you'd like. :)

contracts/cw-proposal-single/src/msg.rs
```diff
@@ -32,6 +32,9 @@ pub struct InstantiateMsg {
     /// proposal. None if there is no deposit requirement, Some
     /// otherwise.
     pub deposit_info: Option<DepositInfo>,
+    /// Failed proposal executions will be marked failed,
+    /// instead of reverted state, if this feature is enabled
```
ezekiiel, line 36: Same comments for here, `ExecuteMsg`, and `Config` on the docstring if you'd like to take the suggested one.

ezekiiel (approved):
LFG. One suggestion for the docstring then lets merge this!

packages/voting/src/reply.rs
Buckram123: Thanks!

JakeHartnell (approved):

packages/voting/src/reply.rs
```diff
@@ -56,3 +57,33 @@ pub mod error {
...
+            TaggedReplyId::FailedProposalExecution(proposal_id)
+        );
+        assert_eq!(
+            TaggedReplyId::new(m_proposal_hook_idx).unwrap(),
+            TaggedReplyId::FailedProposalHook(proposal_hook_idx)
+        );
+        assert_eq!(
+            TaggedReplyId::new(m_vote_hook_idx).unwrap(),
+            TaggedReplyId::FailedVoteHook(vote_hook_idx)
+        );
```
ezekiiel, line 87: Love having little tests like this.

Buckram123: Same, simple test, that shows how to use this feature.

PR #394, opened 2022-07-13
[WIP] Gas Benchmarks

de-husk:
* Uses [Cosm-Orc](https://github.com/de-husk/cosm-orc) to create gas benchmarks that will run in CI.
* Uses [Cosm-orc-gas-diff github action](https://github.com/marketplace/actions/cosm-orc-gas-diff-github-action) to post the PR's gas benchmark results as a PR comment

As I was starting to add more benchmarks, I kept finding myself wanting to validate my assumptions at each step. Essentially making integration tests. Since we want to have integration tests as well, and we want a high test coverage, I think we should just benchmark the integration tests rather than have separate benchmarks. We can still output the gas diff still. 

Thoughts?

------

Also, I put the detailed report in an expandable spoiler. The github action will only post above the spoiler if there is a diff of more / less than 0.5%. For example: https://github.com/de-husk/dao-contracts/pull/3

The report contains the file name / line number of the originating call to cosm-orc, so that we can easily associate gas usage reports with the actual smart contract operation parameters. 

For example:
-  cw_core::msg::ExecuteMsg__src/main.rs:124[6]:
    - GasUsed: 130069
    - GasWanted: 173586
    - File: src/main.rs:124
  

This gas report is for this  `cw_core` ExecuteMsg message: 

https://github.com/DA0-DA0/dao-contracts/blob/29cc35a7bc87da3ad5944daa0180ba60e91549b0/ci/gas_benchmarks/src/main.rs#L117-L119

The process message call was on line 124, and it is the 6th (zero indexed) msg passed.


de-husk:
Closing: see: https://github.com/DA0-DA0/dao-contracts/issues/403

Issue #403, opened 2022-07-17
Add Integration Tests


de-husk:
It would be really nice to have integration tests run in CI that hit a real juno node.  Ideally it would be a local juno node running in CI so that we don't have to worry about figuring out how to refund a wallet.  It's also useful to easily see how a PR will affect gas usage of our critical path.

Instead of having 2 separate sets of gas profiler code and integration test code, I think we should just profile the integration tests and aim to have a high test coverage of our important functionality.

- [x] Make an integration test package that will run in PR and main commits using [cosm-orc](https://github.com/de-husk/cosm-orc)
- [x] Save the gas profiler outputs when we run the integration tests as an artifact in github
- [ ] Diff the gas profiler output in current PR with the latest, saved one from the main branch (it doesnt change, so no reason to keep computing it) and show it in the PR as a comment

Relevant gas benchmark POC: https://github.com/DA0-DA0/dao-contracts/pull/394/files

PR #410, opened 2022-07-20
Initial integration tests

de-husk:
Adds some DAO DAO integration tests that are run with `cargo test` using [cosm-orc](https://github.com/de-husk/cosm-orc).

First step in https://github.com/DA0-DA0/dao-contracts/issues/403. Ill do the gas diff in another PR, but this is uploading the gas report file as a github artifact.

How to add new integration tests:

```rust
#[test_context(Chain)]
#[test]
fn new_dao_has_no_items(chain: &mut Chain) {
    let res = create_dao(
        chain, 
        Some(chain.user.addr.clone()), 
        "ex_create_dao", 
       chain.user.addr.clone()
    );
    let dao = res.unwrap();

    // use the native rust types to interact with the contract
     let res = chain
        .orc
        .query(
            "cw_core",
            "exc_items_get",
            &cw_core::msg::QueryMsg::GetItem {
                key: "meme".to_string(),
            },
        )
        .unwrap();
    let res: GetItemResponse = res.data().unwrap();

    assert_eq!(res.item, None);
}
```


ci/integration_tests/src/test_harness/test_runner.rs
```diff
@@ -0,0 +1,79 @@
...
+                failed += 1;
+            } else {
+                passed += 1;
+            }
+        } else {
+            todo!()
+        }
+
+        // create a clean slate for next test
+        Chain::clear_deploys();
```
de-husk, line 43: Cosm-orc will save the contract address when a contract is instantiated (like it does with the code ids when contracts are stored). Each test reuses the same contract IDs but the contract addresses are cleared after each test, to create a clean starting state for each test.

ci/integration_tests/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Adding New Integration Tests
+
+```rust
+#[test]
+fn new_dao_has_no_items() {
+     let dao = create_dao(
+        Some(admin_addr.clone()),
+        admin_addr,
+        "cw20_staked_balance_voting",
+        "cw_proposal_single",
```
ezekiiel, line 30: Do these correspond to names of `.wasm` files? Might be nice to include that info here. :)

ezekiiel (commented):
Tiny comments. This looks super cool!

Would be nice to schedule some time to go over this PR this week. Seems like there is a lot of context here with your `cosm-orc` package that would be nice to talk through / would make it easier to review. :)

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,151 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    name: Build Tests
```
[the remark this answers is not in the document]
de-husk: So, they wont run because I excluded the `ci/*` folder from the root Cargo.toml, because I didn't want to add the annotations to ignore the wasm target architecture (otherwise `cargo release` will break when it tries to compile the integration tests). Also it kinda makes sense that the integration tests are separate from the rest of the workspace. 

But needing to do this is annoying I agree. I will just add it back into the workspace and add the wasm arch release ignore annotations. We do this for the unit tests package, so I can just make the integration tests a package and keep everything uniform.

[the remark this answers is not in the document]
de-husk: Now I also have to add a way to only run the integration / unit tests too, but that shouldnt be too bad.

ci/integration_tests/README.md
de-husk: Agreed, I need to at least add a comment like I did below with:
```
// NOTE: `cw_core` wasm was stored in test_harness::test_runner.rs:setup()
```

but I need to make this more straightforward to use in general now that the cosm-orc v1 usage has solidified here.

de-husk: Im actually just removing these params, because as `create_dao()` is currently written it really needs to just be these strings so Im going to just use them in the test. Later on I need to make `create_dao()` take in optional overrides for each of the fields once we need tests with different modules and configurations.

.github/workflows/integration_tests.yml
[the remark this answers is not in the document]
de-husk: Actually, Im going to leave `ci` folder outside of the workspace as long as you dont have any complaints about that, because I forgot I would also have to make the entire root workspace use nightly as well. I dont want to do that just for the integration tests.

ci/integration_tests/src/helpers/helper.rs
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    chain
+        .orc
+        .instantiate("cw_core", op_name, &msg, &chain.key)?;
+
+    let res = chain
+        .orc
+        .query("cw_core", op_name, &cw_core::msg::QueryMsg::DumpState {})?;
+    let state: DumpStateResponse = serde_json::from_slice(res.data.as_ref().unwrap().value())?;
```
Callum-A, line 91: Is there a way we could prettify decoding the data or is this the best way we can decode resps?

ci/integration_tests/src/tests/cw20_stake_test.rs
```diff
@@ -0,0 +1,130 @@
...
+    chain
+        .orc
+        .contract_map
+        .add_address("cw20_stake", staking_addr.to_string())
+        .unwrap();
+    let res = chain
+        .orc
+        .query(
+            "cw20_stake",
+            "exc_stake_q_cfg",
```
Callum-A, line 57: this string name doesnt match up to the query?

ci/integration_tests/src/tests/cw_core_test.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    assert_eq!(res.item, None);
+}
+
+// #### InstantiateMsg #####
+
+#[test_context(Chain)]
+#[test]
+fn instantiate_with_no_admin(chain: &mut Chain) {
+    let key: SigningKey = chain.key.clone().try_into().unwrap();
```
Callum-A, line 222: seeing this at the top of each test makes me think there should be an easier way to get this info?

ci/integration_tests/src/helpers/helper.rs
de-husk: Fixed here: https://github.com/de-husk/cosm-orc/pull/26/files#diff-3727f84401756c9f56560f3e419868ccee1b96893ffe85546f061ad82a094201R398-R407

ci/integration_tests/src/tests/cw_core_test.rs
de-husk: Great point. Ill fix this.

ci/integration_tests/src/tests/cw20_stake_test.rs
de-husk: Good catch!

ci/integration_tests/src/tests/cw20_stake_test.rs
```diff
@@ -0,0 +1,120 @@
...
+
+// #### ExecuteMsg #####
+
+#[test_context(Chain)]
+#[test]
+fn execute_stake_tokens(chain: &mut Chain) {
+    let voting_contract = "cw20_staked_balance_voting";
+
+    let res = create_dao(chain, None, "exc_stake_create_dao", chain.user.addr.clone());
+    assert!(res.is_ok());
```
ezekiiel, line 18: Suggest that we remove assertions like this. With them we don't get to see the error that occured in tests. Better to just unwrap so the panic has that information inside it.

ci/integration_tests/configs/local.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+# Add code ids if you dont need to store them:
+code_ids:
+  "cw20_base": 1
+  "cw4_group": 2
+  "cw_core": 8
+  "cw_proposal_single": 13
+  "cw4_voting": 5
+  "cw20_staked_balance_voting": 4
+  "cw20_stake": 3
```
ezekiiel, line 17: When running this locally I get these:
```suggestion
code_ids:
  "cw20_base": 1
  "cw4_group": 2
  "cw_core": 9
  "cw_proposal_single": 14
  "cw4_voting": 6
  "cw20_staked_balance_voting": 5
  "cw20_stake": 4

Comment (ezekiiel) [reply ci/integration_tests/configs/local.yaml:17]: Could we also perhaps add a section in the README about how to keep this up to date? I foresee some CI failures when people modify the deploy local script.

**Review by ezekiiel — COMMENTED** [review]
So just a review as a "user". This is great. A couple comments about improving the README, but once this works, it works. :)

Super good work!

File: ci/integration_tests/configs/local.yaml (hunk 8 above)
Comment (de-husk) [reply ci/integration_tests/configs/local.yaml:17]: So, this wont break the CI, because the CI uses [`configs/ci.yaml`](https://github.com/DA0-DA0/dao-contracts/blob/39903678b8fdf7a74f8f5754512c4eb37aec269b/ci/integration_tests/configs/ci.yaml) which doesn't have any contract ids stored. It will always just store all of the contracts from a fresh slate before all of the tests run.

I will make the local usage options clearer in the README though.

File: README.md [hunk 9]
```diff
@@ -77,6 +77,12 @@ simply run the following from the repo root:
 ./scripts/schema.sh
 ```
 
+## Testing
+
+Run `cargo t` or `cargo unit-test` from the project root to run the unit tests.
```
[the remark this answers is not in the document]
de-husk: Also, since the `SKIP_CONTRACT_STORE=true` is flakey because of the need to have the correct contract_ids stored, Im going to just recommend people not use that locally. If they just want to simply run all of the integration tests locally  then we should show them the most reliable way.

ci/integration_tests/src/tests/cw20_stake_test.rs
de-husk: Great point. I also updated the Display impl for this CosmosSDK error to actually include the relevant base error.

README.md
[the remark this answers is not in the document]
de-husk: Also, to be clear the line you commented on was for the unit tests, I assume you were talking about the line below about pointing to the integration test readme.

ci/integration_tests/src/tests/cw20_stake_test.rs
```diff
@@ -0,0 +1,118 @@
...
+fn execute_stake_tokens(chain: &mut Chain) {
+    let voting_contract = "cw20_staked_balance_voting";
+
+    let res = create_dao(chain, None, "exc_stake_create_dao", chain.user.addr.clone());
+    let dao = res.unwrap();
+
+    let voting_addr = dao.state.voting_module.as_str();
+
+    // stake dao tokens:
+    chain
```
blue-note, line 23: maybe we can do this in the create_dao method so other tests can use it later?

ci/integration_tests/src/tests/cw20_stake_test.rs
```diff
@@ -0,0 +1,118 @@
...
+use test_context::test_context;
+
+// #### ExecuteMsg #####
+
+#[test_context(Chain)]
+#[test]
+fn execute_stake_tokens(chain: &mut Chain) {
+    let voting_contract = "cw20_staked_balance_voting";
+
+    let res = create_dao(chain, None, "exc_stake_create_dao", chain.user.addr.clone());
```
blue-note, line 17: nit: maybe we can make the op_names like `"exc_stake_create_dao"` into constants for easier re-use?

blue-note (approved):
Left a couple nit comments but it LGTM, I'm down to merge!

ci/integration_tests/src/tests/cw20_stake_test.rs
de-husk: So, I wasnt planning on reusing them. These `op_name` strings are only for annotating the gas report instrumentation output with the name of the message that is associated with that gas cost.

ci/integration_tests/src/tests/cw20_stake_test.rs
de-husk: Yeah, great point. I want to make `create_dao()` more configurable first, and Im going to do that in another PR with more integration tests.

ezekiiel (approved):
This is amazing. Running tests with a real chain is amazing. Having good ways to measure gas is amazing. Great work @de-husk !

PR #412, opened 2022-07-20
Query by proposal status and voting status. Easy implementation

### LINKED ISSUE #237: Query by proposal status and voting status.
Examples:

1. What proposals have I voted yes on?
2. What proposals have I voted on?
3. What proposals are open that I have not voted on?
4. What proposals did I vote abstain on?
5. What proposals have been passed but not executed?

```rust
pub enum QueryMsg {
  FindProposals {
    wallet: String,
    status: Option<Status>,
    walletVote: Option<Vote>,
  }
}
```

Buckram123:
It's simple implementation of #237 query. What do you think @ezekiiel? Unsure if we should show created proposals too?
How do i test gas usage? Via [cosmwasm-vm](https://crates.io/crates/cosmwasm-vm)?

I can try to implement https://github.com/DA0-DA0/dao-contracts/issues/237#issuecomment-1187977788
But I think this way we increase gas usage on votes?

TODO(If we ok with this):
- [ ] test gas
- [x] implement same for `cw-proposal-multiple` module 
- [x] test pagination

Closes #237


contracts/cw-proposal-single/src/contract.rs
```diff
@@ -609,9 +609,47 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+    wallet: String,
+    status: Option<Status>,
+    wallet_vote: Option<Vote>,
+    start_after: Option<u64>,
+    limit: Option<u64>,
+) -> StdResult<Binary> {
+    let wallet: Addr = deps.api.addr_validate(&wallet)?;
+    let min = start_after.map(Bound::exclusive);
+    let limit = limit.unwrap_or(DEFAULT_LIMIT);
+    let props: Vec<ProposalResponse> = PROPOSALS
```
Callum-A, line 634: i like this, and fairly certain as it's paginated gas wouldn't be an issue!

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -609,9 +609,47 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+        .filter_map(|(p_id, prop)| {
+            if status.map_or(true, |st| st == prop.status) {
+                if let Ok(Some(ballot)) = BALLOTS.may_load(deps.storage, (p_id, wallet.clone())) {
+                    if wallet_vote.map_or(true, |v| v == ballot.vote) {
+                        return Some(prop.into_response(&env.block, p_id));
+                    }
+                }
+            }
+            None
+        })
```
ezekiiel, line 648: I'm not entirely sure this works.

Say I have proposals (1), (2), and (3) and I have voted no proposals (1) and (2). If I run this query to find "the first one" (`limit == 1`) I have not voted on, the `limit` call here will cause us to only check the first proposal and I think I'll get an empty list back.

I think fixing this is a pretty hard problem if we want to be gas efficient. My strong suspicion is that we'll need additional state.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -609,9 +609,47 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+    let wallet: Addr = deps.api.addr_validate(&wallet)?;
+    let min = start_after.map(Bound::exclusive);
+    let limit = limit.unwrap_or(DEFAULT_LIMIT);
+    let props: Vec<ProposalResponse> = PROPOSALS
+        .range(deps.storage, min, None, Order::Ascending)
+        .take(limit as usize)
+        .collect::<StdResult<Vec<(u64, SingleChoiceProposal)>>>()?
+        .into_iter()
+        .filter_map(|(p_id, prop)| {
+            if status.map_or(true, |st| st == prop.status) {
```
ezekiiel, line 640: Great use of `filter_map` and `map_or`!

contracts/cw-proposal-single/src/contract.rs
ezekiiel: One solution that would likely be fine would just be to move the limit call to after `filter_map`. Queries very rarely run out of gas because of compute limits. I suspect we could have thousands of proposals before this ran out of gas.

Another would be to use your suggestion of the `proposals_by_addr` map. Would love to talk through this in the Discord! https://github.com/DA0-DA0/dao-contracts/issues/237#issuecomment-1187977788

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -4166,3 +4166,361 @@ fn test_return_deposit_to_dao_on_proposal_failure() {
     // Deposit should now belong to the DAO.
     assert_eq!(balance.balance, Uint128::new(1));
 }
+
+#[test]
+fn test_find_proposals() {
```
ezekiiel, line 4171: Love how comprehensive this test is. :)

contracts/cw-proposal-single/src/contract.rs
Buckram123: My idea was basically to make `FindProposals`, but sorted, maybe we could rename it to smth like `FilterListProposals` then, what you think?(not sure about new naming for `limit`) 

And I think `start_after` can become pretty user-unfriendly, let's say user put some `start_after` do we `start_after` `proposal_id` or index of user proposals. 

But maybe I'm just **too worried** about the gas and no one will have enough proposals for it to break :DD

contracts/cw-proposal-single/src/tests.rs
Buckram123: Was it a hint that I should separate pagination into a separate test? Or we ok with this? :DD

contracts/cw-proposal-single/src/contract.rs
Buckram123: Thanks

contracts/cw-proposal-single/src/contract.rs
ezekiiel: Great point about `start_after` here. I like your idea of making the query `FilterListProposals`, maybe in order to make `start_after` play nice, we could return a response like this:

```
struct FilterListProposalsResponse {
  proposals: Vec<Proposal>,
  last_checked: String
}
```

Then, a frontend can use that `last_checked` value to set `start_after` for the next iteration. I think that if we do this we can have pagination (no gas issues!) and get the queries we want.

contracts/cw-proposal-single/src/tests.rs
ezekiiel: Aha - is a genuine compliment.

Buckram123: Still added extra blocks and comments for easier read (:

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -597,9 +600,53 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+    start_after: Option<u64>,
+    limit: Option<u64>,
+) -> StdResult<Binary> {
+    let wallet: Addr = deps.api.addr_validate(&wallet)?;
+    let min = start_after.map(Bound::exclusive);
+    let limit = limit.unwrap_or(DEFAULT_LIMIT);
+    let props: Vec<ProposalResponse> = PROPOSALS
+        .range(deps.storage, min, None, cosmwasm_std::Order::Ascending)
+        .collect::<StdResult<Vec<(u64, MultipleChoiceProposal)>>>()?
+        .into_iter()
```
ezekiiel, line 630: Do we need to `collect` here? Doing this will cause the iterator to move over the entire collection every time we call this. Could we filter map directly on the iterator somehow?

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -597,9 +600,53 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                    if wallet_vote.map_or(true, |v| v == ballot.vote) {
+                        return prop.into_response(&env.block, p_id).ok();
+                    }
+                }
+            }
+            None
+        })
+        .take(limit as usize)
+        .collect();
+    let last_proposal_id = props.last().map_or(0, |res| res.id);
```
ezekiiel, line 643: This should probably be set as the last proposal that we have visited, not the last one we're returning information about. Could we set this based on the IDs we see in `filter_map` above?

For example, currently if we look at proposals 1 through 10 to get proposals 1, 2, and 5 this will return 5. I _think_ we want to return 10 as that is the last one we checked.

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -597,9 +600,53 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+    let wallet: Addr = deps.api.addr_validate(&wallet)?;
+    let min = start_after.map(Bound::exclusive);
+    let limit = limit.unwrap_or(DEFAULT_LIMIT);
+    let props: Vec<ProposalResponse> = PROPOSALS
+        .range(deps.storage, min, None, cosmwasm_std::Order::Ascending)
+        .collect::<StdResult<Vec<(u64, MultipleChoiceProposal)>>>()?
+        .into_iter()
+        .filter_map(|(p_id, prop)| {
+            if status.map_or(true, |st| st == prop.status) {
+                if let Ok(Some(ballot)) = BALLOTS.may_load(deps.storage, (p_id, wallet.clone())) {
```
ezekiiel, line 633: I'm a little worried about silencing the `Err` case here. `may_load` returning an error as opposed to `None` has pretty different semantics. Wdyt?

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -597,9 +600,53 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                    if wallet_vote.map_or(true, |v| v == ballot.vote) {
```
ezekiiel, line 634: Great use of `map_or`. :)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -609,9 +610,53 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+            wallet_vote,
+            start_after,
+            limit,
+        } => {
+            query_filter_list_proposals(deps, env, wallet, status, wallet_vote, start_after, limit)
+        }
     }
 }
 
+fn query_filter_list_proposals(
```
ezekiiel, line 625: Same feedback on this method as on the one for `cw-proposal-multiple`. :)

ezekiiel:
@Buckram123 how are things going? If you're tight on time or otherwise occupied (which would be extremely valid given your situation) I'm happy to pick up the last few things here. :)

contracts/cw-proposal-multiple/src/contract.rs
Buckram123: Changed it, does it looks better now?

contracts/cw-proposal-multiple/src/contract.rs
Buckram123: Thanks, it's very clever!

contracts/cw-proposal-multiple/src/contract.rs
Buckram123: You are right, thanks!

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -646,9 +649,61 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                    match BALLOTS.may_load(deps.storage, (p_id, wallet.clone())) {
+                        Ok(Some(ballot)) => {
+                            if wallet_vote.map_or(true, |v| v == ballot.vote) {
+                                return Some(prop.into_response(&env.block, p_id));
+                            }
+                        }
+                        Err(err) => return Some(Err(err)),
+                        _ => (),
+                    }
+                }
```
ezekiiel, line 693: If no wallet vote filter is applied, how would you feel about returning the proposal here if they have not voted?

This would let us find "open but not voted on" proposals by passing in `None` for wallet vote and `Open` for status. Something like this should do that (haven't actually tested this code):

```suggestion
                if status.map_or(true, |st| st == prop.status) {
                    if let Some(wallet_vote_filter) = wallet_vote {
                        if let Some(ballot) = BALLOTS.may_load(deps.storage, (p_id, wallet.clone()))? {
                            if ballot == wallet_vote_filter {
                                // Proposal has correct status and wallet vote.
                                return Some(prop.into_response(&env.block, p_id));
                            }
                        }
                    } else {
                         // No wallet vote filter specified so the proposal having the correct status is enough.
                         return Some(prop.into_response(&env.block, p_id));
                    }
                }
```

contracts/cw-proposal-multiple/src/msg.rs
```diff
@@ -158,6 +158,13 @@ pub enum QueryMsg {
     },
     ProposalHooks {},
     VoteHooks {},
+    FilterListProposals {
```
ezekiiel, line 161: Would be great to add a little docstring to this method describing what each field does. :)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -622,9 +623,60 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+    start_after: Option<u64>,
+    limit: Option<u64>,
+) -> StdResult<Binary> {
+    let wallet: Addr = deps.api.addr_validate(&wallet)?;
+    let min = start_after.map(Bound::exclusive);
+    let limit = limit.unwrap_or(DEFAULT_LIMIT);
+    let mut last_proposal_id = 0;
+    let props: Vec<ProposalResponse> = PROPOSALS
+        .range(deps.storage, min, None, Order::Ascending)
+        .filter_map(|res| match res {
```
ezekiiel, line 653: Same feedback here about the filtering as above. :)

ezekiiel (commented):
Looks great! Just one last logic change I think we may need in the filtering and then this is good to go.

contracts/cw-proposal-multiple/src/contract.rs
Buckram123: Wow, I like that idea! But what if fhey want to find all the proposals they voted on,  with any vote?Imo good to have both options, maybe in different query. @ezekiiel Wdyt?

ezekiiel: Oh, great point. :) Maybe we could make the `wallet_vote` field an enum?

```rust
pub enum WalletVote {
  Voted {
    position: Option<SingleChoiceVote>,
  },
  DidNotVote {},
}
```

Buckram123: Also for `DidNotVote` we want `voting_power` to be > 0, right?

ezekiiel (approved):
incredible, incredible work. love how this has turned out. thank you. :)

ezekiiel:
@Buckram123 after much iteration in this area, i think we've decided to go in the direction of using indexers for these kinds of queries. that gives us wins in contract simplicity and in performance. this query is superseded by the indexer that we built [here](https://github.com/DA0-DA0/subquery-indexers/tree/main/indexers/proposals).

i really appreciate your work getting this ready. not landing this is not in any way a reflection of the code here or how talented you are. :) this work got us started iterating on this question, and the end result is that indexers are probably the way. thank you again for the really nice PR!

Buckram123:
@ezekiiel thank you for giving me such a great opportunity to get the best reviews in my life. (:
I like the direction you decided to pick, would love to play around with the indexer you guys are creating!

PR #413, opened 2022-07-21
add proposal module prefixes in cw-core

### LINKED ISSUE #377: Disable proposal modules instead of removing them.
Currently to disable a proposal module it is removed from the core contract. This causes trouble in the frontend because it makes assigning globally unique names to proposals hard. Instead of removing modules from the module list we should instead mark them as "disabled". They will still appear in queries, but will not be able to execute messages on the core module.

For example, say we have two proposal modules. They each keep their own proposal counter. We might label proposals by mapping letters to an index and then prepending the mapped index to the proposal id. This would make proposal 1 in the first module proposal "a1" and proposal 1 in the second module proposal "b1".

This works fine, but as soon as a module is removed from the list we're in trouble as our names now change.

Implementing this will require two changes:

1. Make the order modules are returned in queries is a function of when they were added and not a function of their address (right now it is alphabetical).
2. Disable modules instead of removing them from the module list. This will likely involve changing the type of the proposal module list item to something like `{ address, enabled }`.

blue-note:
closes #377


contracts/cw-core/src/contract.rs
```diff
@@ -687,9 +687,33 @@ pub fn query_cw20_balances(
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
-    // Don't do any state migrations.
-    Ok(Response::default())
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
+    match msg {
+        MigrateMsg::FromBeta {} => {
+            let keys_raw: Vec<(usize, Addr)> = PROPOSAL_MODULES
```
janitachalam, line 693: this still does not deserialize properly into Addr

contracts/cw-core/src/contract.rs
```diff
@@ -728,3 +764,57 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
         _ => Err(ContractError::UnknownReplyID {}),
     }
 }
+
```
janitachalam, line 767: O(logbase26(num_proposal_modules))

contracts/cw-core/src/contract.rs
janitachalam: aha it works now! needed to deserialize to string first then reconstruct address

contracts/cw-core/src/contract.rs
```diff
@@ -180,7 +180,13 @@ pub fn execute_proposal_hook(
...
 ) -> Result<Response, ContractError> {
     // Check that the message has come from one of the proposal modules
-    if !PROPOSAL_MODULES.has(deps.storage, sender) {
+    if !PROPOSAL_MODULES.has(deps.storage, sender.clone()) {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // Check that the message has come from an active module
+    let module = PROPOSAL_MODULES.load(deps.storage, sender)?;
+    if module.status == ProposalModuleStatus::Disabled {
```
janitachalam, line 189: what do we think about this usage of disabled?

contracts/cw-core/src/contract.rs
```diff
@@ -728,3 +786,60 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
...
+        let remainder = dividend % 26;
+        dividend /= 26;
+        let remainder_str = std::str::from_utf8(&[(remainder + 65) as u8])?.to_owned();
+        prefix.push_str(&remainder_str);
+        if dividend == 0 {
+            break;
+        }
+    }
+    Ok(prefix.chars().rev().collect())
+}
```
NoahSaso, line 805: I think we want `26` to yield `AA`, `27` to yield `AB`, and `28` to yield `AC`, increasing in that pattern. This algorithm doesn't seem to ever output `AA`, as 25 is `Z` and 26 is `BA`.

Here is the TypeScript algorithm we found online, converted to Rust:

```suggestion
pub(crate) fn derive_proposal_module_prefix(mut dividend: usize) -> StdResult<String> {
    dividend += 1;
    // Pre-allocate string
    let mut prefix = String::with_capacity(10);
    loop {
        let remainder = (dividend - 1) % 26;
        dividend = (dividend - remainder) / 26;
        let remainder_str = std::str::from_utf8(&[(remainder + 65) as u8])?.to_owned();
        prefix.insert_str(0, &remainder_str);
        if dividend == 0 {
            break;
        }
    }
    Ok(prefix.chars().rev().collect())
}
```

Would that work? Or did you modify the algorithm for any reason?

janitachalam: yeah this works better! mine was slightly different because I was converting to base26 where all of the "numbers" are A-Z, and since A maps to 0, a prefix wouldn't start with A unless it is 0. but agreed its better to have 27 be AA

contracts/cw-core/src/contract.rs
```diff
@@ -180,7 +180,13 @@ pub fn execute_proposal_hook(
...
     // Check that the message has come from one of the proposal modules
-    if !PROPOSAL_MODULES.has(deps.storage, sender) {
+    if !PROPOSAL_MODULES.has(deps.storage, sender.clone()) {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // Check that the message has come from an active module
+    let module = PROPOSAL_MODULES.load(deps.storage, sender)?;
+    if module.status == ProposalModuleStatus::Disabled {
         return Err(ContractError::Unauthorized {});
```
Callum-A, line 190: nit: could be a better error here

contracts/cw-core/src/tests.rs
```diff
@@ -144,7 +149,7 @@ fn test_instantiate_with_n_gov_modules(n: usize) {
 }
 
 #[test]
-#[should_panic(expected = "Execution would result in no governance modules being present.")]
+#[should_panic(expected = "Execution would result in no governance modules being active.")]
```
ezekiiel, line 152: How would you feel about changing this error to say "proposal modules" instead of "governance modules" while we're at it?

contracts/cw-core/src/tests.rs
```diff
@@ -2117,7 +2106,33 @@ fn test_pause() {
 }
 
 #[test]
-fn test_migrate() {
+fn test_dump_state_proposal_modules() {
```
ezekiiel, line 2109: Another great one to test. Thanks!

contracts/cw-core/src/tests.rs
```diff
@@ -2199,10 +2214,31 @@ fn test_migrate() {
     assert_eq!(new_state, state);
 }
 
+#[test]
+fn test_migrate_from_beta() {
```
ezekiiel, line 2218: Might be nice to write a version of this test using the integration test approach used by `test_migrate_from_compatible`. IMO, those tests tend to cover a little more / make us a little more resistant to weird regressions later (fewer builtin assumptions about state).

contracts/cw-core/src/tests.rs
```diff
@@ -2230,3 +2266,107 @@ fn test_execute_stargate_msg() {
...
+
+    let module_2 = &modules[1];
+    assert_eq!(module_2.status, ProposalModuleStatus::Active {});
+    assert_eq!(module_2.prefix, "B");
+    assert_eq!(&module_2.address, &modules[1].address);
+
+    let module_3 = &modules[2];
+    assert_eq!(module_3.status, ProposalModuleStatus::Active {});
+    assert_eq!(module_3.prefix, "C");
+    assert_eq!(&module_3.address, &modules[2].address);
```
ezekiiel, line 2353: THIS IS SO COOL

contracts/cw-core/src/tests.rs
```diff
@@ -2230,3 +2266,107 @@ fn test_execute_stargate_msg() {
...
+}
+
+fn get_active_modules(app: &App, gov_addr: Addr) -> Vec<ProposalModule> {
```
ezekiiel, line 2356: How would you feel about having a query for this? This is likely what frontends will care about most.

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -2552,7 +2552,7 @@ fn test_active_threshold_none() {
         .unwrap();
 
     assert_eq!(governance_modules.len(), 1);
-    let govmod_single = governance_modules.into_iter().next().unwrap();
+    let govmod_single = governance_modules.into_iter().next().unwrap().address;
```
ezekiiel, line 2555: So many of these changes lol. Love how many tests we have for these modules.

contracts/cw20-stake/src/tests.rs
```diff
@@ -387,6 +390,42 @@ fn test_update_config() {
     assert_eq!(err, ContractError::Unauthorized {})
 }
 
+#[test]
+fn test_migrate_from_beta() {
```
ezekiiel, line 394: THANK YOU!

packages/cw-paginate/src/lib.rs
```diff
@@ -44,6 +44,29 @@ where
     }
 }
 
+/// Same as `paginate_map` but only returns the values.
+pub fn paginate_map_values<'a, K, V>(
```
ezekiiel, line 48: Great to add this. Thanks!

ezekiiel (commented):
This is so good. Awesome work. Just a couple nits and questions. Seriously, good shit. Love how well tested and thought out this is.

contracts/cw-core/src/contract.rs
```diff
@@ -562,8 +584,9 @@ pub fn query_dump_state(deps: Deps, env: Env) -> StdResult<Binary> {
     let config = CONFIG.load(deps.storage)?;
     let voting_module = VOTING_MODULE.load(deps.storage)?;
     let proposal_modules = PROPOSAL_MODULES
-        .keys(deps.storage, None, None, cosmwasm_std::Order::Descending)
-        .collect::<Result<Vec<Addr>, _>>()?;
+        .range(deps.storage, None, None, cosmwasm_std::Order::Descending)
+        .map::<StdResult<ProposalModule>, _>(|kv| Ok(kv?.1))
+        .collect::<Result<Vec<ProposalModule>, _>>()?;
```
Buckram123, line 589:
```suggestion
        .map(|kv| Ok(kv?.1))
        .collect::<StdResult<Vec<ProposalModule>>>()?;
```
@ezekiiel ez kill of turbofish

contracts/cw-core/src/contract.rs
```diff
@@ -687,9 +710,32 @@ pub fn query_cw20_balances(
...
+            keys.into_iter()
+                .try_for_each::<_, StdResult<()>>(|(index, key)| {
+                    let path = PROPOSAL_MODULES.key(key.clone());
+                    let prefix = derive_proposal_module_prefix(index)?;
+                    let proposal_module = to_vec(&ProposalModule {
+                        address: key,
+                        status: ProposalModuleStatus::Active {},
+                        prefix,
+                    })?;
+                    deps.storage.set(&path, &proposal_module);
```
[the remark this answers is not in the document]
janitachalam: I think it's safer to use a new namespace like in the other PR, so going to go that route

contracts/cw-core/src/contract.rs
```diff
@@ -312,15 +318,43 @@ pub fn execute_update_proposal_modules(
...
     }
 
-    for addr in to_remove {
+    for addr in to_disable {
         let addr = deps.api.addr_validate(&addr)?;
-        PROPOSAL_MODULES.remove(deps.storage, addr);
+        let mut module = PROPOSAL_MODULES
+            .load(deps.storage, addr.clone())
+            .map_err(|_| ContractError::ProposalModuleDoesNotExist {
+                address: addr.clone(),
```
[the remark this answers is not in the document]
janitachalam: since it's occurring in a closure does that mean the clone will only occur if the error occur? If not I guess I can put it behind an if statement

contracts/cw-core/src/contract.rs
```diff
@@ -698,7 +744,19 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
         PROPOSAL_MODULE_REPLY_ID => {
             let res = parse_reply_instantiate_data(msg)?;
             let prop_module_addr = deps.api.addr_validate(&res.contract_address)?;
-            PROPOSAL_MODULES.save(deps.storage, prop_module_addr, &Empty {})?;
+            let num_modules = PROPOSAL_MODULES
```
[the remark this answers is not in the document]
janitachalam: No particular reason. I actually think it's a good idea to store both of those values in state to avoid the problem you mentioned up there.

contracts/cw-core/src/contract.rs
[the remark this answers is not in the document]
JakeHartnell: @ethanfrey / @orkunkl, any thoughts here? Should we be paranoid?

contracts/cw-core/src/contract.rs
```diff
@@ -312,15 +318,43 @@ pub fn execute_update_proposal_modules(
...
+            .load(deps.storage, addr.clone())
+            .map_err(|_| ContractError::ProposalModuleDoesNotExist {
+                address: addr.clone(),
+            })?;
+        // If module is already disabled, it will be a no-op.
+        module.status = ProposalModuleStatus::Disabled {};
+        PROPOSAL_MODULES.save(deps.storage, addr, &module)?;
+    }
+
+    let active_count = PROPOSAL_MODULES
```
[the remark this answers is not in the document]
janitachalam: I agree about avoiding having an O(number of modules) operation. I think it might be better to keep this API but add counts for active / total modules in state like you mentioned so we can avoid this check

contracts/cw-core/src/tests.rs
janitachalam: sure thing

Callum-A (approved):

contracts/cw-core/src/contract.rs
```diff
@@ -179,9 +180,13 @@ pub fn execute_proposal_hook(
...
 ) -> Result<Response, ContractError> {
-    // Check that the message has come from one of the proposal modules
-    if !PROPOSAL_MODULES.has(deps.storage, sender) {
-        return Err(ContractError::Unauthorized {});
+    let module = PROPOSAL_MODULES
+        .may_load(deps.storage, sender.clone())?
+        .ok_or(ContractError::Unauthorized {})?;
+
+    // Check that the message has come from an active module
+    if module.status == ProposalModuleStatus::Disabled {
```
ezekiiel, line 188: nit: could we check here for `module.status != ProposalModuleStatus::Enabled` instead? Conditions like that are helpful because they fail closed if we add another status later.

contracts/cw-core/src/contract.rs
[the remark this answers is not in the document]
ezekiiel: Just a reflection on the rust compiler being more strict than it needs to be. I think your approach here is fine. :)

contracts/cw-core/src/contract.rs
```diff
@@ -312,15 +317,38 @@ pub fn execute_update_proposal_modules(
...
+    for addr in to_disable {
         let addr = deps.api.addr_validate(&addr)?;
-        PROPOSAL_MODULES.remove(deps.storage, addr);
+        let mut module = PROPOSAL_MODULES
+            .load(deps.storage, addr.clone())
+            .map_err(|_| ContractError::ProposalModuleDoesNotExist {
+                address: addr.clone(),
+            })?;
+
+        if module.status == ProposalModuleStatus::Disabled {
```
ezekiiel, line 334: nit: `!= Enabled`

contracts/cw-core/src/contract.rs
```diff
@@ -312,15 +317,38 @@ pub fn execute_update_proposal_modules(
...
+
+        // If disabling this module will cause there to be no active modules, return error.
+        // We don't check the active count before disabling because there may erroneously be
+        // modules in to_disable which are already disabled.
+        ACTIVE_PROPOSAL_MODULE_COUNT.update(deps.storage, |count| {
+            if count <= 1 && to_add.is_empty() {
+                return Err(ContractError::NoActiveProposalModules {});
+            }
+            Ok(count - 1)
+        })?;
```
ezekiiel, line 348: How would you feel about doing this at the end of this for loop based on the length of `to_disable`? Stops us from doing a storage operation every time.

contracts/cw-core/src/contract.rs
```diff
@@ -541,24 +560,50 @@ pub fn query_proposal_modules(
...
+    start_after: Option<String>,
+    limit: Option<u32>,
+) -> StdResult<Binary> {
+    let values = paginate_map_values(
+        deps,
+        &PROPOSAL_MODULES,
+        start_after
+            .map(|s| deps.api.addr_validate(&s))
+            .transpose()?,
+        None,
```
ezekiiel, line 592: Could we add a comment to this query somewhere that this is not gas effective and needs to visit all modules? Fine for it to be this way IMO as the information is exposed elsewhere in a happy way.

contracts/cw-core/src/state.rs
```diff
@@ -49,8 +67,15 @@ pub const PAUSED: Item<Expiration> = Item::new("paused");
 
 /// The voting module associated with this contract.
 pub const VOTING_MODULE: Item<Addr> = Item::new("voting_module");
-/// The proposal modules assocaited with this contract.
-pub const PROPOSAL_MODULES: Map<Addr, Empty> = Map::new("proposal_modules");
+
+/// The proposal modules associated with this contract.
+pub const PROPOSAL_MODULES: Map<Addr, ProposalModule> = Map::new("proposal_modules_v1");
```
ezekiiel, line 72: Could we add a comment here about why we need to use the `_v1` storage key?

ezekiiel: Also, how about naming this `v2` as we'll be releasing it as DAO DAO v2.

contracts/cw-core/src/contract.rs
```diff
@@ -706,9 +756,31 @@ pub fn query_cw20_balances(
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
-    // Don't do any state migrations.
-    Ok(Response::default())
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
+    match msg {
+        MigrateMsg::FromBeta {} => {
```
ezekiiel, line 761: How would you feel about calling this variant `FromV1` as that's technically what we're migrating from?

contracts/cw-core/src/contract.rs
```diff
@@ -717,10 +789,31 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
         PROPOSAL_MODULE_REPLY_ID => {
             let res = parse_reply_instantiate_data(msg)?;
             let prop_module_addr = deps.api.addr_validate(&res.contract_address)?;
-            PROPOSAL_MODULES.save(deps.storage, prop_module_addr, &Empty {})?;
+            let total_module_count = TOTAL_PROPOSAL_MODULE_COUNT
+                .may_load(deps.storage)?
```
ezekiiel, line 793: Instead of `may_load` here, how would you feel about setting this to zero during instantiation?

Otherwise, there is a period of time where this has no value when modules are being instantiated. If they were to query back to the DAO for this info it would error as there is no value.

contracts/cw-core/src/contract.rs
```diff
@@ -717,10 +789,31 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
...
+                address: prop_module_addr.clone(),
+                status: ProposalModuleStatus::Active,
+                prefix,
+            };
+
+            PROPOSAL_MODULES.save(deps.storage, prop_module_addr, &prop_module)?;
+
+            // Save active and total proposal module counts.
+            let active_count = ACTIVE_PROPOSAL_MODULE_COUNT
+                .may_load(deps.storage)?
```
ezekiiel, line 807: Same as above, consider setting to zero during instantiation.

contracts/cw-core/src/state.rs
ezekiiel, line 72:
```suggestion
pub const PROPOSAL_MODULES: Map<Addr, ProposalModule> = Map::new("proposal_modules_v2");
```

ezekiiel: Might be nice to add a comment here about why we don't use the key `proposal_modules` for future viewers.

ezekiiel (commented):
Looking good. Couple of nits and then let's ship this!

contracts/cw-core/src/contract.rs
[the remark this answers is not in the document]
blue-note: oh I get what you were saying

contracts/cw-core/src/contract.rs
blue-note: so I agree with the nit above, but by that same logic I think we may actually want to check for `== Disabled` here since we are specifically trying to disable modules

contracts/cw-core/src/contract.rs
blue-note: much better idea haha

contracts/cw-core/src/contract.rs
blue-note: good point!

contracts/cw-core/src/contract.rs
ezekiiel: great point!

contracts/cw-core/src/contract.rs
```diff
@@ -312,35 +321,47 @@ pub fn execute_update_proposal_modules(
...
+
+        module.status = ProposalModuleStatus::Disabled {};
+        PROPOSAL_MODULES.save(deps.storage, addr, &module)?;
     }
 
+    // If disabling this module will cause there to be no active modules, return error.
+    // We don't check the active count before disabling because there may erroneously be
+    // modules in to_disable which are already disabled.
+    ACTIVE_PROPOSAL_MODULE_COUNT.update(deps.storage, |count| {
+        if count <= disable_count && to_add.is_empty() {
```
ezekiiel, line 353:
```suggestion
        if count < disable_count && to_add.is_empty() {
```

In the `count == disable_count` case I _think_ we're actually OK here as we check that we will be adding at least one module with the `to_add.is_empty()` part of the condition. Otherwise I think we prevent disabling all modules and adding a new one here which is probably a pretty frequent operation. Might be nice to add a test for this.

contracts/cw-core/src/contract.rs
```diff
@@ -541,24 +565,52 @@ pub fn query_proposal_modules(
...
-    to_binary(&paginate_map_keys(
+    to_binary(&paginate_map_values(
         deps,
         &PROPOSAL_MODULES,
         start_after
             .map(|s| deps.api.addr_validate(&s))
             .transpose()?,
         limit,
-        cosmwasm_std::Order::Descending,
+        cosmwasm_std::Order::Ascending,
```
ezekiiel, line 582: Nice. :) This is good UX.

contracts/cw-core/src/contract.rs
ezekiiel: Bumping this. :)

contracts/cw-core/src/query.rs
```diff
@@ -20,9 +20,13 @@ pub struct DumpStateResponse {
...
     /// The governance modules associated with the governance
     /// contract.
-    pub proposal_modules: Vec<Addr>,
+    pub proposal_modules: Vec<ProposalModule>,
     /// The voting module associated with the governance contract.
     pub voting_module: Addr,
+    /// The number of active proposal modules.
+    pub active_proposal_module_count: u64,
+    /// The total number of proposal modules.
+    pub total_proposal_module_count: u64,
```
ezekiiel, line 29: Nice. :)

contracts/cw-core/src/state.rs
```diff
@@ -49,8 +67,15 @@ pub const PAUSED: Item<Expiration> = Item::new("paused");
...
+
+/// The count of active proposal modules associated with this contract.
+pub const ACTIVE_PROPOSAL_MODULE_COUNT: Item<u64> = Item::new("active_proposal_module_count");
+
+/// The count of total proposal modules associated with this contract.
+pub const TOTAL_PROPOSAL_MODULE_COUNT: Item<u64> = Item::new("total_proposal_module_count");
```
ezekiiel, line 78: We can use `u32` here safely. The WASM VM is 32 bit so using a u64 here is going to generate some weird machine code / doesn't actually improve the safety.

packages/cw-paginate/src/lib.rs
```diff
@@ -85,6 +85,29 @@ where
...
+    map: &Map<'a, K, V>,
+    start_after: Option<K>,
+    limit: Option<u32>,
+    order: Order,
+) -> StdResult<Vec<V>>
+where
+    K: Bounder<'a> + KeyDeserialize<Output = K> + 'static,
+    V: serde::de::DeserializeOwned + serde::Serialize,
+{
+    let items = map
```
ezekiiel, line 100: Need to patch with @onewhiskeypls's fix here: https://github.com/DA0-DA0/dao-contracts/pull/447#discussion_r944961140

ezekiiel (commented):
Two small things around condition for failing the update modules call and integrating the pagination work @onewhiskeypls did. Thanks for your patience with all these nits.

contracts/cw-core/src/contract.rs
blue-note: sorry! missed this

contracts/cw-core/src/contract.rs
blue-note: if we just do `count < disable_count && to_add.is_empty()`, we miss the case where `count == disable_count && to_add.is_empty()`, meaning we could disable all modules and not add any. The case where we disable all modules and add a new one will not pass this condition (because to_add.is_empty() will be false). I believe [this test should cover that scenario](https://github.com/DA0-DA0/dao-contracts/blob/19f06a42b55f9e118f8533a19c1d00df4d69e82c/contracts/cw-core/src/tests.rs#L431) (the (1,1) case)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -722,9 +724,71 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
-    // Don't do any state migrations.
-    Ok(Response::default())
+pub fn migrate(deps: DepsMut, env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
```
blue-note, line 727: This was merged in from another stamped PR

contracts/cw-core/examples/schema.rs
```diff
@@ -45,12 +45,16 @@ fn main() {
...
     export_schema_with_title(&schema_for!(Vec<Addr>), &out_dir, "Cw20TokenListResponse");
     export_schema_with_title(&schema_for!(Vec<Addr>), &out_dir, "Cw721TokenListResponse");
+    export_schema_with_title(&schema_for!(Vec<String>), &out_dir, "ListItemsResponse");
+    export_schema_with_title(&schema_for!(Addr), &out_dir, "VotingModuleResponse");
     export_schema_with_title(
-        &schema_for!(Vec<Addr>),
+        &schema_for!(Vec<ProposalModule>),
         &out_dir,
-        "GovernanceModulesResponse",
+        "ProposalModulesResponse",
```
ezekiiel, line 53: Nice catch here. :)

contracts/cw-core/src/contract.rs
```diff
@@ -179,9 +184,13 @@ pub fn execute_proposal_hook(
...
-    // Check that the message has come from one of the proposal modules
-    if !PROPOSAL_MODULES.has(deps.storage, sender) {
-        return Err(ContractError::Unauthorized {});
+    let module = PROPOSAL_MODULES
+        .may_load(deps.storage, sender.clone())?
+        .ok_or(ContractError::Unauthorized {})?;
+
+    // Check that the message has come from an active module
+    if module.status != ProposalModuleStatus::Enabled {
+        return Err(ContractError::ModuleAlreadyDisabled { address: sender });
```
ezekiiel, line 193: Tiny nit: maybe this should have its own error message? Seems `ModuleDisabled` or something?

contracts/cw-core/src/contract.rs
ezekiiel: Oh, you're so right. Thank for the detailed response!

contracts/cw-core/src/contract.rs
```diff
@@ -312,35 +321,47 @@ pub fn execute_update_proposal_modules(
...
     }
 
+    // If disabling this module will cause there to be no active modules, return error.
+    // We don't check the active count before disabling because there may erroneously be
+    // modules in to_disable which are already disabled.
+    ACTIVE_PROPOSAL_MODULE_COUNT.update(deps.storage, |count| {
+        if count <= disable_count && to_add.is_empty() {
+            return Err(ContractError::NoActiveProposalModules {});
+        }
+        Ok(count - disable_count)
```
ezekiiel, line 356: Really love that you set this to `count - disable_count` and not `count - disable_count + to_add.len()`. Tiny detail, but means the state is correct during the dispatch of proposal module instantiation messages. Good stuff.

contracts/cw-core/src/state.rs
```diff
@@ -49,8 +67,17 @@ pub const PAUSED: Item<Expiration> = Item::new("paused");
 
 /// The voting module associated with this contract.
 pub const VOTING_MODULE: Item<Addr> = Item::new("voting_module");
-/// The proposal modules assocaited with this contract.
-pub const PROPOSAL_MODULES: Map<Addr, Empty> = Map::new("proposal_modules");
+
+/// The proposal modules associated with this contract.
+/// When we change the data format of this map, we update the key (previously "proposal_modules")
+/// to create a new namespace for the changed state.
```
ezekiiel, line 73: Great comment. :)

ezekiiel (approved):
<3

Amazing work.

PR #417, opened 2022-07-25
add proposal timestamp migration 

janitachalam:


contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+                        min_voting_period: prop.min_voting_period,
+                        expiration: prop.expiration,
+                        threshold: prop.threshold,
+                        total_power: prop.total_power,
+                        msgs: prop.msgs,
+                        status: prop.status,
+                        votes: prop.votes,
+                        allow_revoting: prop.allow_revoting,
+                        deposit_info: prop.deposit_info,
+                        created: env.block.time,
```
janitachalam, line 783: does anyone know of a better way to get the created timestamp based on the start_time block height?

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+                .keys(deps.storage, None, None, Order::Ascending)
+                .count();
+            println!("keys count {}", count);
+
+            let current = current_map
+                .range(deps.storage, None, None, Order::Ascending)
+                .collect::<StdResult<Vec<(u64, BetaProposal)>>>()?;
+
+            // Add migrated entries to new map.
+            // TODO: determine whether this needs to be paginated
```
janitachalam, line 765: pending gas analysis

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+        /// The messages that will be executed should this proposal pass.
+        pub msgs: Vec<CosmosMsg<Empty>>,
+
+        pub status: Status,
+        pub votes: Votes,
+        pub allow_revoting: bool,
+
+        /// Information about the deposit that was sent as part of this
+        /// proposal. None if no deposit.
+        pub deposit_info: Option<CheckedDepositInfo>,
```
ezekiiel, line 748: How would you feel about removing the documentation comments here? Might be nice to just add one at the top describing what these are / linking to the commit where the proposal state was this.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+    }
+
+    match msg {
+        MigrateMsg::FromBeta {} => {
+            // Retrieve current map from storage
+            let current_map: Map<u64, BetaProposal> = Map::new("proposals");
+            let count = current_map
+                .keys(deps.storage, None, None, Order::Ascending)
+                .count();
+            println!("keys count {}", count);
```
ezekiiel, line 758: Probably want to get rid of this if we're done debugging. :)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+
+            let current = current_map
+                .range(deps.storage, None, None, Order::Ascending)
+                .collect::<StdResult<Vec<(u64, BetaProposal)>>>()?;
```
ezekiiel, line 762: To save gas, how would you feel about getting the `count` from this? Getting the length of a vector is constant time, whereas the way above is `O(N)` I think.

contracts/cw-proposal-single/src/contract.rs
ezekiiel: Maybe for these where we don't actually know the time we could just set it to zero. That way the frontend could tell the difference between a proposal actually created at this time, and one that was migrated.

Could also put the block height there. Wouldn't be surprised if the block height version of a timestamp was LONG before Juno was created and that way a frontend could do this calculation for us.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+
+        /// Information about the deposit that was sent as part of this
+        /// proposal. None if no deposit.
+        pub deposit_info: Option<CheckedDepositInfo>,
+    }
+
+    match msg {
+        MigrateMsg::FromBeta {} => {
+            // Retrieve current map from storage
+            let current_map: Map<u64, BetaProposal> = Map::new("proposals");
```
ezekiiel, line 754: This is so megamind.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,90 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+                        allow_revoting: prop.allow_revoting,
+                        deposit_info: prop.deposit_info,
+                        created: env.block.time,
+                        last_updated: env.block.time,
+                    };
+
+                    PROPOSALS.save(deps.storage, id, &migrated_proposal)?;
+
+                    // Remove entry from old map.
+                    current_map.remove(deps.storage, id);
```
ezekiiel, line 790: Really wish we understood how storage gas costs worked. Part of me is like "why not just leave this around in storage"? We're not reading from it. 🤷‍♂️

Asked a question in the CosmWasm discord related to this today, hopefully we can figure it out. https://discord.com/channels/737637324434833438/737643344712171600/1001298045004161065

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -4196,3 +4238,57 @@ fn test_return_deposit_to_dao_on_proposal_failure() {
     // Deposit should now belong to the DAO.
     assert_eq!(balance.balance, Uint128::new(1));
 }
+
+#[test]
+fn test_migrate_from_beta() {
```
ezekiiel, line 4243: Same as other PR - might be nice to make this an integration test. No biggie if not though.

ezekiiel (commented):
Thanks so much for your care making sure this gets landed and all your work figuring out how CosmWasm storage works. Looks great. :) Just a couple questions and then we can merge.

contracts/cw-proposal-single/src/contract.rs
Callum-A: +1

contracts/cw-proposal-single/src/msg.rs
```diff
@@ -191,4 +191,7 @@ pub enum QueryMsg {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
-pub struct MigrateMsg {}
+pub enum MigrateMsg {
+    FromBeta,
+    FromCompatible,
```
Callum-A, line 196: just to make sure i understand FromCompatible is for a `v1` DAO that just missed these updates?

contracts/cw-proposal-single/src/contract.rs
de-husk: We can profile the gas usage with varying numbers of proposals here. It would be really useful if we could just find the max and average number of proposals in mainnet so we can use real numbers. Not sure if that's easy to get from the indexer or somewhere.

contracts/cw-proposal-single/src/contract.rs
janitachalam: I think @NoahSaso had a PR recently that dealt with this if you have any thoughts!

contracts/cw-proposal-single/src/contract.rs
janitachalam: agreed

contracts/cw-proposal-single/src/msg.rs
janitachalam: from compatible is intended for any non-state changing code migration!

contracts/cw-proposal-single/src/contract.rs
de-husk: Here are the top 5 proposal counts for daos in mainnet right now (thanks for the raw data @ezekiiel):
```
5
5
5
7
16
```

So, we can just test the gas usage for 20 proposals, to know the worst case scenario here.

de-husk: Also, the 16 is 🥇 🐾 dog DAO 🐾  🥇

janitachalam: update, I was able to migrate at least 220 proposals at a time without running out of gas (unable to determine true limit because I ran out of junox and my local docker container keeps crashing) but I think we are good to proceed without pagination for now

contracts/cw-proposal-single/src/contract.rs
janitachalam: fair point! it might also future-proof in case we want to read the old data later for some reason. Not sure if there's an official recommendation from them on this scenario, I can ask

contracts/cw-proposal-single/src/contract.rs
janitachalam: hm but I do need to access the proposals in the old map so I can create the new structs -- unless you meant something else?

contracts/cw-proposal-single/src/contract.rs
janitachalam: have not received an answer. I'm going to delete this line.

contracts/cw-proposal-single/src/contract.rs
janitachalam: wasn't able to find a way to query for block time given height in cosmwasm, so going to set it to 0 for now.

contracts/cw-proposal-single/src/contract.rs
Callum-A: I believe using .count() is linear here, we have encountered gas issues with it with howl

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -711,9 +713,70 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+                        msgs: prop.msgs,
+                        status: prop.status,
+                        votes: prop.votes,
+                        allow_revoting: prop.allow_revoting,
+                        deposit_info: prop.deposit_info,
+                        // AFAIK, Cosmwasm does not expose a way to query the timestamp
+                        // of a block given block height.
+                        // We assign migrated proposals a created timestamp of 0 so that the frontend may
+                        // query for the true timestamp given _height.
+                        created: Timestamp::from_seconds(0),
```
Callum-A, line 766: I agree this is the best way, could also argue to set it to the current block time? Although that may mess up sorting by time

contracts/cw-proposal-single/src/tests.rs
```diff
@@ -3638,6 +3663,81 @@ fn test_migrate() {
...
+            title: "Text proposal".to_string(),
+            description: "This is a simple text proposal".to_string(),
+            msgs: vec![],
+        },
+        &[],
+    )
+    .unwrap();
+
+    // Migrate from beta
+    app.execute(
```
Callum-A, line 3714: very noice

Callum-A (approved):

contracts/cw-proposal-single/src/contract.rs
janitachalam: yeah this way the frontend can distinguish which have been migrated

contracts/cw-proposal-single/src/contract.rs
janitachalam: the reason I had to collect here was because the closure below requires unique access to deps

Callum-A (approved):

contracts/cw-proposal-single/src/contract.rs
ezekiiel: Sounds good! Thanks for the thoughts. :)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -726,9 +728,71 @@ pub fn query_info(deps: Deps) -> StdResult<Binary> {
...
+                        total_power: prop.total_power,
+                        msgs: prop.msgs,
+                        status: prop.status,
+                        votes: prop.votes,
+                        allow_revoting: prop.allow_revoting,
+                        deposit_info: prop.deposit_info,
+                        // AFAIK, Cosmwasm does not expose a way to query the timestamp
+                        // of a block given block height.
+                        // We assign migrated proposals a created timestamp of 0 so that the frontend may
+                        // query for the true timestamp given _height.
```
ezekiiel, line 781:
```suggestion
                        // Cosmwasm does not expose a way to query the timestamp
                        // of a block given block height. As such, we assign migrated
                        // proposals a created timestamp of 0 so that the frontend may
                        // query for the true timestamp given `start_height`.
```

contracts/cw-proposal-single/src/msg.rs
```diff
@@ -207,4 +207,7 @@ pub enum QueryMsg {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
-pub struct MigrateMsg {}
+pub enum MigrateMsg {
+    FromBeta,
```
ezekiiel, line 211:
```suggestion
    FromV1,
```

contracts/cw-proposal-single/src/state.rs
```diff
@@ -61,7 +61,7 @@ pub struct Config {
 pub const CONFIG: Item<Config> = Item::new("config");
 /// The number of proposals that have been created.
 pub const PROPOSAL_COUNT: Item<u64> = Item::new("proposal_count");
-pub const PROPOSALS: Map<u64, SingleChoiceProposal> = Map::new("proposals");
+pub const PROPOSALS: Map<u64, SingleChoiceProposal> = Map::new("proposals_v1");
```
ezekiiel, line 64:
```suggestion
pub const PROPOSALS: Map<u64, SingleChoiceProposal> = Map::new("proposals_v2");
```

PR #418, opened 2022-07-25
Add test for SnapshotMap behavior during unstaking.

ezekiiel:
Wrote this during a fit of paranoia this morning. Everything is OK and works as expected. Figured we may as well add the test.


Callum-A (approved):

de-husk (approved):

de-husk:
>Wrote this during a fit of paranoia this morning

This is how the best tests are born

PR #420, opened 2022-07-26
Staking denom voting

### LINKED ISSUE #415: Native token voting contract where the native token is the native staking denom
So we can allow sub DAOs where we use the main token of the chain. This can be achieved by querying the balance of the staking module on chain, for example for juno:
```rust
let total_bonded_amount = deps
    .querier
    .query_balance("juno1fl48vsnmsdzcv85q5d2q4z5ajdha8yu3rf257t", "ujuno")?
    .amount;
```

This will give the bonded amount. This allows us to create native token DAOs without needing the user to do anything else!

User story:

I as a user want to create a DAO where all existing network participants may vote using the staking token of the chain.

Callum-A:
Closes #415 
big up pupmos


contracts/cw-staking-denom-voting/src/contract.rs
```diff
@@ -0,0 +1,119 @@
...
+        .iter()
+        .filter_map(|d| {
+            if d.amount.denom == denom {
+                Some(d.amount.amount)
+            } else {
+                None
+            }
+        })
+        .reduce(|a, b| a.checked_add(b).unwrap())
+        .unwrap_or_default();
```
JakeHartnell, line 78: Silly we have to do this but, it's fine.

JakeHartnell (approved):
Nice and simple, but powerful! Thanks for this!

contracts/cw-staking-denom-voting/src/contract.rs
```diff
@@ -0,0 +1,119 @@
...
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let validated_staking_module = deps.api.addr_validate(&msg.staking_module_address)?;
+
+    STAKING_MODULE.save(deps.storage, &validated_staking_module)?;
```
ezekiiel, line 25: Can we do any additional validation here? This is fine, but would be nice to be able to do a proper check.

ezekiiel (commented):
Looks great. Just a couple questions about validation. Little paranoid about configuration errors this week.

contracts/cw-staking-denom-voting/src/contract.rs
Callum-A: Yeah I need to look into this, but initially my answer is no. The sort of thing here would be front end hard coded for the chain deployed on.

contracts/cw-staking-denom-voting/src/contract.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+pub fn query_voting_power_at_height(
+    deps: Deps,
+    env: Env,
+    address: String,
+    height: Option<u64>,
+) -> StdResult<VotingPowerAtHeightResponse> {
+    // We can ignore height as we are protected by the chain's unstaking
+    // duration
```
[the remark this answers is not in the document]
Callum-A: Ah yep, I like this. I'll need to look into decoding the GetConfig resp in the best way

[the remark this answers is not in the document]
ezekiiel: Cool! May be able to do something like what we did for ics721: https://github.com/public-awesome/ics721/blob/f031e3fe07626cd353c9b27968df6b20eb721851/contracts/cw-ics721-bridge/src/state.rs#L23-L30

ezekiiel (approved):
:shipit:

PR #423, opened 2022-07-28
Fix deploy local script to use latest cw20_stake contract name

de-husk:
This contract was renamed and it broke deploy_local.sh.


janitachalam (approved):

ezekiiel (approved):
Thank you!

PR #426, opened 2022-07-28
Allow setting owner as instantiator in native token staking.

### LINKED ISSUE #419: cw-native-staked-balance-voting should allow setting the contract owner to the instantiator.
Without this we can not set the DAO as the owner or manager during instantiation of cw-core.

https://github.com/DA0-DA0/dao-contracts/blob/c37d7d76713e27f2ba86b41b31810b8279022237/contracts/cw721-stake/src/msg.rs#L10-L20

ezekiiel:
Resolves #419

Also runs `cargo update` to get our dependencies up to date and adds the native token staking voting module to the test suite for `cw-proposal-single` so we've got complete integration tests for that.

### DIFF
--- a/contracts/cw-native-staked-balance-voting/src/msg.rs
+++ b/contracts/cw-native-staked-balance-voting/src/msg.rs
@@ -4,10 +4,22 @@ use cw_utils::Duration;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum Owner {
+    /// Set the owner to a specific address.
+    Addr(String),
+    /// Set the owner to the address that instantiates this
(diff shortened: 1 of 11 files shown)


janitachalam (approved):

contracts/cw-native-staked-balance-voting/src/msg.rs
```diff
@@ -4,10 +4,22 @@ use cw_utils::Duration;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum Owner {
```
Callum-A, line 9: I think we should move this to a package soon

Callum-A (approved):

JakeHartnell (approved):

contracts/cw-native-staked-balance-voting/src/msg.rs
JakeHartnell: Would be good to have a look through our code and find more things we should put into packages at some point.

PR #441, opened 2022-08-09
add timestamps to multiple choice proposals 

### LINKED ISSUE #409: Add timestamps to multiple choice proposals
mirrors the changes to cw-proposal-single

blue-note:
closes #409

### DIFF
--- a/contracts/cw-proposal-multiple/src/contract.rs
+++ b/contracts/cw-proposal-multiple/src/contract.rs
@@ -182,6 +182,8 @@ pub fn execute_propose(
             allow_revoting: config.allow_revoting,
             deposit_info: config.deposit_info.clone(),
             choices: checked_multiple_choice_options,
+            created: env.block.time,
+            last_updated: env.block.time,
         };
         // Update the proposal's status. Addresses case where proposal
         // expires on the same block as it is created.
@@ -350,6 +352,9 @@ pub fn execute_execute(
     }
 
     prop.status = Status::Executed;
+    // Update proposal's last updated timestamp.
+    prop.last_updated = env.block.time;
+
     PROPOSALS.save(deps.storage, proposal_id, &prop)?;
 
     let refund_message = match &prop.deposit_info {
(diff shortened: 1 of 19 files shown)


contracts/cw-proposal-multiple/src/proposal.rs
```diff
@@ -35,6 +35,10 @@ pub struct MultipleChoiceProposal {
     /// Information about the deposit that was sent as part of this
     /// proposal. None if no deposit.
     pub deposit_info: Option<CheckedDepositInfo>,
+    /// The timestamp at which this proposal was created.
+    pub created: Timestamp,
+    /// The timestamp at which this proposal's status last changed (e.g. Passed, Executed).
```
ezekiiel, line 40: Could we add a note here about how quorum expiration stuff makes this a little strange? Technically, it's possible for a proposal to become "passed" without this being updated so we don't want folks relying on this to be up to date.

ezekiiel: Hmm so looking at this more it looks like we currently don't update this when passed anywhere. Is that intentional? If not, should we be updating this in `execute_vote` if it causes the proposal to pass?

ezekiiel (commented):
Looks good! One question about updating status on pass.

contracts/cw-proposal-multiple/src/proposal.rs
janitachalam: so the timestamp is updated in `update_status` which is called by `execute_vote,` I think that should suffice?

ezekiiel: Oh I see. Yeah that's great!

ezekiiel (approved):
Awesome work! Thank you!

de-husk (commented):
Looks like you need to regenerate the schema, since you updated the `contracts/cw-proposal-single` doc comments.

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -414,6 +419,8 @@ pub fn execute_close(
     };
 
     prop.status = Status::Closed;
+    // Update proposal's last updated timestamp.
+    prop.last_updated = env.block.time;
```
de-husk, line 423: I see that we did the same thing in the single-choice PR, but Im just wondering why we didn't do this `prop.last_updated` update here as well?: https://github.com/DA0-DA0/dao-contracts/blob/main/contracts/cw-proposal-multiple/src/contract.rs#L750

de-husk (approved):

PR #442, opened 2022-08-09
NETA DAO custom proposal module review.

ezekiiel:
NOT FOR MERGING.

NETA DAO is doing some custom work having a "council" that will be able to veto proposals. Opening this so we can review their changes and make sure that the first DAO to use a custom governance setup doesn't have bugs. :)

### DIFF
--- a/contracts/cw-proposal-single/src/contract.rs
+++ b/contracts/cw-proposal-single/src/contract.rs
@@ -30,7 +30,7 @@ use crate::{
     state::{Ballot, BALLOTS, CONFIG, PROPOSALS, PROPOSAL_COUNT, PROPOSAL_HOOKS, VOTE_HOOKS},
 };
 
-const CONTRACT_NAME: &str = "crates.io:cw-govmod-single";
+const CONTRACT_NAME: &str = "crates.io:cw-govmod-single-plus";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[cfg_attr(not(feature = "library"), entry_point)]
@@ -61,6 +61,7 @@ pub fn instantiate(
         dao: dao.clone(),
         deposit_info,
         allow_revoting: msg.allow_revoting,
+        executor_addr: msg.executor_addr,
     };
 
     // Initialize proposal count to zero so that queries return zero
@@ -87,6 +88,7 @@ pub fn execute(
             msgs,
         } => execute_propose(deps, env, info.sender, title, description, msgs),
         ExecuteMsg::Vote { proposal_id, vote } => execute_vote(deps, env, info, proposal_id, vote),
+        ExecuteMsg::Veto { proposal_id } => execute_veto(deps, env, info, proposal_id),
         ExecuteMsg::Execute { proposal_id } => execute_execute(deps, env, info, proposal_id),
         ExecuteMsg::Close { proposal_id } => execute_close(deps, env, info, proposal_id),
         ExecuteMsg::UpdateConfig {
@@ -118,6 +120,7 @@ pub fn execute(
         ExecuteMsg::RemoveVoteHook { address } => {
             execute_remove_vote_hook(deps, env, info, address)
         }
+        ExecuteMsg::AssignExecutor { address } => execute_assign_executor(deps, env, info, address),
     }
 }
 
@@ -182,6 +185,7 @@ pub fn execute_propose(
             deposit_info: config.deposit_info.clone(),
             created: env.block.time,
             last_updated: env.block.time,
+            vetoed: false,
         };
         // Update the proposal's status. Addresses case where proposal
         // expires on the same block as it is created.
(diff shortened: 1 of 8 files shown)


contracts/cw-proposal-single/src/contract.rs
```diff
@@ -53,6 +53,11 @@ pub fn instantiate(
     let (min_voting_period, max_voting_period) =
         validate_voting_period(msg.min_voting_period, msg.max_voting_period)?;
 
+    let executor_addr: Option<Addr> = match msg.executor_addr {
+        None => None,
+        Some(addr) => Some(deps.api.addr_validate(&addr)?),
+    };
```
ezekiiel, line 59:
```suggestion
    let executor_addr = msg.executor_addr.map(|human| deps.api.addr_validate(&human)).transpose()?;
```

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -383,6 +392,54 @@ pub fn execute_vote(
...
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    if config.executor_addr.is_none() {
+        return Err(ContractError::NoExecutorAssigned {});
+    }
+
+    if config.executor_addr.unwrap() != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
```
ezekiiel, line 410:
```suggestion
    if config.executor_addr != Some(info.sender) {
        return Err(ContractError::Unauthorized {});
    }
```

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -587,6 +645,36 @@ pub fn execute_remove_vote_hook(
...
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    address: Option<String>,
+) -> Result<Response, ContractError> {
+    let mut config = CONFIG.load(deps.storage)?;
+
+    // Only the DAO via prop OR the executor directly may call this method.
+    if info.sender == config.dao
+        || (config.executor_addr.is_some() && config.executor_addr.unwrap() == info.sender)
```
ezekiiel, line 658:
```suggestion
        || (config.executor_addr == Some(info.sender))
```

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -587,6 +645,36 @@ pub fn execute_remove_vote_hook(
...
+    {
+        // executor can be removed or reassigned to another address
+        let new_executor_addr: Option<Addr> = match address {
+            None => None,
+            Some(addr_str) => Some(deps.api.addr_validate(&addr_str).unwrap()),
+        };
```
ezekiiel, line 664: Suggest doing same `.map().transpose()` maneuver as suggested above.

Reason being: generally speaking the people writing the standard library know a lot more about how to write good, high performance Rust than we do. Where we can use their stuff, we should. :)

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -596,7 +684,7 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
             query_list_proposals(deps, env, start_after, limit)
         }
         QueryMsg::ProposalCount {} => query_proposal_count(deps),
-        QueryMsg::GetVote { proposal_id, voter } => query_vote(deps, proposal_id, voter),
+        QueryMsg::Vote { proposal_id, voter } => query_vote(deps, proposal_id, voter),
```
ezekiiel, line 687: @NoahSaso should they be supporting the v1 or v2 interface to be supported by the frontend? This module has a custom name, but is compatible with `cw-proposal-single`. Perhaps they should keep the name the same?

contracts/cw-proposal-single/src/contract.rs
onewhiskeypls: > 

good to know, ill keep that in mind next time!

contracts/cw-proposal-single/src/contract.rs
onewhiskeypls: Note: i branched off main before the autoclose failed execution was merged in. it included the failed proposal refund work

JakeHartnell:
Looking good. I was writing a custom proposal module on the plane and it made me wonder if we should be making these a bit more extensible, so there is less copy pasta and code that may potentially become outdated.

We could do something similar to [cw-nfts](https://github.com/CosmWasm/cw-nfts) and use trait based extension for the Instantiate, Execute, and Query messages as well as the config. Maybe overkill?

ezekiiel:
> @JakeHartnell: We could do something similar to [cw-nfts](https://github.com/CosmWasm/cw-nfts) and use trait based extension for the Instantiate, Execute, and Query messages as well as the config.

I actually think this is a great idea! The design there has really grown on me and I think supporting stuff like this more would be great.

ezekiiel:
Closing for now as I think we've gotten good feedback / learning.

PR #450, opened 2022-08-15
Improve documentation about generating schema files.

ezekiiel:
Based on feedback from Whisky.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,12 +64,15 @@ execute messages. We use those schema files to generate Typescript
 code for interacting with those contracts via
 [ts-codegen](https://github.com/CosmWasm/ts-codegen). This generated
 Typescript code is then [used in the
-UI](https://github.com/DA0-DA0/dao-dao-ui/tree/40f3cbfe676a98bf7b9db7b646e74e5b2dae4502/packages/state/clients).
+UI](https://github.com/DA0-DA0/dao-dao-ui/tree/40f3cbfe676a98bf7b9db7b646e74e5b2dae4502/packages/state/clients)
+to interact with our contracts. Generating this code means that
+frontend developers don't have to manually write query and execute
+messages for each method on our contracts.
 
 To make sure these files are generated correctly, make sure to add any
 new data types used in contract messages to `examples/schema.rs`.
 
-If you are adding a new query ts-codegen expects:
+If you are adding a new query, ts-codegen expects:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -55,25 +55,64 @@ From the repository root.
...
 
-To generate schema files, run:
+### Generating schema files
 
+We generate JSON schema files for all of our contracts' query and
+execute messages. We use those schema files to generate Typescript
+code for interacting with those contracts via
+[ts-codegen](https://github.com/CosmWasm/ts-codegen). This generated
+Typescript code is then [used in the
+UI](https://github.com/DA0-DA0/dao-dao-ui/tree/40f3cbfe676a98bf7b9db7b646e74e5b2dae4502/packages/state/clients).
```
elsehow, line 67: "Used in the UI to...." Important to tell people why they need to do this.

CONTRIBUTING.md
```diff
@@ -55,25 +55,64 @@ From the repository root.
...
+
+To make sure these files are generated correctly, make sure to add any
+new data types used in contract messages to `examples/schema.rs`.
+
+If you are adding a new query ts-codegen expects:
```
elsehow, line 72: > If you are adding a new query, `ts-codegen` expects:

elsehow (approved):
Looks great! A few minor comments. Otherwise, ready to go!

CONTRIBUTING.md
ezekiiel: Great feedback. Added a blurb about frontend folks not needing to manually write query and execute messages.

blue-note (approved):
Awesome! This is very helpful

CONTRIBUTING.md
```diff
@@ -55,25 +55,67 @@ From the repository root.
 Before making a PR, you'll need to do two things to get CI passing:
 
 1. Generate schema files for the contracts.
```
blue-note, line 57: should we also make a note about running `cargo schema`?

ezekiiel: I think this is covered by the earlier section that mentions running `scripts/schema.sh` which does that under the hood.

ezekiiel: oh shit wait you're right the order is wrong..

ezekiiel: or wait no this is correct. This section is giving an overview, but the end result is just prompting the reader to run the `schema.sh` script.

PR #453, opened 2022-08-17
added subdao qry and exec msgs

### LINKED ISSUE #424: Add queries / methods for setting / listing subDAOs.
Talking to @de-husk this morining:

Need to add some formal queries / a method for registering and listing subDAOs in order to be in line with the designs. The current plan is to just do this via an [item](https://github.com/DA0-DA0/dao-contracts/wiki/DAO-DAO-v1-Contracts-Design#items) but this seems sub-optimal.

One way this may be implemented:

- `struct SubDAO { addr: Addr, charter: Option<String> }`
- `UpdateSubDaos { to_add: Vec<SubDAO>, to_remove: Vec<String> }` an execute method only callable by the core contract. Modifies the list of subDAOs. `to_remove` only needs addresses to do the removal.
- `ListSubDaos { start_after: String, limit: u32 }` a query for listing the subDAOs that have been registered.
- `const SUBDAOS: Map<Addr, SubDAO> = Map::new("sub_daos")`

I suggest having an optional charter for subDAOs. When a subDAO is created (via a template) the DAO may set a charter which is the purpose of the subDAO (kinda like a constitution for the subDAO). We can show this in the subDAO listing in the UI to make it easier for folks to understand the org structure of a DAO when viewing it.

onewhiskeypls:
Issue #424 .  See also cw-paginate modification for map to handle refs @ezekiiel

Closes #424


contracts/cw-core/src/contract.rs
```diff
@@ -705,6 +737,34 @@ pub fn query_cw20_balances(
...
+
+    let subdaos = subdaos
+        .into_iter()
+        .map(|(address, charter)| SubDao {
+            addr: address.into_string(),
+            charter,
+        })
+        .collect();
+
+    to_binary(&ListSubDaosResponse { subdaos })
```
ezekiiel, line 765: How would you feel about returning `to_binary(subdaos)` here? Would match the other queries and what you have in the `schema.rs` file above.

contracts/cw-core/src/query.rs
```diff
@@ -56,3 +56,17 @@ pub struct AdminNominationResponse {
     /// pending.
     pub nomination: Option<Addr>,
 }
+
+/// Returned by the `ListSubDaos` qry
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+pub struct ListSubDaosResponse {
+    pub subdaos: Vec<SubDao>,
+}
```
ezekiiel, line 64: Likely can remove this if we go the `Vec<SubDao>` way.

contracts/cw-core/src/state.rs
```diff
@@ -61,3 +61,6 @@ pub const CW20_LIST: Map<Addr, Empty> = Map::new("cw20s");
 /// Set of cw721 tokens that have been registered with this contract's
 /// treasury.
 pub const CW721_LIST: Map<Addr, Empty> = Map::new("cw721s");
+
+/// List of SubDAOs associated to this DAO
+pub const SUBDAO_LIST: Map<&Addr, Option<String>> = Map::new("sub_daos");
```
ezekiiel, line 66: Little comment explaining what this field is for:

```suggestion
/// List of SubDAOs associated to this DAO. Each SubDAO has an optional charter.
pub const SUBDAO_LIST: Map<&Addr, Option<String>> = Map::new("sub_daos");
```

contracts/cw-core/src/tests.rs
```diff
@@ -2354,3 +2354,108 @@ fn test_execute_stargate_msg() {
...
+    );
+
+    let to_add: Vec<SubDao> = vec![
+        SubDao {
+            addr: "subdao001".to_string(),
+            charter: None,
+        },
+        SubDao {
+            addr: "subdao002".to_string(),
+            charter: Some("cool charter bro".to_string()),
```
ezekiiel, line 2378: 😎

ezekiiel (commented):
This looks perfect. :) Great work. Just one little comment about the return type of the query. Either need to update `schema.rs` to use the `ListSubDaoResponse` type, or just need to return `Vec<SubDao>` from the query.

contracts/cw-core/src/contract.rs
onewhiskeypls: updated

onewhiskeypls:
@ezekiiel this should be ready for review and all issues should be correct now!

contracts/cw-core/src/contract.rs
```diff
@@ -478,6 +481,32 @@ pub fn execute_remove_item(
...
+    if env.contract.address != sender {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    for addr in to_remove {
+        let addr = deps.api.addr_validate(&addr)?;
+        SUBDAO_LIST.remove(deps.storage, &addr);
+    }
+
+    for subdao in to_add {
```
ezekiiel, line 500: Totally ignore this for this PR, but one cool style flourish you can do in Rust is:

```rust
for SubDao { addr, charter } in to_add {
  // ...
}
```

You may want to do this to flex in your coding interviews. ;)

ezekiiel (approved):
LFG 😎

JakeHartnell (approved):
:shipit:

PR #455, opened 2022-08-17
Add migration for `close_proposal_on_execution_failure` field.

### LINKED ISSUE #452: handle state migration for close_proposal_on_execution_failure
Similar to https://github.com/DA0-DA0/dao-contracts/pull/417

since config changed, need to update config in state when migrating

ezekiiel:
Resolves #452

This removes the non-mocked migration test. The test was effectively a
no-op as (so far as I am aware) we don't have a good way worked out to
mock v1 state. The new change errors if the `"config"` key isn't
present wheras the old one would just do nothing if the `"proposals"`
key was not present so the test had to go.

I tried mocking v1 state using the `AppBuilder` from multi-test and
instantiating it with a custom deps. From memory since I deleted the
code (so not real code):

```rust
let mut deps = mock_dependencies();
let v1_item: Item<V1Config> = Item::new("config");
let v1_config = V1Config { /* ... */ };
v1_item.save(&mut deps.storage, &v1_config).unwrap();
let mut app =
AppBuilder::default().with_storage(deps.storage).build(|_, _, _| ());
```

This doesn't work which makes some amount of sense as you'd expect
storage to be set on a per-contract level anyhow. There may be
something here we could explore.

### DIFF
--- a/contracts/cw-proposal-single/src/tests.rs
+++ b/contracts/cw-proposal-single/src/tests.rs
@@ -9,7 +9,7 @@ use cw20_staked_balance_voting::msg::ActiveThreshold;
 use cw_multi_test::{next_block, App, BankSudo, Contract, ContractWrapper, Executor, SudoMsg};
 
 use cw_core::{msg::ModuleInstantiateInfo, state::ProposalModule};
-use cw_storage_plus::Map;
+use cw_storage_plus::{Item, Map};
 use cw_utils::Duration;
 use cw_utils::Expiration;
 
@@ -37,7 +37,7 @@ use crate::{
 const CREATOR_ADDR: &str = "creator";
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
-struct BetaProposal {
+struct V1Proposal {
     pub title: String,
     pub description: String,
     pub proposer: Addr,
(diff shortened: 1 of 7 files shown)


contracts/cw-proposal-single/src/tests.rs
```diff
@@ -3847,82 +3858,6 @@ fn test_migrate_from_compatible() {
     assert_eq!(config, new_config);
 }
 
-#[test]
-fn test_migrate_from_beta() {
```
blue-note, line 3851: yeah, I don't think there's a good way to manipulate state with multi-test. Let's keep the mock test for now and add an integration test later

blue-note: made an issue: https://github.com/DA0-DA0/dao-contracts/issues/456

blue-note:
looks good to me! merged the PR this is based off of so feel free to rebase onto main. The only thing is to decide if we want to propagate close_proposal_on_execution_failure to the proposal struct re: this comment --  https://discord.com/channels/895922260047720449/895926400970801173/1009531095181697136. The impact this would have is that any proposals that are open when the contract is migrated will not pick up the close_proposal_on_execution_failure change; only proposals moving forward will. This matches the behavior for allow_revoting and deposit_info.

blue-note (approved):

JakeHartnell (approved):
LGTM!

PR #458, opened 2022-08-19
Integration test improvements + Fix deploy script

de-husk:
I was originally going to use `cw-optimizoor` instead of the cosmwasm `workspace-optimizer`. But when comparing the runtimes of `cw-optimizoor` and `workspace-optimizer` in CI,`cw-optimizoor` seems to be only about 1-2 minutes faster at building + optimizing (when using a pre-built `cw-optimizoor` binary):
- [optimizoor](https://github.com/DA0-DA0/dao-contracts/runs/7927041477?check_suite_focus=true)
- [workspace-optmizer](https://github.com/DA0-DA0/dao-contracts/runs/7927311270?check_suite_focus=true)

Since it is still generating larger final wasms than `workspace-optimzer` and needs a way to be distributable as a pre-built binary,  Im going to hold off on using `cw-optimizoor` in dao dao for now.

This PR contains some ci build-time and code improvements I made while doing these test runs.


ci/integration_tests/src/helpers/chain.rs
```diff
@@ -3,7 +3,6 @@ use cosm_orc::orchestrator::cosm_orc::CosmOrc;
 use cosm_orc::{
     config::cfg::Config, config::key::SigningKey, profilers::gas_profiler::GasProfiler,
 };
-use ctor::ctor;
```
de-husk, line 6: We really dont need this dependency, I just did the same ctor annotation functionality with `get_or_init()`.

.github/workflows/integration_tests.yml
```diff
@@ -7,44 +7,14 @@ on:
...
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - name: Optimize
-        run: optimize_workspace.sh
-      - name: Upload contracts
-        uses: actions/upload-artifact@v3
-        with:
-          name: dao-dao-contract-artifacts-${{ github.event.number }}
-          path: artifacts
-          retention-days: 1
```
de-husk, line 37: I removed this job because I dont want to have to recompile the workspace when we run the integration tests, and making it one job makes that easier. Also, we have to wait for the contracts to optimize either way, so it might as well just be one job.

de-husk: Also, while it's nice to have wasm builds uploaded, I realized that it eventually costs money (as opposed to the cache).

.github/workflows/integration_tests.yml
```diff
@@ -60,23 +30,16 @@ jobs:
             -p 1317:1317 \
             -p 26656:26656 \
             -p 26657:26657 \
+            -p 9090:9090 \
```
de-husk, line 33: cosm-orc needs the gRPC endpoint exposed now.

.github/workflows/integration_tests.yml
```diff
@@ -102,6 +82,7 @@ jobs:
         run: echo "GIT_BRANCH=$(echo ${{ github.ref }} | sed 's|/|-|g')" >> $GITHUB_ENV
 
       - name: Upload Gas Report
+        if: ${{ github.ref == 'refs/heads/main' }}
         uses: actions/upload-artifact@v3
         with:
           name: dao-dao-gas-report-${{ env.GIT_BRANCH }}
```
de-husk, line 88: This is now the only place we are using github action storage, and it's very small file (now only pushed in main commits), so we will stay in the free tier.

.github/workflows/integration_tests.yml
de-husk: Actually, nm it looks like storage is free for public repos. Well we still don't need this job.

.cargo/config
```diff
@@ -2,4 +2,5 @@
 wasm = "build --release --target wasm32-unknown-unknown"
 wasm-debug = "build --target wasm32-unknown-unknown"
 t = "test --workspace --exclude integration_tests"
-unit-test = "test --workspace --exclude integration_tests"
\ No newline at end of file
+unit-test = "test --workspace --exclude integration_tests"
+integration-test = "test --package integration_tests -- --ignored --test-threads 1"
```
de-husk, line 6: I decided to just ignore all integration tests, so that people who run `cargo test` from workspace root wont run them.

While the integration tests are a part of the workspace, I think this is the best approach for now.

JakeHartnell (approved):
LGTM!

.cargo/config
ezekiiel: nice!! thank you!

.github/workflows/integration_tests.yml
```diff
@@ -87,10 +49,31 @@ jobs:
...
+      # then we will get os permission errors trying to later build the workspace on a cache miss,
+      # because the workspace-optimizer docker container is creating those folders with incorrect perms
+      # if they dont already exist.
+      - name: Download smart contract dependencies
+        run: |
+          mkdir -p artifacts target
+          wget https://github.com/CosmWasm/cw-plus/releases/download/v0.11.1/cw20_base.wasm \
+            -O artifacts/cw20_base.wasm
+          wget https://github.com/CosmWasm/cw-plus/releases/download/v0.11.1/cw4_group.wasm \
+            -O artifacts/cw4_group.wasm
```
ezekiiel, line 65:
```suggestion
          wget https://github.com/CosmWasm/cw-plus/releases/latest/download/cw20_base.wasm \
            -O artifacts/cw20_base.wasm
          wget https://github.com/CosmWasm/cw-plus/releases/latest/download/cw4_group.wasm \
            -O artifacts/cw4_group.wasm
```

From the bottom of this page: https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases

Probably good to test against the latest stuff. 🤷‍♂️

.github/workflows/integration_tests.yml
```diff
@@ -87,10 +49,31 @@ jobs:
...
+
+      - name: Build + Optimize contracts
+        run: |
+          docker run --rm -v "$(pwd)":/code \
+            --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+            --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+            --platform linux/amd64 \
+            cosmwasm/workspace-optimizer:0.12.6
```
ezekiiel, line 73: How would you feel about using `latest` here as well?

```suggestion
            cosmwasm/workspace-optimizer:latest
```

README.md
```diff
@@ -38,13 +38,12 @@ be found in [CONTRIBUTING.md](./CONTRIBUTING.md).
...
 
-Run `cargo t` or `cargo unit-test` from the project root to run the unit tests.
+Run `cargo test` or `cargo unit-test` from the project root to run the unit tests.
 
 ### Integration tests
 
 * `./scripts/deploy_local.sh juno10j9gpw9t4jsz47qgnkvl5n3zlm2fz72k67rxsg`
-* `cd ci/integration_tests`
-* `CONTRACT_DIR="../../artifacts" GAS_OUT_DIR="gas_reports" CONFIG="configs/local.yaml" cargo t`
+* `RUST_LOG=info CONFIG="configs/local.yaml" cargo integration-test`
```
ezekiiel, line 46: Really appreciate you keeping this up to date. <3

ci/integration_tests/src/tests/cw_core_test.rs
```diff
@@ -16,6 +16,7 @@ use voting::{deposit::CheckedDepositInfo, threshold::PercentageThreshold, thresh
 
 #[test_context(Chain)]
 #[test]
+#[ignore]
```
ezekiiel, line 19: This is really cool. Didn't realize this was a pattern. Maybe we could add a little blurb to the README for the integration tests just linking to the docs? https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested

ezekiiel (approved):
Awesome! Didn't know `#[ignore]` was a pattern. Fits the need here really well. Also appreciate you fixing the `cargo test` thing. :)

.github/workflows/integration_tests.yml
de-husk: Im fine with it, but they do not have a `latest` tag published: https://hub.docker.com/r/cosmwasm/workspace-optimizer/tags?page=1&name=latest (latest is not a feature of docker its just a manual tag name that needs to be kept up sadly.

.github/workflows/integration_tests.yml
de-husk: Great catch, will do.

PR #461, opened 2022-08-22
Update contract naming scheme and cleanup docs.

ezekiiel:
Update contract naming scheme, and cleanup docs.

This updates the naming scheme for DAO DAO contracts. All contracts
are given the prefix `cw-dao`. The core module is renamed to be
`cw-dao-core`. Proposal modules are named in the format
`cw-dao-proposal-<proposal_type>` and voting modules are named in the
format `cw-dao-voting-<voting_type>`.

This also improves the README situation for many contracts.

Possibly controversially, this removes the `cw-named-groups` and
`cw-names-registry` contracts. The thinking being: our design
direction has changed since those were written. For the names
registry, I'm currently imagining that we will build out a much more
sophisticated system here when we do vote delegation that will replace
this contract. For named groups, we now support strings in
`cw-dao-core` items. This probably replaces the need for this contract
as frontends will likely just want to store this information in that
string.

Notably as well, this PR bumps our cargo requirement to the nightly
channel so that we may use the accepted-but-not-yet-mainline [cargo
workspace dependencies
changes](https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html). This
has less-than-ideal UX as we will have to describe to contributors
that they must either `rustup default nightly` or run `cargo +nightly`
to interact with the contracts. It comes with what I see as
significant advantages. In particular:

1. We can unify package versions across packages and contracts
   allowing us to enforce that we are using the same version of
   CosmWasm dependencies everywhere.
2. We can unify package metadata across contracts which makes
   maintaining shared metadata across packages simple.

I belive that (1) provides both ergonomic and security advantages as
we are unlikely to publish a package with outdated or insecure
dependencies.

It also allows us to enforce at the workspace level that packages are
using `version = "*"` when importing local contracts. Doing this is
important because it makes sure that we will detect breaking API
changes when updating one contract and keep our contracts working
together cohesively.

Overall, using workspace dependencies makes package versioning much
more cohesive at the expense of some additional cognative overhead
when first interacting with the contracts.

## Remaining TODO

1. Make sure README and `lib.rs` package level documentation is in
   sync for all crates.
2. Remove `debug/cw-dao-proposal-sudo` and
   `debug/cw-dao-voting-cw20-balance` contracts as they are no longer
   needed.
3. Investigate interaction between `workspace-optimizer` and cargo
   nightly requirement. This may require a PR into workspace
   optimizer.
4. Update typescript type generation code to handle nested contract
   directories (`contracts/proposal-modules` and
   `contracts/voting-moudles`).


Cargo.toml
```diff
@@ -1,6 +1,88 @@
...
+license = "Apache-2.0"
+repository = "https://github.com/DA0-DA0/dao-contracts"
+
+[workspace.dependencies]
+# Packages.
+cw-dao-core-interface = { path = "packages/cw-dao-core-interface", version = "*" }
+cw-dao-core-macros = { path = "packages/cw-dao-core-macros", version = "*" }
+cw-paginate = { path = "packages/cw-paginate", version = "*" }
+cw721-controllers = { path = "packages/cw721-controllers", version = "*" }
+indexable-hooks = { path = "packages/indexable-hooks", version = "*" }
```
JakeHartnell, line 28: Do we want to prefix some of these packages with `cw`?

Note sure if `voting` will be available on crates. :joy:

Cargo.toml
```diff
@@ -1,6 +1,88 @@
...
+
+# Contracts.
+cw-dao-core = { path = "contracts/cw-dao-core", features = ["library"], version = "*" }
+
+# Proposal modules.
+cw-dao-proposal-single = { path = "contracts/proposal-modules/cw-dao-proposal-single", features = ["library"], version = "*" }
+cw-dao-proposal-multiple = { path = "contracts/proposal-modules/cw-dao-proposal-multiple", features = ["library"], version = "*" }
+
+# Voting modules.
+cw-dao-voting-cw20-stake = { path = "contracts/voting-modules/cw-dao-voting-cw20-stake", features = ["library"], version = "*" }
```
JakeHartnell, line 42: `cw-dao-voting-staked-cw20`?

Cargo.toml
ezekiiel: Great point. :) How about:

`voting` -> `cw-dao-voting`
`indexable-hooks` -> `cw-indexable-hooks`
`proposal-hooks` -> `cw-dao-proposal-hooks`
`vote-hooks` -> `cw-dao-vote-hooks`
`proposal-module-testing` -> `cw-dao-proposal-module-testing`

Sets some pattern were we're doing `cw-` when it's CosmWasm relevant and `cw-dao` if it's DAO DAO relevant.

Cargo.toml
ezekiiel: oh yeah great idea. gave this that name because it uses the cw20-stake contract under the hood, but no need to leak implementation details in the name. :)

ezekiiel:
The cargo`workspace-inheritance` feature that we use here has been stabilized and will be released in the next Rust release: https://forge.rust-lang.org/#current-release-versions

contracts/cw-dao-core/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "cw-dao-core"
+authors = ["ekez <ekez@withoutdoing.com>"]
+description = "The core module of DAO DAO DAOs."
+license = { workspace = true }
+repository = { workspace = true }
+edition = { workspace = true }
+version = { workspace = true }
```
blue-note, line 8: what does `workspace` = true do?

ezekiiel: this just sets the fields to the same as the workspace `Cargo.toml` file. If we change those values at the root, it'll get propagated everywhere. :) This is nice because all of these fields should be the same for our contracts anyhow (i think).

ezekiiel:
Closing for now. I think we've landed on good names here.

I'm going to roll back using the cargo workspace-dependencies feature. Despite that feature being slated for stabilization in the next release, it is still in beta so using it would require that we use a beta version of the Rust compiler. I think we could make a compelling case for this being OK, but explaining that when we're deploying these contracts on permissioned chains may be politically sub-optimal.

Will re-open once we're closer to the v2 release again.

PR #464, opened 2022-08-23
Add "set contract version" to migrate methods 

blue-note:
I know how we do the versioning will change with [this PR](https://github.com/DA0-DA0/dao-contracts/pull/461), but let's also make sure we update the contract version in the migrate methods. This will enable us to start testing v2 contracts from frontend as well (checks the contract version)

### DIFF
--- a/ci/integration_tests/Cargo.toml
+++ b/ci/integration_tests/Cargo.toml
@@ -15,11 +15,11 @@ cw20 = "0.13"
 cw20-base = "0.13"
 cw-utils = "0.13"
 cosmwasm-std = { version = "1.0.0", features = ["ibc3"] }
-cw-core = { version = "0.1.0", path = "../../contracts/cw-core" }
+cw-core = { version = "*", path = "../../contracts/cw-core" }
 cw20-stake = { version = "*", path = "../../contracts/cw20-stake" }
-cw20-staked-balance-voting = { version = "0.1.0", path = "../../contracts/cw20-staked-balance-voting" }
-cw-proposal-single = { version = "0.1.0", path = "../../contracts/cw-proposal-single" }
-cw-core-interface = { version = "0.1.0", path = "../../packages/cw-core-interface" }
+cw20-staked-balance-voting = { version = "*", path = "../../contracts/cw20-staked-balance-voting" }
+cw-proposal-single = { version = "*", path = "../../contracts/cw-proposal-single" }
+cw-core-interface = { version = "*", path = "../../packages/cw-core-interface" }
 voting = { version = "*", path = "../../packages/voting" }
 
 assert_matches = "1.5"
--- a/ci/integration_tests/Cargo.toml
+++ b/ci/integration_tests/Cargo.toml
@@ -15,7 +15,7 @@ cw20 = "0.13"
 cw20-base = "0.13"
 cw-utils = "0.13"
 cosmwasm-std = { version = "1.0.0", features = ["ibc3"] }
-cw-core = { version = "0.1.0", path = "../../contracts/cw-core" }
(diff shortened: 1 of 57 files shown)


contracts/cw721-stake/src/contract.rs
```diff
@@ -525,3 +526,10 @@ pub fn query_staked_nfts(
 
     to_binary(&res)
 }
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
+    // Set contract to version to latest
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    Ok(Response::default())
+}
```
ezekiiel, line 535: How would you feel about adding some really simple tests for these? Could probably just write one like (pseudocode):

```
let mut deps = mock_dependencies();
cw2::set_contract_version(deps.storage, "mock", "mock").unwrap();
migrate(deps.as_mut(), mock_env(), MigrateMsg {}).unwrap();
let version = cw2::get_contract_version(deps.storage).unwrap();
assert_eq!(version.version, env!("CARGO_PKG_VERSION");
```

and copy it into all of the contracts. would catch if we forget about this again.

ezekiiel (commented):
Great catch!!

ezekiiel (approved):
Approving so you're not blocked on another review from me. :)

contracts/cw721-stake/src/contract.rs
blue-note: yess good call

types/yarn.lock
```diff
@@ -7,7 +7,7 @@
   resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
   integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
   dependencies:
-    "@jridgewell/gen-mapping" "^0.1.0"
```
de-husk, line 10: Looks like these `types/yarn.lock` changes were accidentally added through find and replace. We should remove these.

de-husk (changes requested):

types/yarn.lock
blue-note: whoops!

ci/integration_tests/Cargo.toml
```diff
@@ -15,11 +15,11 @@ cw20 = "0.13"
...
 cosmwasm-std = { version = "1.0.0", features = ["ibc3"] }
-cw-core = { version = "0.1.0", path = "../../contracts/cw-core" }
+cw-core = { version = "*", path = "../../contracts/cw-core" }
 cw20-stake = { version = "*", path = "../../contracts/cw20-stake" }
-cw20-staked-balance-voting = { version = "0.1.0", path = "../../contracts/cw20-staked-balance-voting" }
-cw-proposal-single = { version = "0.1.0", path = "../../contracts/cw-proposal-single" }
-cw-core-interface = { version = "0.1.0", path = "../../packages/cw-core-interface" }
+cw20-staked-balance-voting = { version = "*", path = "../../contracts/cw20-staked-balance-voting" }
+cw-proposal-single = { version = "*", path = "../../contracts/cw-proposal-single" }
+cw-core-interface = { version = "*", path = "../../packages/cw-core-interface" }
```
de-husk, line 22: Thanks for fixing this! Now that we are actually versioning the smart contracts, the versions here should all be the latest.

de-husk (approved):

PR #465, opened 2022-08-23
WIP New deposit features

### LINKED ISSUE #386: "Anyone can submit" configuration for proposal modules
Some DAOs may want to allow non-members to submit proposals.

Use case: Dog DAO wants to let anyone propose to add their dog to Dog DAO, users submit a deposit and can put up a vote to add their dog and mint a dog token.

### LINKED ISSUE #387: Support using native tokens for proposal deposits
Some DAOs may want to use native tokens like Juno for proposal deposits.

This would work well for more open "anyone can submit" style DAOs like in #386 to help fight spam or as part of the process for raising a treasury.

### LINKED ISSUE #388: Pay to propose
Currently we have two settings for proposal deposits:
- Fully refundable (even if it fails)
- Not refundable if it fails

We should add a third option: _Not refundable. Pay to propose_

This is useful for certain types of DAOs that may have a more open "anyone can propose" aspect.

JakeHartnell:
Refactor proposal deposits to support new features.

- Closes #388
- Closes #386
- Closes #387

DON'T MERGE. We may choose a different approach.

### DIFF
--- a/ci/integration_tests/src/helpers/helper.rs
+++ b/ci/integration_tests/src/helpers/helper.rs
@@ -8,7 +8,9 @@ use cw_core::{
 };
 use cw_utils::Duration;
 use voting::{
-    deposit::DepositInfo, deposit::DepositToken, threshold::PercentageThreshold,
+    deposit::DepositInfo,
+    deposit::{DepositRefundPolicy, DepositToken},
+    threshold::PercentageThreshold,
     threshold::Threshold,
 };
 
@@ -71,9 +73,10 @@ pub fn create_dao(
                 deposit_info: Some(DepositInfo {
                     token: DepositToken::VotingModuleToken {},
                     deposit: Uint128::new(1000000000),
-                    refund_failed_proposals: true,
+                    refund_policy: DepositRefundPolicy::OnlyPassed,
                 }),
                 close_proposal_on_execution_failure: false,
+                open_proposal_submission: false,
             })?,
             admin: Admin::CoreContract {},
             label: "DAO DAO Proposal Module".to_string(),
--- a/contracts/cw-proposal-multiple/src/contract.rs
+++ b/contracts/cw-proposal-multiple/src/contract.rs
@@ -12,7 +12,7 @@ use proposal_hooks::{new_proposal_hooks, proposal_status_changed_hooks};
 
 use vote_hooks::new_vote_hooks;
 use voting::{
-    deposit::{get_deposit_msg, get_return_deposit_msg, DepositInfo},
+    deposit::{get_deposit_msg, get_return_deposit_msg, DepositInfo, DepositRefundPolicy},
     proposal::{DEFAULT_LIMIT, MAX_PROPOSAL_SIZE},
     reply::{mask_proposal_execution_proposal_id, TaggedReplyId},
     status::Status,
@@ -65,6 +65,7 @@ pub fn instantiate(
         dao,
         deposit_info,
         close_proposal_on_execution_failure: msg.close_proposal_on_execution_failure,
+        open_proposal_submission: msg.open_proposal_submission,
     };
 
     // Initialize proposal count to zero.
@@ -101,6 +102,7 @@ pub fn execute(
             dao,
             deposit_info,
             close_proposal_on_execution_failure,
+            open_proposal_submission,
         } => execute_update_config(
             deps,
             info,
@@ -112,6 +114,7 @@ pub fn execute(
             dao,
             deposit_info,
             close_proposal_on_execution_failure,
+            open_proposal_submission,
         ),
         ExecuteMsg::AddProposalHook { address } => {
             execute_add_proposal_hook(deps, env, info, address)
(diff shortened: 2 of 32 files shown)


ci/integration_tests/src/helpers/helper.rs
```diff
@@ -71,9 +73,10 @@ pub fn create_dao(
                 deposit_info: Some(DepositInfo {
                     token: DepositToken::VotingModuleToken {},
                     deposit: Uint128::new(1000000000),
-                    refund_failed_proposals: true,
+                    refund_policy: DepositRefundPolicy::OnlyPassed,
```
ezekiiel, line 76: Love this refactor. Probably should have been using an enum the whole time lol

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -428,17 +442,14 @@ pub fn execute_close(
...
-                &config.dao
-            };
-            get_return_deposit_msg(deposit_info, receiver)?
-        }
+        Some(deposit_info) => match deposit_info.refund_policy {
+            // If we are always refunding deposits, funds go to proposer
+            DepositRefundPolicy::Always => get_return_deposit_msg(deposit_info, &prop.proposer)?,
+            // If no refunds, or failed proposals not refund, funds go to DAO
+            _ => get_return_deposit_msg(deposit_info, &config.dao)?,
+        },
```
ezekiiel, line 452: This is so much more clean.

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -413,16 +428,12 @@ pub fn execute_close(
...
-                // the depost to the DAO treasury on close.
-                &config.dao
-            };
-            get_return_deposit_msg(deposit_info, receiver)?
-        }
+        Some(deposit_info) => match deposit_info.refund_policy {
+            // If we are always refunding deposits, funds go to proposer
+            DepositRefundPolicy::Always => get_return_deposit_msg(deposit_info, &prop.proposer)?,
+            // If no refunds, or failed proposals not refund, funds go to DAO
+            _ => get_return_deposit_msg(deposit_info, &config.dao)?,
```
ezekiiel, line 435: These match statements might be nice to actually fully write out. It's probably a desirable feature that adding a new variant will cause a compile error and force us to check this logic again. I do love how clean this is though lol

```suggestion
            // If no refunds, or failed proposals not refund, funds go to DAO
            DepositRefundPolicy::Never | DepositRefundPolicy::OnlyPassed => get_return_deposit_msg(deposit_info, &config.dao)?,
```

packages/voting/src/deposit.rs
```diff
@@ -61,20 +73,33 @@ impl DepositInfo {
...
+                    .query_wasm_smart(addr, &cw20::Cw20QueryMsg::TokenInfo {})?;
+                Ok(CheckedDepositInfo {
+                    token,
+                    deposit,
+                    refund_policy,
+                })
+            }
+            _ => Err(StdError::GenericErr {
+                msg: String::from("Unsupported asset"),
+            }),
```
ezekiiel, line 101: Will want to add a custom error here if we choose to merge this before doing the refactor.

ezekiiel (commented):
This is great. :) Cool to see an example of cw-asset being used - simpler code and we get to support both native and cw20 tokens.

Left comments as if we'd like to merge. I'm thinking the play here though is probably to leave this open for a bit and use it as a jumping off point to start on https://github.com/DA0-DA0/dao-contracts/discussions/462.

Appreciate you getting this moving and all the thoughts on deposits!

packages/voting/src/deposit.rs
JakeHartnell: Yeah, I was lazy.

packages/voting/src/deposit.rs
```diff
@@ -27,20 +28,28 @@ pub struct DepositInfo {
...
     /// deposits.
-    pub token: Addr,
+    pub token: AssetInfo,
     /// The number of tokens that must be deposited to create a
     /// proposal.
     pub deposit: Uint128,
     /// If failed proposals should have their deposits refunded.
-    pub refund_failed_proposals: bool,
+    pub refund_policy: DepositRefundPolicy,
 }
```
ezekiiel, line 53: This is a subtle one, but this change actually breaks v1 -> v2 migration logic. That logic expects that a v1 deposit info struct will deserialize into this type. By changing the layout, we break that.

Can probably hold off on updating the code until I've got a pre-propose module PR up and we decide if we'd like to merge this.

ezekiiel (changes requested):

packages/voting/src/deposit.rs
```diff
@@ -88,16 +113,14 @@ pub fn get_deposit_msg(
...
-                        owner: sender.to_string(),
-                        recipient: contract.to_string(),
-                        amount: info.deposit,
-                    })?,
-                };
-                let transfer_msg: CosmosMsg = transfer_msg.into();
+                let asset = AssetBase::new(info.token.clone(), info.deposit);
+                let transfer_msg = match asset.info {
+                    AssetInfoBase::Cw20(ref _addr) => asset.transfer_from_msg(sender, contract),
+                    AssetInfoBase::Native(ref _denom) => asset.transfer_msg(contract),
```
ezekiiel, line 119: Another tricky one. This won't actually require the deposit to be paid if, at the time of proposal creation, there are are `info.deposit` or more tokens in the proposal module.

Tests don't catch this because they're only testing the case where a single proposal being created. In that case the contract has zero native tokens so not paying the deposit will cause it to transfer `info.deposit` tokens to itself which it does not have. If it already had a deposit inside of it that would work regardless of if the proposer paid the deposit.

Great example of why breaking this logic out and having it fail open is so important..

JakeHartnell: Ah, really good catch.

JakeHartnell:
Closing in favor of #474

PR #467, opened 2022-08-26
Implement DAOstar standard: https://daostar.one/EIP

JakeHartnell:


elsehow (approved):
This is a pretty straightforward change and I am ready to approve (do see my one comment though). 

We may want to think (in the future, and when the spec is no longer a draft!) about how to read, validate, and display this standard nicely. It also occurs to me that we could auto-populate the JSON this standard calls for to a pretty great degree using the UI. We may want to add that to our roadmap once this standard becomes more mature.

contracts/cw-core/src/msg.rs
```diff
@@ -76,6 +76,8 @@ pub struct InstantiateMsg {
     /// items map. The value is an enum that either uses an existing
     /// address or instantiates a new contract.
     pub initial_items: Option<Vec<InitialItem>>,
+    /// The URI for the DAO linking to
```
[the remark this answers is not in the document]
JakeHartnell: Indeed it is.

JakeHartnell:
> This is a pretty straightforward change and I am ready to approve (do see my one comment though).
> 
> We may want to think (in the future, and when the spec is no longer a draft!) about how to read, validate, and display this standard nicely. It also occurs to me that we could auto-populate the JSON this standard calls for to a pretty great degree using the UI. We may want to add that to our roadmap once this standard becomes more mature.

Yes, the idea would be that we would auto-populate this in the UI.

contracts/cw-core/src/state.rs
```diff
@@ -20,6 +20,9 @@ pub struct Config {
     /// If true the contract will automatically add received cw721
     /// tokens to its treasury.
     pub automatically_add_cw721s: bool,
+    /// The URI for the DAO as defined by the DAOstar standard
+    /// https://daostar.one/EIP
+    pub dao_uri: Option<String>,
```
blue-note, line 25: If we are modifying existing state we will have to handle this in `migrate`

JakeHartnell: Oh, yes! Want to add a commit and make sure it's tested?

blue-note: sure will do

contracts/cw-core/src/contract.rs
```diff
@@ -850,6 +872,22 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
                     PROPOSAL_MODULES.save(deps.storage, address, proposal_module)?;
                     Ok(())
                 })?;
+
+            // Update config to have the V2 "dao_uri" field.
+            let config_item: Item<V1Config> = Item::new("config");
```
ezekiiel, line 877: @blue-note - in previous PRs we've made a new storage key `config_v2` and stored the new value there, keeping the old one. I think overriding is a fine approach, curious if there is a reason we did that here and not elsewhere?

ezekiiel (approved):
Looks great!

contracts/cw-core/src/contract.rs
blue-note: oh yeah good point -- I also don't think technically it is a problem to override, but might be better to create a new namespace just to be safe

PR #469, opened 2022-08-29
Fix flakey execute_stake_tokens integration test

de-husk:
We were sleeping 5 seconds here before I commented it out because it was flakey.

### DIFF
--- a/ci/integration_tests/src/tests/cw20_stake_test.rs
+++ b/ci/integration_tests/src/tests/cw20_stake_test.rs
@@ -1,6 +1,7 @@
 use crate::helpers::{chain::Chain, helper::create_dao};
 use cosmwasm_std::{to_binary, Uint128};
 use cw20_stake::{msg::StakedValueResponse, state::Config};
+use cw_core_interface::voting::VotingPowerAtHeightResponse;
 use test_context::test_context;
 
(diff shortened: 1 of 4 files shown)


ci/integration_tests/src/tests/cw20_stake_test.rs
```diff
@@ -95,20 +96,20 @@ fn execute_stake_tokens(chain: &mut Chain) {
...
-    //         &cw_core::msg::QueryMsg::VotingPowerAtHeight {
-    //             address: chain.user.addr.clone(),
-    //             height: None,
-    //         },
-    //     )
-    //     .unwrap();
-    // let power: VotingPowerAtHeightResponse = res.data().unwrap();
-    //
-    // assert_eq!(power.power, Uint128::new(100));
+    chain.orc.poll_for_n_blocks(1, 20_000).unwrap();
```
de-husk, line 99: This will block the current thread until we process `1` block, erroring if we wait more than 20 seconds.

ezekiiel: such a cool api!!

ezekiiel (approved):

JakeHartnell (approved):
LGTM

PR #472, opened 2022-08-30
Move module instantiation types to core interface packages.

ezekiiel:
This allows usage of these types by other modules which is required to
implement the pre proposal module design discussed here: <https://github.com/DA0-DA0/dao-contracts/discussions/462>

The cw721 and native token staking contracts are updated to use the
`Admin` type provided here instead of their `Owner` types.


ezekiiel:
ah sorry meant to close this. thanks for taking a look! turned out i could rebase this fine off your deposit work so just did that: https://github.com/DA0-DA0/dao-contracts/pull/473

PR #473, opened 2022-08-31
Implement new deposit module interface.

ezekiiel:
This implements the proposal deposit interface described in https://github.com/DA0-DA0/dao-contracts/discussions/462. It is based on @JakeHartnell's work on proposal deposits in https://github.com/DA0-DA0/dao-contracts/pull/465.

At the moment, this only implements the interface on `cw-proposal-single`. I'll add additional commits which implement a deposit module and test it as this comes along. Putting up so people can follow along if they'd like.


contracts/cw-pre-propose-base-proposal-single/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "cw-pre-propose-base-proposal-single"
```
JakeHartnell, line 2: Fine for a name, but when we reorg the folder structure it may be good to put these all in a `pre-propose` folder.

packages/voting/src/denom.rs
```diff
@@ -0,0 +1,226 @@
...
+/// should never be constructed literally and should always be built
+/// by calling `into_checked` on an `UncheckedDenom` instance.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum CheckedDenom {
+    /// A native (bank module) asset.
+    Native(String),
+    /// A cw20 asset.
+    Cw20(Addr),
+}
```
JakeHartnell, line 34: This is great. We may as well break it out into another package?

ci/integration_tests/Cargo.toml
```diff
@@ -15,6 +15,7 @@ cw20 = "0.13"
 cw20-base = "0.13"
 cw-utils = "0.13"
 cosmwasm-std = { version = "1.0.0", features = ["ibc3"] }
+cw-asset = "2.2.0"
```
JakeHartnell, line 18: Want to split up the fixing tests work?

contracts/cw-pre-propose-base-proposal-single/examples/schema.rs
```diff
@@ -0,0 +1,21 @@
...
+    let mut out_dir = current_dir().unwrap();
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema(&schema_for!(InstantiateMsg<Empty>), &out_dir);
+    export_schema(&schema_for!(ExecuteMsg<ProposeMessage, Empty>), &out_dir);
+    export_schema(&schema_for!(QueryMsg<Empty>), &out_dir);
+
+    // FIXME: best way to export response types? Should we wrap
```
JakeHartnell, line 19: We could use CosmWasm 1.1 which has this. There is a release candidate.

JakeHartnell: Can do that in a separate ticket though. There is already a lot here. :joy:

contracts/cw-pre-propose-base-proposal-single/src/contract.rs
```diff
@@ -0,0 +1,295 @@
...
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn execute(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, PreProposeError> {
+    type ExecuteInternal = ExecuteBase<ProposeMessageInternal, Empty>;
+    let internalized = match msg {
```
JakeHartnell, line 69: Can we add a comment here? This pattern seems to deviate a bit from what CW devs are use to.

contracts/cw-pre-propose-base/src/execute.rs
```diff
@@ -0,0 +1,223 @@
...
+            } else if proposal_completed {
+                // If the proposer doesn't get the deposit, the DAO does.
+                let dao = self.dao.load(deps.storage)?;
+                deposit_info.get_return_deposit_message(&dao)?
+            } else {
+                vec![]
+            }
+        } else {
+            vec![]
+        };
```
JakeHartnell, line 187: So, hear me out... what if we have two base contracts:
- `cw-pre-propose-base`: more general... no extra logic, used for custom pre-propose steps
- `cw-pre-propose-deposit-base`: uses `cw-pre-propose-base` for implementing deposit logic

There are other pre-propose flows that can be built like approvals or drafting.

JakeHartnell, line 187: So, hear me out... what if we have two base contracts:
- `cw-pre-propose-base`: more general... no extra logic, used for custom pre-propose steps
- `cw-pre-propose-deposit-base`: uses `cw-pre-propose-base` for implementing deposit logic

There are other pre-propose flows that can be built like approvals or drafting.

JakeHartnell, line 187: So, hear me out... what if we have two base contracts:
- `cw-pre-propose-base`: more general... no extra logic, used for custom pre-propose steps
- `cw-pre-propose-deposit-base`: uses `cw-pre-propose-base` for implementing deposit logic

There are other pre-propose flows that can be built like approvals or drafting.

contracts/cw-pre-propose-base/src/msg.rs
```diff
@@ -0,0 +1,55 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg<ProposalMessage, ExecuteExt> {
+    /// Creates a new proposal in the pre-propose module. MSG will be
+    /// serialized and used as the proposal creation message.
+    Propose { msg: ProposalMessage },
+
+    /// Extension message. Contracts that extend this one should put
+    /// their custom execute logic here. The default implementation
+    /// will do nothing if this variant is executed.
+    Ext { msg: ExecuteExt },
```
JakeHartnell, line 31: CosmWasm should develop some conventions for this pattern. In `cw-nfts` we always call it `Extension`.

contracts/cw-proposal-multiple/Cargo.toml
```diff
@@ -23,6 +23,7 @@ library = []
 
 [dependencies]
 cw-utils = {  version = "0.13" }
+cw-asset = "2.2.0"
```
JakeHartnell, line 26:
```suggestion
```

contracts/cw-proposal-single/Cargo.toml
```diff
@@ -21,16 +13,10 @@ backtraces = ["cosmwasm-std/backtraces"]
...
-optimize = """docker run --rm -v "$(pwd)":/code \
-  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
-  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-  cosmwasm/workspace-optimizer:0.12.6
-"""
-
 [dependencies]
 cosmwasm-std = { version = "1.0.0", features = ["ibc3"] }
 cosmwasm-storage = { version = "1.0.0" }
+cw-asset = "2.2.0"
```
JakeHartnell, line 19:
```suggestion
```

contracts/cw-pre-propose-base-proposal-single/Cargo.toml
ezekiiel: yep! planning on doing that. need to split do renaming fist because currently if we have a folder `contracts/pre-propose` cargo will read that as a rust project (not a folder of rust projects) as we specify `contracts/*` as our workspace. the renaming addresses this by moving every module into a subfolder with that module's implementations so that we can specify `contracts/proposal-modules/*, contracts/pre-propose/*` in our `Cargo.toml`.

packages/voting/src/denom.rs
ezekiiel: great idea! made `cw-denom`.

contracts/cw-pre-propose-base/src/execute.rs
ezekiiel: this is an interesting idea. I think you touch on a good point here that the current naming of this isn't great. IMO, we should rename what we have here to `cw-pre-propose-deposit-base`. I'm a little cautious about building a lower abstraction than that though yet as I feel that I'm not sure what it is.

For example: will all pre-propose modules need to interact with proposal status change hooks? What sort of instantiation information will they want? What sort of config options do they all share? I'm honestly pretty unsure. Maybe we wait on building that lower abstraction until we start writing more complex ones and can better identify shared logic?

contracts/cw-pre-propose-base/src/msg.rs
ezekiiel: Great point. Verbose is good. I'll update this to use `Extension`.

contracts/cw-pre-propose-base-proposal-single/src/contract.rs
ezekiiel: yes :) this is a little weird. spent a while coming up with this.

ezekiiel:
todo: remove assumption that proposal module is instantiator in pre-propose instantiation. Ether that, or remove `PreProposeInfo::AddrMayPropose {}` as an option during instantiation.

PR #477, opened 2022-09-02
Allow voting up to expiration time

### LINKED ISSUE #463: Allow voting on all proposals up to expiration time.
We should allow people to vote on proposals that have already passed, or already been rejected up to their expiration time. This allows better final tallies.

For example, currently with a majority threshold, if 99% of voters support it the final tally will only ever show that 50% voted yes. If we allow voting on passed proposals that 99% of people could still vote and have their position recorded on chain.

Credit to @maurolacy for thinking of this: https://github.com/CosmWasm/cw-plus/pull/784

NinoLipartiia:
Closes #463.


ezekiiel:
hey! thanks for putting this together! skimmed this and it looks good. we're really busy trying to get v2 released here so might be a couple days before i can get to a proper review. <3

contracts/cw-proposal-single/src/error.rs
```diff
@@ -31,9 +31,6 @@ pub enum ContractError {
     #[error("Proposal is ({size}) bytes, must be <= ({max}) bytes")]
     ProposalTooLarge { size: u64, max: u64 },
 
-    #[error("Proposal is not open ({id})")]
-    NotOpen { id: u64 },
-
     #[error("Proposal is expired ({id})")]
```
blue-note, line 34: nit: can we make the message #[error("Proposal ({id}) is expired")]

NinoLipartiia: ok, changed this error

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -244,8 +244,8 @@ pub fn execute_vote(
         return Err(ContractError::InvalidVote {});
     }
 
-    if prop.current_status(&env.block)? != Status::Open {
-        return Err(ContractError::NotOpen { id: proposal_id });
+    if prop.expiration.is_expired(&env.block) {
```
blue-note, line 247: ah I also realized that if the proposal is not expired, but it has either been `Closed` or `Executed`, we probably should not allow voting. I don't think there's anything technically wrong with that, but it seems unintuitive. We should probably also explicitly check for the states `Open`, `Passed`, or `Rejected` here.

blue-note:
apologies for the delay in reviewing here -- lets push this through! Also, for the integration test failure, rebasing on main should fix it.

blue-note (approved):

NinoLipartiia:
thanks for the review! I rebased over main, hope it fixes the failure

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -316,8 +316,8 @@ pub fn execute_vote(
     let mut prop = PROPOSALS
         .may_load(deps.storage, proposal_id)?
         .ok_or(ContractError::NoSuchProposal { id: proposal_id })?;
-    if prop.current_status(&env.block) != Status::Open {
-        return Err(ContractError::NotOpen { id: proposal_id });
+    if prop.expiration.is_expired(&env.block) {
+        return Err(ContractError::Expired { id: proposal_id });
```
ezekiiel, line 320: could we add a comment here explaining what is going on?

> Allow voting on proposals until they expire. Voting on a non-open proposal will never change their outcome as if an outcome has been determined, it is because no possible sequence of votes may cause a different one. This then serves to allow for better tallys of opinions in the event that a proposal passes or is rejected early.

ezekiiel (commented):
thanks for the PR! i think this would be a really nice change to do some fuzzing on. specifically, the assertion we're making here that:

> if a proposal's outcome has been determined, no sequence of additional votes will cause that outcome to change.

you could probably fuzz this pretty easily by manually selecting a number of interesting "passed" vote counts, generating a sequence of random votes, casting those votes, and then making sure the proposal's outcome has not changed.

if you'd like some example code to build this on top of, i did something pretty similar recently to fuzz a rate limiter [here](https://github.com/ezekiiel/cw721-proxy/blob/main/contracts/cw721-rate-limited-proxy/src/tests.rs#L297).

0xekez:
closing this for now. @NinoLipartiia if you have plans for picking this up feel free to just push to this branch and we'll reopen and do more review :) just want to clear the pr queue.

bekauz:
I'll pick this up

JakeHartnell (approved):
Great work! Thanks for helping this get across the finish line for v2.

contracts/proposal/cwd-proposal-multiple/src/testing/adversarial_tests.rs
```diff
@@ -229,3 +235,138 @@ fn test_execute_proposal_more_than_once() {
         .unwrap();
     assert!(matches!(err, ContractError::NotPassed {}));
 }
+
+// Users should be able to submit votes past the proposal
+// expiration date. Such votes do not affect the outcome
+// of the proposals; instead, they are meant to allow
+// voters to voice their opinion.
+#[test]
```
0xekez, line 243: this is a great test. love this docstring and the comments inside to walk through what it's doing

contracts/proposal/cwd-proposal-multiple/src/testing/tests.rs
```diff
@@ -3802,3 +3802,92 @@ fn test_no_double_refund_on_execute_fail_and_close() {
     let balance = query_balance_cw20(&app, token_contract.to_string(), CREATOR_ADDR);
     assert_eq!(balance, Uint128::new(1));
 }
+
+// Casting votes is only allowed within the proposal expiration timeframe
+#[test]
+pub fn test_not_allow_voting_on_expired_proposal() {
```
0xekez, line 3808: also a great test!

0xekez (approved):
amazing work. thank you @bekauz !

PR #480, opened 2022-09-11
Pre-propose multichoice

JakeHartnell:
Implements the new pre-propose flow for `cw-proposal-multiple`.

Cleaned up a few things. Figured it would be nice to have all of the contracts in our `testing` package. There may be a few more things we want in there. Went ahead and made testing modules public because some may find them useful.

PR #481, opened 2022-09-13
Implement new pre-propose interface and support native deposits in cw-proposal-single

ezekiiel:
This PR implements a new pre-propose interface on cw-proposal-single. The
pre-propose interface gives proposal modules the ability to delegate permission
to create proposals to a different contract. If one is present, only the
pre-propose module associated with a proposal module may create proposals.

In order to allow pre-propose modules to take actions like refunding proposal
deposits, pre-propose modules receive proposal hooks from their associated
proposal module. Proposal hooks fire when a proposal is created and when its
status changes. Like other receivers of proposal hooks, if the pre-propose
module fails to handle a hook message and errors it will be removed and the
proposal-module will fail-open by allowing any address to create a proposal
after the failure.

In summary, when a pre-propose module is associated with a proposal module:

1. Only that pre-propose module may create new proposals.
2. The pre-propose module will receive proposal hooks.
3. If the pre-propose module fails to handle a hook, it is removed.

## The pre-propose interface.

Pre-propose modules are only required to implement the following execute
message:

```rust
pub enum ExecuteMsg {
    /// Handles proposal hooks fired by the associated proposal
    /// module.
    ProposalHook(proposal_hooks::ProposalHookMsg),
}
```

The semantics of how a proposal is created and are left up to individual
modules.

## Deposits

The initial use case for these pre-propose modules is to allow for more complex
proposal deposit systems. To this end: this PR builds on [Jake's previous
work](https://github.com/DA0-DA0/dao-contracts/pull/465) and adds support for
native token proposal deposits, in addition to retaining support for cw20
deposits.

### cw-pre-propose-base

Proposal deposit logic is currently identical across the `cw-proposal-single`
and `cw-proposal-multiple` proposal modules. The `cw-pre-propose-base` package
introduced in this PR implements this shared logic in the form of a pre-propose
module that may be extended to support the proposal semantics of a particular
module. The design of this base contract is based on and shares many
similarities with
[cw721-base](https://github.com/CosmWasm/cw-nfts/tree/main/contracts/cw721-base).

The `cw-pre-propose-base-proposal-single` contract added in this PR extends this
contract to support `cw-proposal-single` proposal messages, and the
`cw-pre-propose-base-proposal-multiple` contract put together by Jake in [this
PR](https://github.com/DA0-DA0/dao-contracts/pull/480) does the same thing for
`cw-proposal-multiple`.

Here is a flow chart of the proposal creation process with `cw-pre-propose-base`:

In addition to the logic described in that flowchart, the module also supports
updating it's configuration, and the DAO withdrawing deposits via the following
executed messages:

```rust
pub enum ExecuteMsg {
    /// Updates the configuration of this module. This will completely
    /// override the existing configuration. This new configuration
    /// will only apply to proposals created after the config is
    /// updated. Only the DAO may execute this message.
    UpdateConfig {
        deposit_info: Option<UncheckedDepositInfo>,
        open_proposal_submission: bool,
    },
    /// Withdraws funds inside of this contract to the message
    /// sender. The contracts entire balance for the specifed DENOM is
    /// withdrawn to the message sender. Only the DAO may call this
    /// method.
    ///
    /// This is intended only as an escape hatch in the event of a
    /// critical bug in this contract or it's proposal
    /// module. Withdrawing funds will cause future attempts to return
    /// proposal deposits to fail their transactions as the contract
    /// will have insufficent balance to return them. In the case of
    /// `cw-proposal-single` this transaction failure will cause the
    /// module to remove the pre-propose module from its proposal hook
    /// receivers.
    ///
    /// More likely than not, this should NEVER BE CALLED unless a bug
    /// in this contract or the proposal module it is associated with
    /// has caused it to stop receiving proposal hook messages, or if
    /// a critical security vulnerability has been found that allows
    /// an attacker to drain proposal deposits.
    Withdraw {
        /// The denom to withdraw funds for. If no denom is specified,
        /// the denomination currently configured for proposal
        /// deposits will be used.
        ///
        /// You may want to specify a denomination here if you are
        /// withdrawing funds that were previously accepted for
        /// proposal deposits but are not longer used due to an
        /// `UpdateConfig` message being executed on the contract.
        denom: Option<UncheckedDenom>,
    },
}
```

Some important behaviors of this module to note:

1. If this module receives a new proposal hook for a proposal that it did not
   create, it will do nothing and not fail the transaction. This addresses the
   case where this module has replaced an existing module and there are
   outstanding proposals.
2. Proposal deposits are snapshotted meaning that updating the config of this
   module will not change deposit information for in-flight proposals.

## Proposal Module Changes

Proposal modules are largely not impacted by these changes. The existing
proposal hook system (though slightly modified) is used to feed updates to
pre-propose modules.

There are only three changes to the `cw-proposal-single` interface.

1. Deposit information in `InstantiateMsg` is replaced with `pre_propose_info`
   which allows instantiating a pre-propose module as part of the proposal
   module's instantiation.
2. A new `proposer: Option<String>` field is added to the `ExecuteMsg::Propose`
   variant. If a pre-propose module is associated with the proposal module, it
   is expected that this is `Some(address to attribute proposal creation)`. If
   no pre-propose module is associated, this field is expected to be `None`. In
   this case the proposal module will fill in the proposal's proposer with the
   sender of the message.
3. A new `ExecuteMsg::UpdatePreProposeInfo` message variant is added which
   allows the DAO to update its pre-propose module.

Notably, if the DAO swaps out a pre-propose module, the old module will stop
receiving proposal hooks. It is expected (and covered in tests) that, in this
case, the DAO will executed the `Withdraw` method on its pre-propose module and
reimburse proposers of in-flight proposals who would otherwise not receive their
deposits as a result of the pre-propose module holding them no longer receiving
proposal hooks.

If you've made it this far, thanks for reading and taking the time to review my
PR!


packages/cw-pre-propose-base/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "cw-pre-propose-base"
```
onewhiskeypls, line 2: **question**: README.md needed or no?

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -33,7 +34,7 @@ use crate::{
     state::{Ballot, BALLOTS, CONFIG, PROPOSALS, PROPOSAL_COUNT, PROPOSAL_HOOKS, VOTE_HOOKS},
 };
 
-pub(crate) const CONTRACT_NAME: &str = "crates.io:cw-govmod-single";
+pub(crate) const CONTRACT_NAME: &str = "crates.io:cw-proposal-single";
```
onewhiskeypls, line 37: **question**: are we removing usage of `govmod`? there is still usage of `govmod` in `packages/cw-core-macros`

packages/cw-core-interface/src/lib.rs
```diff
@@ -1,11 +1,123 @@
...
+    /// Causes the core module to execute all of MSGS in order. Only
+    /// callabale by a proposal module.1
     ExecuteProposalHook { msgs: Vec<CosmosMsg<Empty>> },
 }
+
+/// Information about the CosmWasm level admin of a contract. Used in
+/// conjunction with `ModuleInstantiateInfo` to instantiate modules.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum Admin {
```
onewhiskeypls, line 20: **question**: how come `None` was removed?

packages/cw-denom/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+            .into(),
+        })
+    }
+}
+
+/// Follows cosmos SDK validation logic. Specifically, the regex
+/// string `[a-zA-Z][a-zA-Z0-9/:._-]{2,127}`.
+///
+/// <https://github.com/cosmos/cosmos-sdk/blob/7728516abfab950dc7a9120caad4870f1f962df5/types/coin.go#L865-L867>
+pub fn validate_native_denom(denom: String) -> Result<CheckedDenom, DenomError> {
```
onewhiskeypls, line 145: **praise**: neat! wen `TokenFactory`???

onewhiskeypls (commented):
**utACK**: Gave it a quick pass with the smoothest brain possible...will need to revisit a few things tomorrow, especially the test coverage. Feel free to ignore...everything

contracts/cw721-stake/src/contract.rs
```diff
@@ -33,8 +34,8 @@ pub fn instantiate(
         .owner
         .as_ref()
         .map(|owner| match owner {
-            Owner::Addr(address) => deps.api.addr_validate(address),
-            Owner::Instantiator {} => Ok(info.sender),
+            Admin::Address { addr } => deps.api.addr_validate(addr),
+            Admin::CoreModule {} => Ok(info.sender),
```
JakeHartnell, line 38: Soooo much nicer than `Instantiator`

packages/cw-denom/src/lib.rs
JakeHartnell: Soon...

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -155,17 +179,6 @@ pub fn execute_propose(
...
-    // Check that the sender is a member of the governance contract.
-    let sender_power = get_voting_power(
-        deps.as_ref(),
-        sender.clone(),
-        config.dao.clone(),
-        Some(env.block.height),
-    )?;
-    if sender_power.is_zero() {
-        return Err(ContractError::Unauthorized {});
-    }
```
JakeHartnell, line 167: I think we need this check? Many DAOs only want members to be able to submit proposals.

Like I don't want people submitting props on my multisigs....

JakeHartnell: Maybe make a comment of where this check happens?

https://github.com/DA0-DA0/dao-contracts/blob/zeke/deposit-modules/packages/cw-pre-propose-base/src/execute.rs#L132

packages/cw-pre-propose-base/src/execute.rs
```diff
@@ -0,0 +1,320 @@
...
+                }
+            };
+            match denom {
+                None => Err(PreProposeError::NoWithdrawalDenom {}),
+                Some(denom) => {
+                    let balance = denom.query_balance(&deps.querier, &env.contract.address)?;
+                    if balance.is_zero() {
+                        Err(PreProposeError::NothingToWithdraw {})
+                    } else {
+                        let withdraw_message = denom.get_transfer_to_message(&dao, balance)?;
```
blue-note, line 221: do you foresee the need to ever withdraw less than the full balance? I know the withdraw is meant as a stopgap measure, but just wondering

contracts/cw-pre-propose-base-proposal-single/src/contract.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use cw_pre_propose_base::{
+    error::PreProposeError,
+    msg::{ExecuteMsg as ExecuteBase, InstantiateMsg as InstantiateBase, QueryMsg as QueryBase},
+    state::PreProposeContract,
+};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+pub(crate) const CONTRACT_NAME: &str = "crates.io:cw-pre-propose-base-proposal-single";
```
blue-note, line 16: this contract name is a bit of a mouthful; could I suggest `cw-pre-propose-single`?

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -846,15 +850,54 @@ pub fn reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractEr
...
         }
         TaggedReplyId::FailedProposalHook(idx) => {
-            let addr = PROPOSAL_HOOKS.remove_hook_by_index(deps.storage, idx)?;
-            Ok(Response::new().add_attribute("removed proposal hook", format!("{addr}:{idx}")))
+            // When firing off proposal hooks indexes `[0,
+            // hook_count)` are used to represent hooks that have been
+            // fired via the regular PROPOSAL_HOOKS method. Index
+            // `hook_count` is used to represent the pre-propose
+            // module.
+            let count = PROPOSAL_HOOKS.hook_count(deps.storage)?;
```
blue-note, line 862: can you explain why it is the case that `hook_count` will correspond to the pre-propose module?

packages/cw-pre-propose-base/src/execute.rs
```diff
@@ -0,0 +1,320 @@
...
+            ExecuteMsg::Withdraw { denom } => {
+                self.execute_withdraw(deps.as_ref(), env, info, denom)
+            }
+
+            ExecuteMsg::Extension { .. } => Ok(Response::default()),
+
+            ExecuteMsg::ProposalHook(ProposalHookMsg::ProposalStatusChanged {
+                id,
+                new_status,
+                // Deposits only care about where you're headed, not
```
blue-note, line 97: 😂

packages/cw-pre-propose-base/src/execute.rs
```diff
@@ -0,0 +1,320 @@
...
+        if info.sender != proposal_module {
+            return Err(PreProposeError::NotModule {});
+        }
+
+        match self.deposits.may_load(deps.storage, id)? {
+            Some((deposit_info, proposer)) => {
+                let messages = if let Some(ref deposit_info) = deposit_info {
+                    // If the proposal is completed, either return to the DAO
+                    // or issue a refund.
+                    let proposal_completed = new_status == "closed" || new_status == "executed";
```
blue-note, line 250: reason to not use the enum?

packages/cw-pre-propose-base/src/state.rs
```diff
@@ -0,0 +1,71 @@
...
+pub struct PreProposeContract<InstantiateExt, ExecuteExt, QueryExt, ProposalMessage> {
+    /// The proposal module that this module is associated with.
+    pub proposal_module: Item<'static, Addr>,
+    /// The DAO (cw-dao-core module) that this module is associated
+    /// with.
+    pub dao: Item<'static, Addr>,
+    /// The configuration for this module.
+    pub config: Item<'static, Config>,
+    /// Map between proposal IDs and (deposit, proposer) pairs.
+    pub deposits: Map<'static, u64, (Option<CheckedDepositInfo>, Addr)>,
```
blue-note, line 29: how will it have to change when allowing multiple people to contribute different amounts to deposits?

packages/voting/src/deposit.rs
```diff
@@ -19,107 +44,349 @@ pub enum DepositToken {
...
+            denom: CheckedDenom::Native(denom),
+            ..
+        } = self
+        {
+            // must_pay > may_pay. The method this is getting called
+            // in is accepting a deposit. It seems likely to me that
+            // if other payments are here it's a bug in a frontend and
+            // not an intentional thing.
+            let paid = must_pay(info, denom)?;
+            if paid != *amount {
```
blue-note, line 142: what if `paid > amount`? still want to error?

packages/voting/src/deposit.rs
```diff
@@ -19,107 +44,349 @@ pub enum DepositToken {
...
+            // posterity.
+            if amount.is_zero() {
+                vec![]
             } else {
-                let transfer_msg = WasmMsg::Execute {
-                    contract_addr: info.token.to_string(),
+                vec![WasmMsg::Execute {
+                    contract_addr: address.to_string(),
                     funds: vec![],
                     msg: to_binary(&cw20::Cw20ExecuteMsg::TransferFrom {
```
blue-note, line 175: does the depositor create an allowance for this? where does that happen / is it triggered by the frontend?

ci/integration_tests/src/helpers/helper.rs
```diff
@@ -1,14 +1,14 @@
 use super::chain::Chain;
 use anyhow::Result;
-use cosmwasm_std::{to_binary, Decimal, Uint128};
+use cosmwasm_std::{to_binary, Decimal, Empty, Uint128};
 use cw20::Cw20Coin;
-use cw_core::{
-    msg::{Admin, ModuleInstantiateInfo},
-    query::DumpStateResponse,
-};
+use cw_core::query::DumpStateResponse;
```
blue-note, line 5: should the pre-propose module be included in cw-core's DumpState?

contracts/cw-proposal-multiple/src/contract.rs
```diff
@@ -154,10 +157,15 @@ pub fn execute_propose(
         return Err(ContractError::InactiveDao {});
     }
 
-    // Check that the sender is a member of the governance contract.
-    let sender_power = get_voting_power(deps.as_ref(), sender.clone(), config.dao.clone(), None)?;
-    if sender_power.is_zero() {
-        return Err(ContractError::MustHaveVotingPower {});
+    // Check whether open_propsoal_submissions are disabled
+    if !config.open_proposal_submission {
```
blue-note, line 161: how do we guarantee that the deposit has already been paid if we are here?

blue-note: maybe i'm misunderstanding but it seems like the flow is frontend --> execute_propose on prepropose module --> execute_propose on propose module --> new proposal hook on the prepropose module

but what if someone bypasses the frontend and executes execute_propose on the proposal_module? also why are the steps in execute_new_proposal_hook in the prepropose module not done during the deposit collection step?

packages/cw-pre-propose-base/src/execute.rs
```diff
@@ -0,0 +1,320 @@
...
+        // The proposal module instantiates us. Same as below, we're
+        // making limited assumptions here. The only way to associate
+        // a deposit module with a proposal module is for the proposal
+        // module to instantiate it.
+        self.proposal_module.save(deps.storage, &info.sender)?;
+
+        // The DAO instantiates the proposal module. Note that we're
+        // making limited assumptions here. The only way to add a
+        // proposal module to a DAO is for it to create it.
+        let module_info: ContractInfoResponse =
```
blue-note, line 45: we talked about including a DAO query on the proposal module

contracts/cw-proposal-single/src/contract.rs
blue-note: we talked about separating out prepropose from the proposal hooks

packages/voting/src/pre_propose.rs
```diff
@@ -0,0 +1,156 @@
...
+        match self {
+            Self::Anyone {} => true,
+            Self::Module { addr } => creator == addr,
+        }
+    }
+
+    /// Determines if ADDR is the module registered to create
+    /// proposals by this proposal creation policy. Returns true if so
+    /// and false otherwise.
+    pub fn addr_is_my_module(&self, addr: &Addr) -> bool {
```
blue-note, line 43: this function doesn't seem to be used anywhere?

packages/voting/src/pre_propose.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl PreProposeInfo {
+    pub fn into_initial_policy_and_messages(
+        self,
+        dao_address: Addr,
+    ) -> StdResult<(ProposalCreationPolicy, Vec<SubMsg<Empty>>)> {
+        Ok(match self {
+            Self::AnyoneMayPropose {} => (ProposalCreationPolicy::Anyone {}, vec![]),
+            Self::ModuleMayPropose { info } => (
+                // Anyone can propose, then, upon instantiation
```
blue-note, line 59: nit: could you add slightly more explanation about how this allows failing open

contracts/cw-proposal-multiple/src/contract.rs
blue-note: nvm, ignore this that is the permitted check, also this is the mc contract which is why I was confused

packages/cw-pre-propose-base/src/execute.rs
```diff
@@ -0,0 +1,320 @@
...
+                                && deposit_info.refund_policy != DepositRefundPolicy::Never);
+
+                        if should_refund_to_proposer {
+                            deposit_info.get_return_deposit_message(&proposer)?
+                        } else {
+                            // If the proposer doesn't get the deposit, the DAO does.
+                            let dao = self.dao.load(deps.storage)?;
+                            deposit_info.get_return_deposit_message(&dao)?
+                        }
+                    } else {
```
blue-note, line 267: yeah upon further reflection about the hooks intention, since we only care about initial and final status, I think it would be nicer to stop using the generic proposal hooks and only call this hook explicitly on `execute_execute` and `execute_close`. as we discussed earlier using the proposal hooks creates some unnecessary custom logic

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -846,15 +850,54 @@ pub fn reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractEr
...
+                        // reply and we don't have a pre-propose
+                        // module installed. This should be
+                        // unreachable.
+                        return Err(ContractError::InvalidHookIndex { idx });
+                    }
+                    ProposalCreationPolicy::Module { addr } => {
+                        // If we are here, our pre-propose module has
+                        // errored while receiving a proposal
+                        // hook. Rest in peace pre-propose module.
+                        CREATION_POLICY.save(deps.storage, &ProposalCreationPolicy::Anyone {})?;
```
blue-note, line 876: would we be able to send deposit funds to the dao upon failure and removal of this module?

blue-note: I like still having withdraw as a stopgap, but also was thinking about avoiding steps the dao has to take to handle a software failure

contracts/cw-proposal-single/src/v1_state.rs
```diff
@@ -0,0 +1,46 @@
...
+};
+
+pub fn v1_percentage_threshold_to_v2(v1: voting_v1::PercentageThreshold) -> PercentageThreshold {
+    match v1 {
+        voting_v1::PercentageThreshold::Majority {} => PercentageThreshold::Majority {},
+        voting_v1::PercentageThreshold::Percent(p) => PercentageThreshold::Percent(p),
+    }
+}
+
+pub fn v1_threshold_to_v2(v1: voting_v1::Threshold) -> Threshold {
```
blue-note, line 17: can you explain again why we need this?

packages/cw-denom/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+repository = "https://github.com/DA0-DA0/dao-contracts"
+description = "A package for validation and handling of cw20 and native Cosmos SDK denominations."
+
+
+[dependencies]
+cosmwasm-std = "1.0"
+schemars = "0.8"
+serde = { version = "1.0", default-features = false, features = ["derive"] }
+thiserror = { version = "1.0" }
+cw20 = "0.13"
```
robert-zaremba, line 15: why not using the latest (v.015.1)?

packages/cw-denom/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+/// asset.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum UncheckedDenom {
+    /// A native (bank module) asset.
+    Native(String),
+    /// A cw20 asset.
+    Cw20(String),
+}
+
```
robert-zaremba, line 68: I've added `Asset` type to combine denom with amount:
https://github.com/42-Hackwasm/yield-aggregator/blob/main/packages/cw-denom/src/lib.rs#L59

Maybe we can copy it over here or do in other PR?

packages/cw-denom/Cargo.toml
ezekiiel: we're not yet on cosmwasm 1.1. will update soon as a followup to this pr!

packages/cw-pre-propose-base/src/execute.rs
ezekiiel: this is such a good idea. perhaps all this module needs is two hooks `ProposalCreatedHook { id: u64 }` and `ProposalCompletedHook { id: u64, passed: bool }`?

packages/cw-denom/src/lib.rs
ezekiiel: i actually intentionally did not include the asset part here. i feel that cw20 and native tokens have reasonably distinct interfaces that are difficult to reconcile. for example, how would you expose something like `transfer_from` on a native asset?

larry0x has a good approach with [`cw-asset`](https://github.com/mars-protocol/cw-asset), but IMO i don't want to use it in the contracts because it is possible to call a method and have it panic if the underlying asset type isn't correct. avoiding panics where possible is really important to me.

would love to hear more thoughts here. :) this is not to say that we should not make an asset library, just to say that i'm not sure how it is best done.

contracts/cw-proposal-single/src/v1_state.rs
ezekiiel: this is used for migrations. rust doesn't guarantee that two types have the same in-memory layout despite having the same fields so we need to manually convert between the two types. if we could guarantee they had the same in-memory layout we could use `std::mem::transmute`. i strongly suspect this would work, just need some way to be sure (perhaps a test?).

contracts/cw-proposal-single/src/contract.rs
```diff
@@ -134,8 +138,28 @@ pub fn execute_propose(
...
+    // Check that the sender is permitted to create proposals.
+    if !proposal_creation_policy.is_permitted(&sender) {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // Determine the appropriate proposer. If this is coming from our
+    // pre-propose module, it must be specified. Otherwise, the
+    // proposer should not be specified.
+    let proposer = match (proposer, proposal_creation_policy) {
+        (None, ProposalCreationPolicy::Anyone {}) => sender.clone(),
```
blue-note, line 155: so say the prepropose module fails open and we default to ``ProposalCreationPolicy::Anyone`, allowing anyone to potentially create spam proposals. Wouldn't we instead want to default to allowing only members to submit proposals as before?

packages/cw-denom/src/lib.rs
robert-zaremba: Asset is just a base object. `cw20` (with `transfer_from`, `send` etc...) is a specialization.

But if it's not needed here, then fair enough to stick with `cw-asset`.

packages/voting/src/deposit.rs
blue-note: yes, triggered by frontend

ci/integration_tests/src/helpers/helper.rs
blue-note: decided it would be better to keep as is and allow the frontend to query the proposal module's proposal_creation_policy instead.. less complicated and handles case where creation policy is Anyone

contracts/cw-proposal-single/src/contract.rs
blue-note: after discussion, we will leave as is -- this is a failure stopgap and if the module ends up in that state we have bigger issues

ezekiiel:
dang. code coverage isn't telling us much here. [this file](https://app.codecov.io/gh/DA0-DA0/dao-contracts/blob/zeke%2Fdeposit-modules/packages/cw-pre-propose-base/src/execute.rs), for example is described as being 83% covered, despite having 100% coverage by visual inspection.

packages/proposal-hooks/src/lib.rs
```diff
@@ -23,15 +24,26 @@ pub enum ProposalHookMsg {
...
-        ProposalHookMsg::NewProposal { id },
+        ProposalHookMsg::NewProposal {
+            id,
+            proposer: proposer.into_string(),
+        },
     ))?;
+
     let mut index: u64 = 0;
-    hooks.prepare_hooks(storage, |a| {
+    let messages = hooks.prepare_hooks(storage, |a| {
```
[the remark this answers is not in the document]
ezekiiel: i am 90% sure that `let mut v = v` will be a no-op at compilation time. a little hard to demonstrate with a compiler explorer as there is a lot of boilerplate for working with vectors, though this is a [reasonably common](https://www.thorsten-hans.com/shadowing-temporary-mutability-rust/) pattern in rust. :)

contracts/cw-proposal-single/src/contract.rs
ezekiiel: great catch! fixed.

packages/cw-core-interface/src/lib.rs
ezekiiel: good q! i feel it's a little more idomatic rust to use an option instead of having our own none variant. this is now specified as `Some(Admin::CoreModule {})` instead of `Admin::CoreModule {}` and the no admin case is now just `None`.

packages/cw-pre-propose-base/src/state.rs
ezekiiel: we'll need a different data structure for that. :) can likely borrow what ion dao [has done](https://github.com/many-things/ion-dao-contracts/blob/dev/contracts/dao/src/state.rs#L80-L82).

packages/cw-pre-propose-base/Cargo.toml
ezekiiel: done! great point.

ezekiiel:
gas cost increase in instantiation is due to the addition of a new module which adds another `WasmMsg::Instantiate` to the creation process. i think this is ok.

JakeHartnell (approved):
:shipit:

bmorphism:
My rather colloquial review on the bird app: https://twitter.com/bmorphism/status/1578923792071815168

really excited for V2! ☺️

Callum-A (approved):

Issue #431, opened 2022-08-02
Add a withdraw method to `cw20-stake-external-rewards`


0xekez:
Currently it is possible for the contract to get into a weird state if a cw20 `Transfer` method (as opposed to a `Send`) is used to fund the contract. This happens because the funds present will be greater than the funds expected and thus this contract will distribute rewards longer than the rewards duration.

If someone gets into this state having a `Withdraw {}` method which the owner can call which withdraws all of the contract's funds will help mitigate the harm caused by something like this.

This method should likely query it's balance (see: `reward_token` field on `Config`) and then return that many tokens to the owner. It should only be callable by the owner.

ezekiiel:
Would also be nice to consider adding methods that let the owner / manager enable or disable the contract. Methods like that are good for stopping the world to investigate things.

PR #486, opened 2022-09-30
Cw20 stake withdraw

bekauz:
Closes https://github.com/DA0-DA0/dao-contracts/issues/431

Also includes the option to pause the contract (https://github.com/DA0-DA0/dao-contracts/issues/431#issuecomment-1202997839)

### DIFF
--- a/contracts/cw20-stake-external-rewards/src/contract.rs
+++ b/contracts/cw20-stake-external-rewards/src/contract.rs
@@ -9,17 +9,16 @@ use crate::state::{
 use crate::ContractError;
 use crate::ContractError::{
     InvalidCw20, InvalidFunds, NoRewardsClaimable, RewardPeriodFinished, RewardPeriodNotFinished,
-    Unauthorized,
+    Unauthorized, UnauthorizedWithdrawal,
 };
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 
 use cosmwasm_std::{
-    from_binary, to_binary, to_vec, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut, Empty,
-    Env, MessageInfo, Response, StdError, StdResult, Uint128, Uint256, WasmMsg,
+    from_binary, to_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut, Empty, Env,
+    MessageInfo, Response, StdError, StdResult, Uint128, Uint256, WasmMsg,
 };
 use cw2::set_contract_version;
-use cw20::Denom::Cw20;
 use cw20::{Cw20ReceiveMsg, Denom};
 use cw20_stake::hooks::StakeChangedHookMsg;
 use std::cmp::min;
@@ -45,7 +44,7 @@ pub fn instantiate(
 
     let reward_token = match msg.reward_token {
         Denom::Native(denom) => Denom::Native(denom),
-        Cw20(addr) => Cw20(deps.api.addr_validate(addr.as_ref())?),
+        Denom::Cw20(addr) => Denom::Cw20(deps.api.addr_validate(addr.as_ref())?),
     };
 
     if msg.reward_duration == 0 {
@@ -95,7 +94,7 @@ pub fn instantiate(
             "reward_token",
             match config.reward_token {
                 Denom::Native(denom) => denom,
-                Cw20(addr) => addr.into_string(),
+                Denom::Cw20(addr) => addr.into_string(),
             },
         )
         .add_attribute("reward_rate", reward_config.reward_rate)
@@ -109,23 +108,21 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
     match msg {
         MigrateMsg::FromV1 { is_paused } => {
-            // Load the current config and introduce the claim pause flag
+            // load the current config
             let current_config =
                 stake_cw20_external_rewards_v1::state::CONFIG.load(deps.storage)?;
-            let is_paused = if let Some(val) = is_paused {
-                val
-            } else {
-                false
-            };
+            // save it with the pause flag under new storage key
+            CONFIG.save(
+                deps.storage,
+                &Config {
+                    owner: current_config.owner,
+                    manager: current_config.manager,
+                    staking_contract: current_config.staking_contract,
+                    reward_token: current_config.reward_token,
+                    is_paused,
+                },
+            )?;
 
-            let new_config = Config {
-                owner: current_config.owner,
-                manager: current_config.manager,
-                staking_contract: current_config.staking_contract,
-                reward_token: current_config.reward_token,
-                is_paused,
-            };
-            deps.storage.set(b"config", &to_vec(&new_config)?);
             Ok(Response::default()
                 .add_attribute("action", "migrate")
                 .add_attribute("from", "v1")
@@ -157,7 +154,7 @@ pub fn execute(
         ExecuteMsg::UpdateManager { new_manager } => {
             execute_update_manager(deps, env, info, new_manager)
         }
-        ExecuteMsg::UpdatePauseStatus { is_paused } => {
+        ExecuteMsg::UpdatePauseStatus { pause: is_paused } => {
             execute_update_pause_status(deps, env, info, is_paused)
         }
     }
@@ -192,7 +189,7 @@ pub fn execute_fund_native(
             let amount = cw_utils::must_pay(&info, &denom).map_err(|_| InvalidFunds {})?;
             execute_fund(deps, env, info.sender, amount)
         }
-        Cw20(_) => Err(InvalidFunds {}),
+        Denom::Cw20(_) => Err(InvalidFunds {}),
     }
 }
 
@@ -204,7 +201,7 @@ pub fn execute_withdraw(
     // only the owner can withdraw
     let config = CONFIG.load(deps.storage)?;
     if config.owner != Some(info.sender.clone()) {
-        return Err(Unauthorized {});
+        return Err(UnauthorizedWithdrawal {});
     };
 
     // withdrawal is only possible after the reward period
@@ -219,13 +216,13 @@ pub fn execute_withdraw(
                 .unwrap()
                 .amount
         }
-        Cw20(denom) => {
+        Denom::Cw20(denom) => {
             let result: cw20::BalanceResponse = deps
                 .querier
                 .query_wasm_smart(
                     denom.to_string(),
                     &cw20::Cw20QueryMsg::Balance {
-                        address: String::from(env.contract.address),
+                        address: env.contract.address.into_string(),
                     },
                 )
                 .unwrap();
@@ -548,7 +545,7 @@ pub fn execute_update_pause_status(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    is_paused: bool,
+    pause: bool,
 ) -> Result<Response<Empty>, ContractError> {
     let mut config = CONFIG.load(deps.storage)?;
     if Some(info.sender.clone()) != config.owner && Some(info.sender) != config.manager {
@@ -558,11 +555,11 @@ pub fn execute_update_pause_status(
     let reward_config = REWARD_CONFIG.load(deps.storage)?;
     if reward_config.period_finish < env.block.height {
         return Err(RewardPeriodFinished {});
-    } else if config.is_paused == is_paused {
+    } else if config.is_paused == pause {
         return Err(ContractError::SettingCurrentPauseState {});
     }
 
-    config.is_paused = is_paused;
+    config.is_paused = pause;
     CONFIG.save(deps.storage, &config)?;
 
     Ok(Response::new()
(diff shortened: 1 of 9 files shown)


contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -99,14 +100,41 @@ pub fn instantiate(
...
-    // Set contract to version to latest
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
-    Ok(Response::default())
+    match msg {
+        MigrateMsg::FromV1 { is_paused } => {
+            // Load the current config and introduce the claim pause flag
+            let current_config =
+                stake_cw20_external_rewards_v1::state::CONFIG.load(deps.storage)?;
+            let is_paused = if let Some(val) = is_paused {
```
blue-note, line 115: maybe we should just make `is_paused` be a `bool` instead of `Option<bool>`? not sure what we get by making `false` indicated by both `None` and `false`

contracts/cw20-stake-external-rewards/src/state.rs
```diff
@@ -12,6 +12,7 @@ pub struct Config {
     pub manager: Option<Addr>,
     pub staking_contract: Addr,
     pub reward_token: Denom,
+    pub is_paused: bool,
 }
 pub const CONFIG: Item<Config> = Item::new("config");
```
blue-note, line 17: let's change this config key to `config_v2`!

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -99,14 +100,41 @@ pub fn instantiate(
...
+            };
+
+            let new_config = Config {
+                owner: current_config.owner,
+                manager: current_config.manager,
+                staking_contract: current_config.staking_contract,
+                reward_token: current_config.reward_token,
+                is_paused,
+            };
+            deps.storage.set(b"config", &to_vec(&new_config)?);
```
blue-note, line 128: in general we want to try to avoid overriding previous state just to be safe. we can instead create a `config_v2` state key and write to that. take look at the migrate method in `cw-proposal-single` to see an example of that :)

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -164,6 +196,51 @@ pub fn execute_fund_native(
     }
 }
 
+pub fn execute_withdraw(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+) -> Result<Response<Empty>, ContractError> {
+    // only the owner can withdraw
+    let config = CONFIG.load(deps.storage)?;
```
blue-note, line 205: should the manager be able to withdraw as well?

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -164,6 +196,51 @@ pub fn execute_fund_native(
...
+    if config.owner != Some(info.sender.clone()) {
+        return Err(Unauthorized {});
```
blue-note, line 207: can we make a more specific unauthorized error message here?

de-husk:
The integration tests failed because `cw-optimizoor` got a silent error while optimizing the contracts. 

I have a PR to fix this so that errors will bubble up in the future: https://github.com/mandrean/cw-optimizoor/pull/22

But for the time being I just checked out the repo and printed the error:
```
Error: There are multiple `stake-cw20-external-rewards` packages in your project, and the specification `stake-cw20-external-rewards` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  file:///Users/harry/cosmos/test-dao-contracts/contracts/cw20-stake-external-rewards#stake-cw20-external-rewards@0.2.6
  https://github.com/DA0-DA0/dao-contracts.git#stake-cw20-external-rewards@0.2.6
```

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -463,6 +544,32 @@ pub fn execute_update_manager(
...
+    let mut config = CONFIG.load(deps.storage)?;
+    if Some(info.sender.clone()) != config.owner && Some(info.sender) != config.manager {
+        return Err(ContractError::Unauthorized {});
+    };
+
+    let reward_config = REWARD_CONFIG.load(deps.storage)?;
+    if reward_config.period_finish < env.block.height {
+        return Err(RewardPeriodFinished {});
+    } else if config.is_paused == is_paused {
+        return Err(ContractError::SettingCurrentPauseState {});
```
blue-note, line 562: hmm, I would argue that if they are trying to set `is_paused` to what it already is, it should be a no-op (ie no error)

contracts/cw20-stake-external-rewards/src/msg.rs
```diff
@@ -26,9 +30,11 @@ pub enum ExecuteMsg {
     Claim {},
     Receive(Cw20ReceiveMsg),
     Fund {},
+    Withdraw {},
     UpdateRewardDuration { new_duration: u64 },
     UpdateOwner { new_owner: Option<String> },
     UpdateManager { new_manager: Option<String> },
+    UpdatePauseStatus { is_paused: bool },
```
blue-note, line 37: nit: change `is_paused` to `pause` :)

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -463,6 +544,32 @@ pub fn execute_update_manager(
         ))
 }
 
+pub fn execute_update_pause_status(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    is_paused: bool,
```
blue-note, line 551: naming nit: can we name this var `pause` instead of `is_paused` since its requesting an action? (

contracts/cw20-stake-external-rewards/src/contract.rs
bekauz: I was thinking about this one a bunch and had the same thought. Will do!

contracts/cw20-stake-external-rewards/src/contract.rs
bekauz: I kind of followed the pattern I saw on prop-single:
https://github.com/DA0-DA0/dao-contracts/blob/693a5d6a36ada1505c5a52e8029b1b80ea557393/contracts/cw-proposal-single/src/contract.rs#L339-L343

No strong preference though, what do others think?

contracts/cw20-stake-external-rewards/src/contract.rs
bekauz: I kind of followed the pattern I saw on prop-single:
https://github.com/DA0-DA0/dao-contracts/blob/693a5d6a36ada1505c5a52e8029b1b80ea557393/contracts/cw-proposal-single/src/contract.rs#L339-L343

No strong preference though, what do others think?

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -18,10 +19,9 @@ use cosmwasm_std::{
     MessageInfo, Response, StdError, StdResult, Uint128, Uint256, WasmMsg,
 };
 use cw2::set_contract_version;
+use cw20::Denom::Cw20;
```
ezekiiel, line 22: probably just leftover from the editor, but no need to import a specific variant of an enum
```suggestion
```

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -164,6 +194,51 @@ pub fn execute_fund_native(
...
+        return Err(RewardPeriodNotFinished {});
+    };
+    let leftover_stake = match &config.reward_token {
+        Denom::Native(denom) => {
+            deps.querier
+                .query_balance(env.contract.address, denom.to_string())
+                .unwrap()
+                .amount
+        }
+        Cw20(denom) => {
```
ezekiiel, line 220:
```suggestion
        Denom::Cw20(denom) => {
```

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -164,6 +194,51 @@ pub fn execute_fund_native(
...
+            let result: cw20::BalanceResponse = deps
+                .querier
+                .query_wasm_smart(
+                    denom.to_string(),
+                    &cw20::Cw20QueryMsg::Balance {
+                        address: String::from(env.contract.address),
```
ezekiiel, line 226:
```suggestion
                        address: env.contract.address.into_string(),
```

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -463,6 +542,32 @@ pub fn execute_update_manager(
...
+pub fn execute_update_pause_status(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    pause: bool,
+) -> Result<Response<Empty>, ContractError> {
+    let mut config = CONFIG.load(deps.storage)?;
+    if Some(info.sender.clone()) != config.owner && Some(info.sender) != config.manager {
+        return Err(ContractError::Unauthorized {});
+    };
```
ezekiiel, line 554: i really like how you've let both the manager and owner pause, but only the owner withdraw. seems like a good system where the manager can stop the world, but the owner can unstop it and withdraw.

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -556,6 +662,26 @@ mod tests {
...
+        )
+        .with_migrate(crate::contract::migrate);
+        Box::new(contract)
+    }
+
+    pub fn v1_external_rewards_contract() -> Box<dyn Contract<Empty>> {
+        let contract = ContractWrapper::new(
+            stake_cw20_external_rewards_v1::contract::execute,
+            stake_cw20_external_rewards_v1::contract::instantiate,
+            stake_cw20_external_rewards_v1::contract::query,
```
ezekiiel, line 679: great work importing and using the v1 version!

ezekiiel (commented):
this looks great! awesome work! will do another pass tomorrow and likely stamp.

i really like your design where the pause only stops claiming, but allows config updates and funding. feels like this provides maximum opportunity for the manager and owner of this contract to get it back in a good state, while pausing to stop a bad state from having consequences.

contracts/cw20-stake-external-rewards/src/contract.rs
ezekiiel: would be interested in @ben2x4's thoughts here. i'd imagine we want to only allow the owner to do this.

ezekiiel (approved):
lg2m! i like all the tests. can tell this was well thought about.

blue-note (approved):

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -160,10 +189,55 @@ pub fn execute_fund_native(
...
+    if config.owner != Some(info.sender.clone()) {
+        return Err(UnauthorizedWithdrawal {});
+    };
+
+    // withdrawal is only possible after the reward period
+    let reward_config = REWARD_CONFIG.load(deps.storage)?;
+    if reward_config.period_finish > env.block.height {
+        return Err(RewardPeriodNotFinished {});
+    };
+    let leftover_stake = match &config.reward_token {
```
ben2x4, line 212: This will break the contract, the reward period may be over but all the rewards might not be claimed, going forward this reward contract will have bugs and a new contract will mostly likely need to be created to distribute future rewards.

contracts/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -243,6 +317,11 @@ pub fn execute_claim(
     env: Env,
     info: MessageInfo,
 ) -> Result<Response<Empty>, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if config.is_paused {
```
ben2x4, line 321: What is the intention of this pause state? It just prevents users from claiming rewards?

ben2x4 (changes requested):
Thanks for the work here, but I don't think I really agree with the implementation here. These changes will lead to contract breaking edge cases. For example, if we withdraw the rewards and some users have not claimed their rewards there will be issues if the contract is refilled and these users with unclaimed prior rewards claim. It will result in an overdraw of the rewards available and making it so some users can't claim the rewards they are owed. I don't think we should include features with such sharp edges.

This contact works because once rewards are being distributed they cannot be withdrawn without serious issues to the underlying math. The better design would be to build a contract that is an distributor for the rewards and refills this contract when the period finishes. This way funds can be withdrawn from the distribution contract easily without causes issues to the underlying math in the rewards contract. 

Happy to chat more about potential options here.

contracts/cw20-stake-external-rewards/src/contract.rs
bekauz: It was meant to disable the claims to get time to investigate things

contracts/cw20-stake-external-rewards/src/contract.rs
JakeHartnell: Agree with @ezekiiel here.

0xekez:
after discussion with @ben2x4 and @bekauz, we're going to try another approach here which makes this method less vulnerable to bad states where a withdraw happens while there are pending rewards.

the new approach is to build a separate funding contract (similar to `cw20-stake-rewards distributor`) with a loose spec as follows:

- admin instantiates contract and specifies the amount that should be funded
- anyone can call fund and it will attempt to fund the contract (this will fail if rewards are ongoing, or insufficient balance)
- admin can withdraw
- admin can update amount that should be funded
- this contract is the "manager" of the external rewards contract
- withdraw allows withdrawing any specified amount / denom
- funding is in terms of a denomination specified at instantiation time

PR #490, opened 2022-10-06
More information pointing to a dead link

CLHRAE:
It looks like https://github.com/DA0-DA0/dao-contracts/wiki/DAO-DAO-v1-Contracts-Design is no longer a valid link. 

Shall we delete the following? 'For more information about how these modules fit together see [this](https://github.com/DA0-DA0/dao-contracts/wiki/DAO-DAO-v1-Contracts-Design) wiki page.'

Or is there information in the wiki we can update to

Or do we need to create information in the wiki about the cw-core contract.


CLHRAE:
I have near enough zero experience with github. Can someone confirm that this is how I should be raising these types of issues?   Here to learn.

ezekiiel:
@CLHRAE , for stuff like this you're best making an issue the contents of which can be the same as the text here. :) generally speaking:

_an issue_, tracks work that may eventually be done. for example: we should update the dead link.

_a PR_, is a request to make a change. for example, if you wanted to go ahead and fix the link you'd make a PR here with that change, people would review it, and then you'd merge it once everyone agreed it was a good idea.

If you want to push a change here to fix the dead link that would be awesome!

Issue #368, opened 2022-06-22
Add absolute count thresholds to multiple choice proposals


blue-note:
should mirror the ActiveThreshold::AbsoluteCount in cw-proposal-single contract

PR #491, opened 2022-10-06
Absolute count threshold for multi-proposals 

0xArbi:
Related issue https://github.com/DA0-DA0/dao-contracts/issues/368

### Approach

- Pull the `{ quorum: PercentageThreshold }` constraint into a `MultipleProposalThreshold` enum that also allows configuring for a `{ threshold: Uint128 }` constraint
- Update logic in `contract.rs` to accomodate for these changes when verifying the status of a proposal
- Amend many of the tests to accomodate for this new logic

### DIFF
--- a/contracts/cw-proposal-multiple/src/proposal.rs
+++ b/contracts/cw-proposal-multiple/src/proposal.rs
@@ -125,7 +125,7 @@ impl MultipleChoiceProposal {
             VotingStrategy::SingleChoice(threshold) => {
                 match threshold {
                     MultipleProposalThreshold::Absoulute { threshold } => {
-                        if self.votes.total() < Uint128::from(*threshold) {
+                        if self.votes.total() < *threshold {
                             return Ok(false);
                         }
                         
--- a/contracts/cw-proposal-multiple/src/proposal.rs
+++ b/contracts/cw-proposal-multiple/src/proposal.rs
@@ -616,7 +616,7 @@ mod tests {
         // High Precision rounding
         let voting_strategy = VotingStrategy::SingleChoice(
             MultipleProposalThreshold::Percentage { quorum: voting::threshold::PercentageThreshold::Percent(
-                cosmwasm_std::Decimal::percent(10),
+                cosmwasm_std::Decimal::percent(100),
             )}
         );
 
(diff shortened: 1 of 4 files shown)


blue-note:
wow so fast! nice!

0xekez:
thanks for this change! looks good at a skim.

we're currently getting really close to releasing v2 of the contracts. thinking we're going to wait on landing this change until that is live. there will be some rebase conflicts here as we're doing some file renaming, but we can return to this / get those fixed once v2 is out!

contracts/cw-proposal-multiple/src/voting_strategy.rs
```diff
@@ -1,24 +1,30 @@
+use cosmwasm_std::Uint128;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use voting::threshold::{validate_quorum, PercentageThreshold, ThresholdError};
+use voting::threshold::{validate_count, validate_quorum, PercentageThreshold, ThresholdError};
+
+/// Determines the way votes are counted.
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum MultipleProposalThreshold {
```
blue-note, line 9: I agree it may be a good idea to have a separate Threshold type for multiple choice to avoid confusion because of the differences in the ways thresholds are used between the two proposal modules. however, can we move this to `packages/cwd-voting/src/threshold.rs`? Also, can we change the names to `PercentageQuorum` and `AbsoluteCountQuorum` just to make it super clear how they are functioning (all MC thresholds will only function as a quorum after which the highest vote will be selected)

contracts/cw-proposal-multiple/src/proposal.rs
```diff
@@ -200,37 +214,33 @@ impl MultipleChoiceProposal {
 
     /// Find the option with the highest vote weight, and note if there is a tie.
     pub fn calculate_vote_result(&self) -> StdResult<VoteResult> {
-        match self.voting_strategy {
```
blue-note, line 203: good catch, it was implemented with multiple voting strategies before

PR #492, opened 2022-10-07
chore: removed last_updated and created timestamps

### LINKED ISSUE #485: Remove timestamps from contracts
this has been a journey, but after discussion with @NoahSaso i'd like to suggest that we remove timestamps used only in the frontend from our contracts. in particular, this would mean removing the `last_updated` field from proposals in `cw-proposal-single` and `cw-proposal-multiple`.

as a guiding philosophy: we should keep our on-chain data as simple as possible. it is much easier to fix a bug in an indexer than to fix a bug in our contracts. in the case of timestamps, we can get that data pretty much effortlessly with an indexer and we already [have a couple](https://github.com/DA0-DA0/subquery-indexers) going and doing so.

yay! we get to remove code!

onewhiskeypls:
NOTE: cw-proposal-multiple is in a bad state, will need to either PR to jake's branch or wait until the 2 have been merged.

partially closes #485


ezekiiel (approved):
awesome! thank you!

PR #493, opened 2022-10-07
node dev dependency

bekauz:
Running `just gen-typescript` would fail to generate schemas because of `String.replaceAll()` being unsupported for Node.js versions below v15.
https://github.com/DA0-DA0/dao-contracts/blob/1a2accbb425ba928845fa0a0f4231cb1fe9824a6/typescript/src/codegen.ts#L109


ezekiiel (approved):
😎

de-husk (approved):
thanks!

PR #504, opened 2022-10-11
Fix Gas Diff

de-husk:
I noticed the main gas file download broke after the name change PR merged: https://github.com/DA0-DA0/dao-contracts/actions/runs/3230162112/jobs/5288290668#step:13:24

But also, this PR will also pull in @0xekez 's fix from the cw-nfts PR (where it wasnt actually gracefully failing if the main's gas file wasn't found).


0xekez (approved):
nice catch! thanks :)

JakeHartnell (approved):

PR #510, opened 2022-10-14
Add cw20-vest and compatibility updates to cw20-staked-balance-voting

teddyknox:
Hi all, this is my first foray into contributing to dao-dao, thanks in advance for considering.

This is a generic vesting contract for locking up DAO tokens and releasing them to vestees over time according to per-vestee schedules. The vesting contract stakes tokens internally to accumulate funding rewards accrued before vested tokens are claimed. This contract integrates modularly with cw20-base and cw20-stake; there is no overlap in functionality with either. cwd-voting-cw20-staked required some changes in order to grant vestees full control over the voting power of their unvested shares.

In a vacuum, usage of this contract goes like this:
1. The contract is instantiated with a token contract, a staking contract, and a list of (address, vesting_schedule).
2. The contract is now in an unfunded state. Once the contract is funded with the full amount to be vested over the lifetime of all vesting schedules, the contract "activates".
3. Once the contract is activated, users can call "vest" on the contract begin the unstaking process for their vested shares. The unstaking_duration still applies here, but with some changes to cw20_stake that could be avoided.
4. Finally, once a user's tokens in c20-vest are unstaked, they can be claimed with a call to "claim".
5. From here the user will likely opt to restake the tokens or sell them.

I also made changes to cwd-voting-cw20-staked to support instantiating/configuring cw20-vest at deploy time.

A vesting schedule is a list of monotonically increasing points (timestamp, amount) which may be piecewise linearly interpolated to describe the total amount vested at any given time. Vesting cliffs can be implemented by making the first point's amount greater than 0 and in the future.

### Testing
* cw20_stake has decent test coverage right now but I will be adding much more.
* cwd-voting-cw20-staked has no tests of this functionality yet (hence the draft status)
* existing tests are passing :)

### Open Questions
* cw20-vest currently assumes usage with a cw20-stake contract. Is this a good assumption or should staking be optional?
* I'm curious whether my query API design passes snuff with respect to conventions.


teddyknox:
Bump. What I will do is write tests and then do a production instantiation to demo this software.

contracts/vesting/cw20-vest/src/contract.rs
```diff
@@ -0,0 +1,1219 @@
...
+                funds: vec![],
+            };
+            Ok(Response::new()
+                .add_message(claim_msg)
+                .add_message(transfer_msg)
+                .add_attribute("action", "claim")
+                .add_attribute("from", info.sender)
+                .add_attribute("amount", release))
+        } else {
+            let transfer_msg = cosmwasm_std::WasmMsg::Execute {
```
0xekez, line 322: these two transfer messages are the same. can probably clean up the if statements here to reflect that.

```
let messages = if balance < release {
  let claim_msg = // ...
  vec![claim_msg]
} else {
  vec![]
};
messages.push(transfer_msg);
```

contracts/vesting/cw20-vest/src/contract.rs
```diff
@@ -0,0 +1,1219 @@
...
+    ) -> Result<Response, ContractError> {
+        let new_owner = new_owner
+            .map(|new_owner| deps.api.addr_validate(&new_owner))
+            .transpose()?;
+        let new_manager = new_manager
+            .map(|new_manager| deps.api.addr_validate(&new_manager))
+            .transpose()?;
+
+        let mut config: Config = CONFIG.load(deps.storage)?;
+        if Some(info.sender.clone()) != config.owner && Some(info.sender.clone()) != config.manager {
```
0xekez, line 353: you can get rid of the clones here with `map_or`. for example, `config.owner.as_ref().map_or(true, |owner| info.sender != owner)`.

contracts/vesting/cw20-vest/src/contract.rs
```diff
@@ -0,0 +1,1219 @@
...
+        info: MessageInfo,
+        new_owner: Option<String>,
+        new_manager: Option<String>
+    ) -> Result<Response, ContractError> {
+        let new_owner = new_owner
+            .map(|new_owner| deps.api.addr_validate(&new_owner))
+            .transpose()?;
+        let new_manager = new_manager
+            .map(|new_manager| deps.api.addr_validate(&new_manager))
+            .transpose()?;
```
0xekez, line 350: if you were making a DAO with this vesting contract would you like to appoint a manger, or could we remove this?

contracts/vesting/cw20-vest/src/contract.rs
```diff
@@ -0,0 +1,1219 @@
...
+                config
+                    .manager
+                    .map(|a| a.to_string())
+                    .unwrap_or_else(|| "None".to_string()),
+            ))
+    }
+
+}
+
+pub mod query {
```
0xekez, line 384: probably want to add a query here for the config and one to get the vesting schedule ofr an address.

contracts/voting/cwd-voting-cw20-staked/src/contract.rs
```diff
@@ -137,8 +138,27 @@ pub fn instantiate(
...
             }
 
+            // Has the acceptable side-effect of pre-sorting the vesting schedules chronologically
+            if let Some(vesting_info) = &mut vesting_info {
+                let mut vest_total = Uint128::from(0u128);
+                let schedules = &mut vesting_info.schedules;
+                for schedule in schedules.into_iter() {
+                    schedule.vests.sort_by_key(|v| v.expiration);
+                    if let Some(last) = schedule.vests.last() {
+                        vest_total = vest_total.checked_add(last.amount)?;
```
0xekez, line 148: isn't the total decided by the last point in the vesting schedule? why sum here?

contracts/voting/cwd-voting-cw20-staked/src/contract.rs
```diff
@@ -137,8 +138,27 @@ pub fn instantiate(
...
+                    }
+                }
+                if vest_total > Uint128::zero() {
```
0xekez, line 151:
```suggestion
                if !vest_total.is_zero() {
```

contracts/voting/cwd-voting-cw20-staked/src/contract.rs
```diff
@@ -407,12 +522,52 @@ pub fn reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractEr
 
                     // Save staking contract addr
                     STAKING_CONTRACT.save(deps.storage, &staking_contract_addr)?;
+                    let vesting_info = VESTING_INFO.load(deps.storage)?;
```
0xekez, line 525: what is the point of keeping the vesting info in state? can't imagine we need this after instantiation?

0xekez (commented):
thanks for the PR. :) left a long review. great work.

generally speaking, the vesting contract seem good. some issues with validation and rewards, but i like this design. the claims handling is smart and i like the idea of actually staking tokens to collect rewards.

that said, i am a little concerned about the modifications to cwd-voting-cw20-staked. we're introducing a fair bit of complexity into what, from a security perspective, is one of the most important contracts in DAO DAO. i'd encourage you to spend some time considering how we could implement this with zero changes to that contract.

it feels like the primitive that you're looking for here is really a voting power module that sums values from two other ones:

image

I think if we were clever with the summation contract's instantiation process we may be able to use that and not change the staked balance voting contract by passing addresses along through its instantiation process. 

happy to chat about this more. feel free to join the discord and give me a ping!

teddyknox:
Thanks so much for your review. I've been swamped but should get another draft out this week.

contracts/vesting/cw20-vest/src/contract.rs
```diff
@@ -0,0 +1,1219 @@
...
+                vest_total: beta_config.vest_total,
+            };
+            deps.storage.set(b"config", &to_vec(&new_config)?);
+            Ok(Response::default())
+        }
+        MigrateMsg::FromCompatible {} => Ok(Response::default()),
+    }
+}
+
+fn calculate_cumulative_vest(env: &Env, vests: &Vec<Vest>) -> StdResult<Uint128> {
```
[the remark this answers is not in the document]
teddyknox: Thanks for the explanation

contracts/vesting/cw20-vest/src/contract.rs
teddyknox: Makes sense, but why the `.as_ref()` ?

contracts/vesting/cw20-vest/src/contract.rs
teddyknox: Adding a vesting schedule query endpoint. Config should be there already.

contracts/voting/cwd-voting-cw20-staked/src/contract.rs
teddyknox: Each recipient has their own vesting schedule defined in absolute token terms. So we need to sum up the endpoints to get the total that will be vested to all recipients.

contracts/voting/cwd-voting-cw20-staked/src/contract.rs
teddyknox: We either instantiate the vest contract immediately or we intantiate it in the reply after instantiating the cw20 and/or the cw20-stake. If we're instantiating in the reply we need to some how persist or transfer vest_info to that reply section. I saw this pattern being used in cw20-stake so I copied it.

teddyknox (commented):
I pushed updates corresponding to your feedback and asked a few clarifying questions. 

Big Todos:
* I need to update the voting contract to make it more modular. In addition it almost seems like we should move the logic for instantiating cw20, cw20-stake, cw20-vest contracts to a factory contract.
* I also want to do a refactor where we don't pass vesting schedules to the contract on instantiation, but instead configure schedules and fund the contract at the same time with a parameterized call to the receive endpoint. As part of this, I would add the possibility for each to have multiple concurrent or non-overlapping vesting schedules.
* I want to add a `execute_cancel_schedule()` method in case a DAO employee leaves etc. In this case the unvested tokens would be returned to the core contract.
* Last but not least, I'll continue to work on tests and getting checks to pass.

Thanks for your patience on this.

0xekez:
@teddyknox how hard do you think it would be to make this such that you could upgrade an existing cw20 to this contract? that would be a really killer feature because existing DAOs with illiquid tokens could upgrade to this configuration.

teddyknox:
@0xekez it should be possible in the current. Currently all you'd need to do is deploy cw20-vest and upgrade the voting contract. Am I understanding you correctly?

0xekez:
@teddyknox i'm wondering more if we could make this so that a DAO with an existing cw20 token and staking contract could upgrade to this and in doing so vest all outstanding tokens? if this is really hard, ignore me. :)

teddyknox:
> and in doing so vest all outstanding tokens?

What is our definition of outstanding tokens in this case? 

@0xekez

0xekez:
closing for now - code is great, but i think we need to spend some more time thinking about what our requirements for vesting are before we land something like this. it's possible that payroll work happening in #537 will supercede our needs for vesting.

thanks for the contribution @teddyknox !

PR #511, opened 2022-10-14
upgrade to cosmwasm 1.1 

blue-note:
upgrades to cosmwasm 1.1 as per [this doc.](https://github.com/CosmWasm/cosmwasm/blob/main/MIGRATING.md#100---110)

things that need careful eyes:

- the query macro modifications (packages/cwd-macros/src/lib.rs)
- the schema and typescript codegen modifications 
- the QueryMsg return attributes


contracts/external/cw-admin-factory/Cargo.toml
```diff
@@ -16,17 +16,16 @@ backtraces = ["cosmwasm-std/backtraces"]
...
-cosmwasm-storage = "1.0.0"
-cw-storage-plus = "0.13.2"
-cw2 = "0.13.2"
-schemars = "0.8.8"
-serde = { version = "1.0.137", default-features = false, features = ["derive"] }
+cosmwasm-std = "1.1"
+cosmwasm-storage = "1.1"
+cosmwasm-schema = "1.1"
+cw-storage-plus = "0.15"
+cw2 = "0.15"
```
0xekez, line 23: could we bump all of our CosmWasm dependencies as part of this change? still some 0.13 in here.

0xekez (commented):
looking great! sorry for being slow to get to this. have been taking a very needed break.

two things:

1. would be great to update all of our CosmWasm dependencies to the latest versions including the cw-plus and cw-nfts ones.
2. as part of this PR, wdyt about moving to a cargo workspace? example [here](https://github.com/CosmWasm/cw-nfts/pull/89).

I think (2) above will actually make this PR a lot easier and less error prone. :)

contracts/external/cw-admin-factory/Cargo.toml
JakeHartnell: Yes please.

contracts/cwd-core/src/msg.rs
```diff
@@ -139,78 +144,95 @@ pub enum ExecuteMsg {
...
-    /// start_after is bound exclusive and asks for a string address
+    /// Returns all SubDAOs with their charters in a vec.
+    /// start_after is bound exclusive and asks for a string address.
+    #[returns(Vec<SubDao>)]
     ListSubDaos {
         start_after: Option<String>,
         limit: Option<u32>,
     },
     /// Implements the DAO Star standard: https://daostar.one/EIP
+    #[returns(DaoURIResponse)]
```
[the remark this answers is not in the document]
JakeHartnell: This is the way the new `cosmwasm-schema` works.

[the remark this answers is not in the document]
blue-note: do you mean why we needed to add `DaoURIResponse`? it's because if we keep it as returning `Option<String`, cosmwasm typescript gen generates a `Nullable_String` type, but then wrongly annotates this as returning `NullableString` for some reason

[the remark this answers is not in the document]
blue-note: let me know if anyone knows what's up with that

contracts/external/cw-admin-factory/Cargo.toml
blue-note: introduced the workspace-inheritance!

contracts/cwd-core/src/msg.rs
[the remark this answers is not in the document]
0xekez: @pyramation any thoughts?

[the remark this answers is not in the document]
pyramation: the latest `cosmwasm/ts-codegen` supports CosmWasm 1.1 — things should work out of the box. 

just make sure to remove the old schemas since it expects a single JSON. 

finally, if there are any issues just let me know we'll take care of them.

Cargo.toml
```diff
@@ -28,3 +28,29 @@ incremental = false
 # Please do not disable these. Doing so will cause overflow checks in
 # all workspace members to stop working. Overflows should be errors.
 overflow-checks = true
+
+[workspace.dependencies]
```
JakeHartnell, line 32: Really nice. :heart:

Cargo.toml
```diff
@@ -28,3 +28,29 @@ incremental = false
...
+cosmwasm-std = { version = "1.1", features = ["ibc3"] }
+cosmwasm-storage = { version = "1.1" }
+cosmwasm-schema = { version = "1.1" }
+cw-utils = "0.15"
+thiserror = { version = "1.0.30" }
+cw-storage-plus = "0.15"
+cw2 = "0.15"
+cw20 = "0.15"
+cw721 = "0.15"
+cw20-base = "0.15"
```
JakeHartnell, line 42: A lot of these are at `0.16` now. May as well bump them to the latest.

contracts/cwd-core/src/msg.rs
```diff
@@ -139,78 +144,95 @@ pub enum ExecuteMsg {
...
+#[cw_serde]
+#[derive(QueryResponses)]
 pub enum QueryMsg {
     /// Get's the DAO's admin. Returns `Addr`.
+    #[returns(Addr)]
     Admin {},
-    /// Get's the currently nominated admin (if any). Returns
-    /// `AdminNominationResponse`.
+    /// Get's the currently nominated admin (if any).
+    #[returns(AdminNominationResponse)]
```
0xekez, line 154: this should fix the name resolution lints you're getting

```suggestion
    #[returns(::query::AdminNominationResponse)]
```

contracts/proposal/cwd-proposal-single/src/contract.rs
```diff
@@ -865,8 +866,8 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
                         description: prop.description,
                         proposer: prop.proposer,
                         start_height: prop.start_height,
-                        min_voting_period: prop.min_voting_period,
-                        expiration: prop.expiration,
+                        min_voting_period: prop.min_voting_period.map(v1_expiration_to_v2),
+                        expiration: v1_expiration_to_v2(prop.expiration),
```
0xekez, line 870: nice :)

0xekez (approved):
lg2m. :) thanks for going through the grind for this.

packages/cwd-macros/src/lib.rs
```diff
@@ -138,8 +174,15 @@ pub fn token_query(metadata: TokenStream, input: TokenStream) -> TokenStream {
     let mut ast: DeriveInput = parse_macro_input!(input);
     match &mut ast.data {
         syn::Data::Enum(DataEnum { variants, .. }) => {
-            let info: Variant = syn::parse2(quote! { TokenContract {} }).unwrap();
+            let mut info: Variant = syn::parse2(quote! { TokenContract {} }).unwrap();
+
+            let info_attr = syn::Attribute::parse_outer
+                .parse2(quote! {
+                    #[returns(Addr)]
```
0xekez, line 181: to avoid lint issues i think we need to make all of these `#[returns(...)]` decorators use the absolute path. for this one:

```suggestion
                    #[returns(::cosmwasm_std::Addr)]
```

0xekez: ah, sweet emergent complexity. this does not actually work for queries here that return stuff from `cwd-interface`.

for example, `#[returns(::cwd_interface::voting::VotingPowerAtHeightResponse)` breaks when we derive that macro inside the `cwd_interface` crate as there `cwd_interface` has no meaning. for that particular crate, we instead want to do `#[returns(::crate::voting::VotingPowerAtHeightResponse)`.

0xekez:
two things from working off a fork that i think we can improve here:

1. the [rfc for this](https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html) uses `dep = { workspace = true }` everywhere (does not use `dep.workspace = true`).  i personally prefer that format as understanding it doesn't require understanding new YAML syntax.
2. it would be really nice to add local dependencies to the workspace. the rfc talks about [this here](https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html#effect-resolution-for-relative-path-dependencies). doing this will make moving files around _much_ easier and makes sure our versioning is consistent everywhere.

@blue-note do you have time to take a look at these? i can as well but it'll bump into other stuff.

blue-note:
> two things from working off a fork that i think we can improve here:
> 
> 1. the [rfc for this](https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html) uses `dep = { workspace = true }` everywhere (does not use `dep.workspace = true`).  i personally prefer that format as understanding it doesn't require understanding new YAML syntax.
> 2. it would be really nice to add local dependencies to the workspace. the rfc talks about [this here](https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html#effect-resolution-for-relative-path-dependencies). doing this will make moving files around _much_ easier and makes sure our versioning is consistent everywhere.
> 
> @blue-note do you have time to take a look at these? i can as well but it'll bump into other stuff.

sure, I can take a look tomorrow!

typescript/src/codegen.ts
```diff
@@ -1,196 +1,177 @@
...
-  const codegenResponses: Promise<void>[] = [];
-  for (const spec of compilationSpecs) {
-    codegenResponses.push(generateTs(spec));
-  }
-  await Promise.all(codegenResponses);
-
-  log(`code generation complete`, LogLevels.Normal);
-}
-
-main();
```
0xekez, line 196: any particular reason to remove this? this new one breaks if you add or remove a contract atm which is a bit of a bummer.

blue-note: the script had to change to deal with the different output schemas / the new codegen, but we can prob find a way to not have to specify contract names specifically

blue-note: no worries, I made the script a bit better

JakeHartnell (approved):

PR #519, opened 2022-11-03
Remove unused contracts.

0xekez:
this removes the native staked token voting contracts and the token swap contract. it is not clear to me that these will be used in the future as token swaps on other chains are likely to be more trouble than they're worth and native staking is likely to be made obsolete by multichain support and a governance module replacement.


0xekez:
ok changed my mind.

token swap good for token swaps with growth fund. native good for x/dao and token factory. :)

PR #527, opened 2022-11-08
Refactor and simplify cw721 voting contract.

0xekez:
- Removes the manger from contract state.
- Refactors the state layout as detailed here: <https://github.com/DA0-DA0/dao-contracts/discussions/523>
- Removes redundant queries.
- Refactors testing (100% coverage now).
- Adds integration tests to ensure `MAX_CLAIMS` in in bounds for gas usage.


JakeHartnell (approved):
:fire:

de-husk:
Let's leave the `MAX_CLAIMS` integration test commented out until cosm-orc supports batch tx operations. We can leave a TODO and come back and uncomment it once the runtime is reasonable.

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
```diff
@@ -30,6 +26,7 @@ pub fn instantiate(
     info: MessageInfo,
     msg: InstantiateMsg,
 ) -> Result<Response<Empty>, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```
blue-note, line 29: could we add a default migrate function where we set contract version as well?

contracts/voting/cwd-voting-cw721-staked/src/error.rs
```diff
@@ -3,33 +3,30 @@ use thiserror::Error;
...
 
     #[error("Nothing to claim")]
     NothingToClaim {},
 
-    #[error("Invalid token")]
+    #[error("Invalid token. Got ({received}), expected ({expected})")]
     InvalidToken { received: Addr, expected: Addr },
 
     #[error("Unauthorized")]
     Unauthorized {},
```
blue-note, line 16: can we make this err a bit more specific?

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
```diff
@@ -348,17 +247,16 @@ pub fn execute_add_hook(
...
 pub fn execute_remove_hook(
     deps: DepsMut,
-    _env: Env,
     info: MessageInfo,
     addr: String,
 ) -> Result<Response, ContractError> {
-    let addr = deps.api.addr_validate(&addr)?;
     let config: Config = CONFIG.load(deps.storage)?;
-    if config.owner != Some(info.sender.clone()) && config.manager != Some(info.sender) {
+    if config.owner.map_or(true, |owner| owner != info.sender) {
```
blue-note, line 254: nice :)

contracts/voting/cwd-voting-cw721-staked/src/state.rs
```diff
@@ -1,43 +1,98 @@
...
+        .map(|_| ())
+}
+
+pub fn register_unstaked_nft(
+    storage: &mut dyn Storage,
+    height: u64,
+    staker: &Addr,
+    token_ids: &[String],
+) -> Result<(), ContractError> {
+    let subtractor = |amount: u128| {
```
blue-note, line 73: i love this

contracts/voting/cwd-voting-cw721-staked/src/testing/execute.rs
```diff
@@ -0,0 +1,148 @@
...
+use cw721_base::MintMsg;
+use cw_multi_test::{App, AppResponse, Executor};
+
+use anyhow::Result as AnyResult;
+use cw_utils::Duration;
+
+use crate::msg::ExecuteMsg;
+
+// Shorthand for an unchecked address.
+macro_rules! addr {
```
blue-note, line 12: cool!

blue-note:
on a first pass this looks really great. I am really enjoying reading your code. also appreciate the detailed write up which helps ratify how to think about contract runtime complexity (avoid > O(1) operations for things which cannot be paginated). In general I think using auxiliary maps to improve runtime complexity is a good way to go. going to return to it tomorrow.

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
0xekez: i'm actually in the camp of not having one. i feel having one that does nothing is more liable to be confusing than helpful. what we've seen is that migration code is actually non-trivial so probably not much benefit to cookie cutter ones. wdyt? if there's an advantage to having one, i'm happy to add it in.

contracts/voting/cwd-voting-cw721-staked/src/error.rs
0xekez: great point. updated to specify that the issue at hand is that the sender is not the owner.

contracts/voting/cwd-voting-cw721-staked/src/state.rs
0xekez: aha yeah always exciting when you get to use a higher order function :)

contracts/voting/cwd-voting-cw721-staked/src/testing/execute.rs
0xekez: thanks! i've been reading this to learn about macros: https://veykril.github.io/tlborm/introduction.html

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
blue-note: my argument for is that I remember working on a frontend change against the contracts some time back and realizing we had forgotten to call `set_contract_version` in the migrate functions so the frontend wasn't getting the updated version. I think it could be helpful to have a migrate function that sets version by default in every contract. the argument could also be made that you only need to add a migrate function when migration will start happening, but it's also easy to forget to add one when there aren't state migrations

blue-note (approved):
lgtm, awesome work! I will leave the migration thing to your discretion

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
0xekez: this is a good point. i'm going to go in the direction of having no migrate function, the ultimate reason being that having code that has no purpose is, imo, suboptimal (in our current circumstances, we'd never use it). let's keep an eye out for these kinds of things though moving forward!

PR #528, opened 2022-11-08
Refactor tests for cw721 voting contract.

0xekez:
This refactors the tests for our cw721 voting contract to use the pattern we started moving towards with the `cwd-proposal-single` refactor. The changes from that refactor being:

1. Multitest contracts are pulled from the `cwd-testing` package instead of duplicated locally.
2. Testing query and execute functions are public so that other contracts can import them.
3. Tests return `anyhow::Result` which _may_ allow for better error checking, though doing that seems to remove the ability to recover the original error which makes matching on error type a little more cumbersome.

Follow up work will start some adversarial tests for this.


JakeHartnell (approved):
Very nice!

PR #529, opened 2022-11-09
Update code coverage exclusions to match new names.

0xekez:


de-husk (approved):

.github/workflows/codecov.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Generate code coverage
         run: >
-          cargo +nightly tarpaulin --verbose --workspace --out Xml --exclude integration_tests --exclude-files debug/* *test*.rs packages/testing/* packages/cw-core-macros/* legacy/* ci/*
+          cargo +nightly tarpaulin --verbose --workspace --out Xml --exclude integration_tests --exclude-files test-contracts/* *test*.rs packages/cwd-macros/* packages/cwd-testing/* ci/*
```
de-husk, line 16: `integration_tests` is also the old name. It is `integration-tests` now. 

But it's working now because the integration tests are ignored by default.

So we might as well just remove this `--exclude integration_tests` here and also in the root `.cargo/config`.

0xekez: done. thanks :) nice catch.

JakeHartnell (approved):

PR #532, opened 2022-11-10
multiple choice cleanups for frontend

blue-note:
Some stuff that is coming up as I work on the frontend (there may be more, but want to get this in)

- fix a couple return type annotations
- simplify proposal.msgs type
- allow multiple choice options to have their own titles
- increase the max number of options


0xekez (approved):
simple and clean! nice!

JakeHartnell (approved):

PR #536, opened 2022-11-16
V2 Staging

### LINKED ISSUE #514: Implement proposal approval flow
It is often the case that DAOs want the ability to "approve" a proposal before it goes up for voting in the main DAO. Typically, this takes the shape of a subDAO which must approve proposals before they go live.

image

This is likely easiest built as an implementation of `cwd-pre-propose-base` with some additional logic around approvals. Building on top of this contract will mean that deposit logic will already be implemented for us and the only additional logic will be approvals.

JakeHartnell:
Closes #514


contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    Ok(resp.set_data(to_binary(&ModuleInstantiateCallback {
+        msgs: vec![CosmosMsg::Wasm(WasmMsg::Execute {
+            contract_addr: addr.to_string(),
+            msg: to_binary(&PreProposeApprovalExecuteMsg::AddProposalSubmittedHook {
+                address: env.contract.address.to_string(),
+            })?,
+            funds: vec![],
+        })],
+    })?))
```
0xekez, line 104: this is really, really clean.

my concern: a single message can only have one piece of associated data. for example, if a submessage calls `set_data`, and then the parent message does as well, the submessage data gets overridden. if we take this approach, it would be great to "prove" that won't be a problem.

a possible alternative: we expose a `ExecuteMsg::PreProposeCallback` on the proposal module. calling that will cause the proposal module to execute `ExecuteMsg::ExecuteProposalHook` on the core module. i _think_ this way we can have this behavior without using set data or modifying the core module. wdyt?

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
```diff
@@ -0,0 +1,272 @@
...
+            })?,
+            funds: vec![],
+        }),
+        Status::Rejected => Some(WasmMsg::Execute {
+            contract_addr: approval_contract.into_string(),
+            msg: to_binary(&PreProposeApprovalExecuteMsg::Extension {
+                msg: ApprovalExt::Approve { id: pre_propose_id },
+            })?,
+            funds: vec![],
+        }),
```
0xekez, line 223: should these be flipped?

passed -> approve
rejected -> reject

it may be cleaner to just do this if status is executed or rejected. otherwise we're going to reject / approve twice. once on passed, and once on executed.

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
```diff
@@ -0,0 +1,272 @@
...
+        // TODO Maybe don't default reject?
+        _ => None,
+    };
+
+    // TODO two match expressions annoys me
+    match msg {
+        Some(msg) => Ok(Response::default().add_message(msg)),
+        // TODO Not sure what we want to do here...
+        None => Err(PreProposeError::Unauthorized {}),
```
0xekez, line 232: this should be unreachable. the status you don't check is `ExecutionFailed` which will never actually fire as a hook. to reach that status:

1. proposal is executed
2. executed status hook fires
3. if proposal execution fails, execution failed status is set

this may be unintuitive though. what would you expect in that case as a contract author?

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
```diff
@@ -0,0 +1,272 @@
...
+                    let pre_propose_id = CURRENT_PRE_PROPOSE_ID.load(deps.storage)?;
+
+                    // Parse SubMessageResponse to get proposal id
+                    let event = &result.events[result.events.len() - 1];
+                    let proposal_id = event.attributes[event.attributes.len() - 1]
+                        .value
+                        .parse::<u64>()?;
+
+                    // Save mapping of proposal id to pre propose id
+                    PROPOSAL_IDS.save(deps.storage, proposal_id, &pre_propose_id)?;
```
0xekez, line 263: i think a less error prone way to get this mapping would be to handle the `ExecuteMsg::ProposalCompletedHook`. that way you don't have to dispatch this as a submessage and read the responses. that hook has information about the proposal id.

packages/cwd-pre-propose-base/src/error.rs
```diff
@@ -15,6 +19,15 @@ pub enum PreProposeError {
...
     Deposit(#[from] DepositError),
 
+    #[error(transparent)]
+    Hooks(#[from] HookError),
+
+    #[error(transparent)]
+    ParseReplyError(#[from] ParseReplyError),
+
+    #[error(transparent)]
+    ParseIntError(#[from] ParseIntError),
```
0xekez, line 29: these should probably be in the implementers error enum. the base contract doesn't actually do any number parsing i don't think?

packages/cwd-pre-propose-base/src/msg.rs
```diff
@@ -70,6 +71,13 @@ pub enum ExecuteMsg<ProposalMessage, ExecuteExt> {
     /// will do nothing if this variant is executed.
     Extension { msg: ExecuteExt },
 
+    /// Adds a proposal submitted hook. Fires when a new proposal is
```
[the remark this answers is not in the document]
0xekez: great idea to have these hooks btw

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// Reject a proposal, only callable by approver
+    Reject { id: u64 },
+}
+
+#[cw_serde]
+pub enum QueryExt {
+    /// List the approver address
+    Approver {},
+    /// List of proposals awaiting approval
+    Proposals {},
```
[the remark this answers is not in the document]
JakeHartnell: Yeah, have a TODO for that somewhere.

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
0xekez: we could modify that hook logic so that only one of `executed` or `execution_failed` is fired. that may be the way here.. if you agree i can do that :)

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,366 @@
...
+                        pre_propose_id: id,
+                    },
+                })?,
+                funds: vec![],
+            };
+            Ok(SubMsg::new(execute_msg))
+        })?;
+
+    // Increment next id
+    CURRENT_ID.save(deps.storage, &(id + 1))?;
```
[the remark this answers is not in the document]
JakeHartnell: Great!

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,366 @@
...
+
+pub fn execute_approve(
+    deps: DepsMut,
+    info: MessageInfo,
+    id: u64,
+) -> Result<Response, PreProposeError> {
+    // Check sender is the approver
+    let approver = APPROVER.load(deps.storage)?;
+    if approver != info.sender {
+        return Err(PreProposeError::Unauthorized {});
```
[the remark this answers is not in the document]
JakeHartnell: :joy:

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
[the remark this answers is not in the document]
JakeHartnell: Nice!

contracts/pre-propose/cwd-pre-propose-approver/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "cwd-pre-propose-approver"
+version = "0.2.0"
+authors = ["ekez <zekemedley@gmail.com>", "Jake Hartnell <no-reply@no-reply.com>"]
+edition = "2021"
+repository = "https://github.com/DA0-DA0/dao-contracts"
+description = "A DAO DAO pre-propose module for cwd-proposal-single for native and cw20 deposits."
```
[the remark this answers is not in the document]
JakeHartnell: Done.

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
JakeHartnell: We talked about this in Discord, but re-posting here for discoverability...

> Thought about doing something like this, but thought the principle of least privilege would be nice... I wanted to limit the ability for the pre-propose contract to only be able to execute messages for the core module on instantiation.

> The proposal module can’t execute ExcuteProposalHook because when it’s being instantiated (and instantiating the pre-propose contract), it hasn’t been registered... so something like this is needed.

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
JakeHartnell: > we could modify that hook logic so that only one of executed or execution_failed is fired. that may be the way here.. if you agree i can do that :)

Feel free to make commits or PRs into this one. Let's make it nice. : )

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
JakeHartnell: Having trouble getting that hook to fire on proposal passed or rejected... need to investigate...

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    // Check proposal id exists
+    PENDING_PROPOSALS.load(deps.storage, id)?;
+
+    // Remove proposal
+    PENDING_PROPOSALS.remove(deps.storage, id);
+
+    // Handle deposit logic
```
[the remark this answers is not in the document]
JakeHartnell: Unfortunately, we can't with this one. The logic is actually slightly different.

contracts/pre-propose/cwd-pre-propose-approver/src/contract.rs
JakeHartnell: I think we need this mapping of the proposal id in the approver DAO to the pre-proposal id in the DAO seeking approval... can't think of another way to get it as `ExecuteMsg::ProposalCompletedHook` is only executed when the proposal is completed.

packages/cwd-pre-propose-base/src/error.rs
JakeHartnell: Ideally we wouldn't have to parse the Int in a submessage. Open to commits on this.

contracts/pre-propose/cwd-pre-propose-approval-single/README.md
```diff
@@ -0,0 +1,68 @@
...
+            │ Creates prop
+            │ on approval
+            ▼
+┌────────────────────────┐
+│                        │
+│     Proposal Single    │
+│                        │
+└───────────┬────────────┘
+            │
+            │ Normal votin
```
JakeHartnell, line 33:
```suggestion
            │ Normal voting
```

contracts/pre-propose/cwd-pre-propose-approval-single/README.md
```diff
@@ -0,0 +1,68 @@
...
+tokens, or no tokens as a deposit. If a proposal deposit is enabled
+the following refund strategies are avaliable:
+
+1. Never refund deposits. All deposits are sent to the DAO on proposal
+   completion.
+2. Always refund deposits. Deposits are returned to the proposer on
+   proposal completion.
+3. Only refund passed proposals. Deposits are only returned to the
+   proposer if the proposal passes. Otherwise, they are sent to the
+   DAO.
```
JakeHartnell, line 59:
```suggestion
1. Never refund deposits. All deposits are sent to the DAO on proposal
   completion.
2. Always refund deposits. Deposits are returned to the proposer on
   proposal completion and even rejection by the `approver`.
3. Only refund passed proposals. Deposits are only returned to the
   proposer if the proposal is approved and passes. Otherwise, they
   are sent to the DAO.
```

contracts/pre-propose/cwd-pre-propose-approver/README.md
```diff
@@ -0,0 +1,47 @@
...
+┌────────────────────────┐   │         ┌────────────────────────┐ │
+│                        │   │         │                        │ │
+│       Main DAO         │   └─────────┤     Approver DAO       ├─┘
+│                        │             │                        │
+└────────────────────────┘             └────────────────────────┘
+```
+
+## Deposits
+
+This contract does not handle deposit logic, as it works in conjunction with the `cwd-pre-propose-approval-single` contract which handles to the deposits.
```
JakeHartnell, line 47:
```suggestion
This contract does not handle deposits. It works in conjunction with the `cwd-pre-propose-approval-single` contract, which handles the proposal deposits.
```

contracts/pre-propose/cwd-pre-propose-approver/README.md
```diff
@@ -0,0 +1,47 @@
+# Proposal Approver Contract
+
+This contract works in conjuction with `cwd-pre-propose-approval-single` and allows for automatically creating approval proposals when a proposal is submitted for approval.
+
+## Approver Logic
+
+On instantiation, this contract registers a hook with the approval contract to automatically create proposals in the approver DAO.
+
+When this contract recieves a prop as hook from `cwd-pre-propose-approval-single`, it makes an approval prop. If approved, the approval prop calls the approve message on this contract when executed. When prop is rejected and closed it fires off reject call.
```
JakeHartnell, line 9:
```suggestion
When this contract recieves a proposal as hook from `cwd-pre-propose-approval-single`, it makes an approval propose in the approval DAO. If approved, the approval proposal calls the approve message on this contract when executed. If the proposal is rejected and closed it fires off reject call.
```

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,350 @@
...
+        // likely created before we were added to the proposal
+        // module. In that case, it's not our problem and we just
+        // do nothing.
+        None => Ok(Response::default()
+            .add_attribute("method", "proposal_rejected")
+            .add_attribute("proposal", id.to_string())),
+    }
+}
+
+pub fn execute_update_approver(
```
0xekez, line 247: nit: oak audit is going to say "pls do admin nomination flow"

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    // Check sender is the approver or the parent DAO
+    if approver != info.sender && dao != info.sender {
+        return Err(PreProposeError::Unauthorized {});
+    }
+
+    // Validate address
+    let addr = deps.api.addr_validate(&address)?;
+
+    // Add hook
```
0xekez, line 309:
```suggestion
    // remove hook
```

contracts/pre-propose/cwd-pre-propose-approval-single/src/msg.rs
```diff
@@ -0,0 +1,78 @@
...
+    PendingProposal { id: u64 },
+    /// List of proposals awaiting approval
+    #[returns(Vec<PendingProposal>)]
+    PendingProposals {
+        start_after: Option<u64>,
+        limit: Option<u32>,
+    },
+    #[returns(Vec<PendingProposal>)]
+    ReversePendingProposals {
+        start_after: Option<u64>,
```
0xekez, line 58:
```suggestion
        start_before: Option<u64>,
```

0xekez: todo: does `cw-paginate` support this properly in its edge cases?

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,350 @@
...
+            msg: propose_msg_internal.clone(),
+        },
+    )?;
+
+    // Save info about deposits when this prop was created
+    pre_propose_base.deposits.save(
+        deps.storage,
+        id,
+        &(config.deposit_info.clone(), info.sender.clone()),
+    )?;
```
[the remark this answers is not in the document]
JakeHartnell: Ahhhhhh, good point, forgot pre-propose modules could be updated. :thinking:

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
JakeHartnell: Lol.

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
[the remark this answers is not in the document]
0xekez: so thinking from "first principles": why do we need two deposit tracking systems in this contract?

> there are two deposits here, with unique rules. there is an "approval" deposit which is the deposit you pay to put your proposal up for approval and there is the "proposal" deposit. the approval deposit has special semantics; specifically, if a proposal is not approved, the deposit is only refunded if the deposit policy is "always". the proposal deposit is the deposit we are already familiar with.

thus, in order to make a proposal, there are two phases for the deposit. first, it goes through the approval phase, then it goes through the regular deposit phase. we need to track these two things in different maps as they are semantically different.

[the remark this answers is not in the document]
0xekez: this makes the lifetime of a deposit:

1. the deposit is processed as an "approval" deposit.
2. if approved, the deposit is moved into a "proposal" deposit. regular proposal deposit rules now apply.
3. if rejected, approval deposit rejection rules are applied.

we can use the same state layout for these two things, and we can use the same `DepositInfo` config. the trick is that we need to transition them along as described here.

0xekez (approved):
at this point, the broader design of this contract is good. there is one bug (https://github.com/DA0-DA0/dao-contracts/issues/553), and will likely be little things to improve.

approving this anyhow as the plan is to merge this and then followup with the addition of a `NextProposalId` query to the proposal module interface which will resolve the only known problem, then we can do more cleanup as needed.

:shipit:

contracts/pre-propose/cwd-pre-propose-approval-single/README.md
```diff
@@ -0,0 +1,68 @@
...
+   proposer if the proposal is approved and passes. Otherwise, they
+   are sent to the DAO.
+
+This module may also be configured to only accept proposals from
+members (addresses with voting power) of the DAO.
+
+Here is a flowchart showing the proposal creation process using this
+module:
+
+![](https://bafkreig42cxswefi2ks7vhrwyvkcnumbnwdk7ov643yaafm7loi6vh2gja.ipfs.nftstorage.link)
```
blue-note, line 67: this link doesn't seem to work

contracts/pre-propose/cwd-pre-propose-approval-single/src/contract.rs
```diff
@@ -0,0 +1,370 @@
...
+        Some(proposal) => {
+            let proposal_module = PrePropose::default().proposal_module.load(deps.storage)?;
+            let propose_messsage = WasmMsg::Execute {
+                contract_addr: proposal_module.into_string(),
+                msg: to_binary(&ProposeMessageInternal::Propose(proposal.msg))?,
+                funds: vec![],
+            };
+
+            PENDING_PROPOSALS.remove(deps.storage, id);
+            DEPOSIT_SNAPSHOT.save(deps.storage, &proposal.deposit)?;
```
blue-note, line 175: would be great to add comment about why this is needed

0xekez: yeah :) this was a failed design attempt. see: https://github.com/DA0-DA0/dao-contracts/issues/553

0xekez: the solution is to add a query to the proposal module interface (which was introduced in the macros PR) for `NextProposalId {}`. Having that, the trouble where the pre-propose module has to wait for the hook to know the ID will go away as it can query that and immediately create the proposal.

0xekez:
rebased on main. we can now stage the `NextProposalId` refactor on this branch.

PR #538, opened 2022-11-16
Formally verify and fix overflow in cw20-stake stake and unstake methods

0xekez:
this fixes a fairly pathological, non-severe, integer overflow bug in the stake-cw20 contract. after discussion with ben2x4, meow, and bluenote, we've decided that this bug is extremely unlikely to be reached in production; if it is reached, it is entirely recoverable. as such, i'm making the patch public.

see here for a description of the bug, and a recovery strategy should anyone ever hit it:

https://github.com/DA0-DA0/dao-contracts/wiki/Overflow-in-stake-cw20

once this lands, i'll work on some detailed upgrade instructions for DAO DAO DAOs. it is nice we have a bug like this, because it is a great chance to get DAOs practicing doing upgrades in a safe environment. we can run a fire drill next week for DAOs that would like to participate and help them practice upgrading.

### DIFF
--- a/contracts/staking/cw20-stake/src/contract.rs
+++ b/contracts/staking/cw20-stake/src/contract.rs
@@ -226,11 +226,15 @@ pub fn execute_unstake(
     let config = CONFIG.load(deps.storage)?;
     let balance = BALANCE.load(deps.storage)?;
     let staked_total = STAKED_TOTAL.load(deps.storage)?;
+    // invariant checks for amount_to_claim
     if staked_total.is_zero() {
         return Err(ContractError::NothingStaked {});
     }
     if amount.saturating_add(balance) == Uint128::MAX {
         return Err(ContractError::Cw20InvaraintViolation {});
+    }
+    if amount > staked_total {
+        return Err(ContractError::ImpossibleUnstake {});
     }
     let amount_to_claim = math::amount_to_claim(staked_total, balance, amount);
     STAKED_BALANCES.update(
--- a/contracts/staking/cw20-stake/src/contract.rs
+++ b/contracts/staking/cw20-stake/src/contract.rs
@@ -10,6 +10,7 @@ use cosmwasm_std::{
 use cw20::Cw20ReceiveMsg;
 
 use crate::hooks::{stake_hook_msgs, unstake_hook_msgs};
+use crate::math;
 use crate::msg::{
     ExecuteMsg, GetHooksResponse, InstantiateMsg, ListStakersResponse, MigrateMsg, QueryMsg,
     ReceiveMsg, StakedBalanceAtHeightResponse, StakedValueResponse, StakerBalanceResponse,
@@ -189,15 +190,7 @@ pub fn execute_stake(
 ) -> Result<Response, ContractError> {
     let balance = BALANCE.load(deps.storage)?;
     let staked_total = STAKED_TOTAL.load(deps.storage)?;
-    let amount_to_stake = if staked_total == Uint128::zero() || balance == Uint128::zero() {
-        amount
-    } else {
-        staked_total
-            .checked_mul(amount)
-            .map_err(StdError::overflow)?
-            .checked_div(balance)
-            .map_err(StdError::divide_by_zero)?
-    };
+    let amount_to_stake = math::amount_to_stake(staked_total, balance, amount);
     STAKED_BALANCES.update(
         deps.storage,
         &sender,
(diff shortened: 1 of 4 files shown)


JakeHartnell (approved):
LGTM.

contracts/staking/cw20-stake/src/contract.rs
```diff
@@ -233,11 +226,13 @@ pub fn execute_unstake(
...
     let staked_total = STAKED_TOTAL.load(deps.storage)?;
-    let amount_to_claim = amount
-        .checked_mul(balance)
-        .map_err(StdError::overflow)?
-        .checked_div(staked_total)
-        .map_err(StdError::divide_by_zero)?;
+    if staked_total.is_zero() {
+        return Err(ContractError::NothingStaked {});
+    }
+    if amount.saturating_add(balance) == Uint128::MAX {
```
blue-note, line 232: nice!

contracts/staking/cw20-stake/src/error.rs
```diff
@@ -9,6 +9,10 @@ pub enum ContractError {
     Cw20Error(#[from] cw20_base::ContractError),
     #[error("Nothing to claim")]
     NothingToClaim {},
+    #[error("Nothing to unstake")]
+    NothingStaked {},
+    #[error("Unstaking this amount violates the invaraint: (cw20 total_suppy <= 2^128).")]
```
blue-note, line 14: super nit spell check: invariant** and total_supply**

contracts/staking/cw20-stake/src/math.rs
```diff
@@ -0,0 +1,128 @@
...
+    fn test_amount_to_claim_invariant_one() {
+        let ask = Uint128::new(2);
+        let balance = Uint128::zero();
+
+        amount_to_claim(balance, balance, ask);
+    }
+
+    #[test]
+    #[should_panic(expected = "ConversionOverflowError")]
+    fn test_amount_to_claim_invaraint_two() {
```
blue-note, line 119: invariant** :)

contracts/staking/cw20-stake/src/math.rs
```diff
@@ -0,0 +1,128 @@
...
+pub(crate) fn amount_to_claim(staked_total: Uint128, balance: Uint128, ask: Uint128) -> Uint128 {
+    // we know that:
+    //
+    // 1. cw20's max supply is 2^128
+    // 2. balance := staked_total + rewards
+    //
+    // for non-malicious inputs:
+    //
+    // 3. 1 => ask + balance <= 2^128
+    // 4. ask <= staked_total
```
blue-note, line 57: should we also check the invariant for 4. above?

contracts/staking/cw20-stake/src/math.rs
```diff
@@ -0,0 +1,128 @@
...
+///
+/// For information on the panic conditions for math, see:
+/// <https://rust-lang.github.io/rfcs/0560-integer-overflow.html>
+pub(crate) fn amount_to_claim(staked_total: Uint128, balance: Uint128, ask: Uint128) -> Uint128 {
+    // we know that:
+    //
+    // 1. cw20's max supply is 2^128
+    // 2. balance := staked_total + rewards
+    //
+    // for non-malicious inputs:
```
blue-note, line 54: are we to assume that the invariant checks prevent all malicious inputs?

blue-note:
really excellent exercise in verifying correctness!!

contracts/staking/cw20-stake/src/math.rs
0xekez: yes! great catch. fixed

contracts/staking/cw20-stake/src/math.rs
0xekez: i think what i mean here is that we're putting bounds on what non-malicious input could be. anything outside of those bounds in a bad input, and we can error when we see it.

the rest of the proof is then showing that, for all possible valid inputs, this will not overflow.

contracts/staking/cw20-stake/src/math.rs
0xekez: i think the reason we didn't see this come up in our test coverage was because immediately below this function, a check was done that this was the case (to stop over withdrawing), and this invariant is overly broad in the sense that you pretty much need a `2^64` in one of those values for it to trigger.

blue-note: mm, so functionally it was fine

blue-note (approved):

PR #542, opened 2022-11-18
Add optional, updatable rationale to single choice votes.

### LINKED ISSUE #540: Allow associating a reason with a vote
When casting a vote, an address should be able to associate a reason. Later, that address can go back and add, remove, or update their reason. Response attributes should allow us to build a changelog of reasons over time using an indexer like Subquery.

0xekez:
resolves #540

i'll follow this up with an identical implementation for cwd-proposal-multiple in a separate PR.

what the vision is here: we're not really building comments, so much as we want to be clear that this is your rationale. having this is important for understanding someone's vote history when you are delegating to them. so that past votes may be annotated and rationales updated according to new information, we let rationales be updated.


JakeHartnell (approved):
Love this so much! Only thing I'm not sure about is the ability to update the rational... perhaps we make that optional?

0xekez:
> should there be an option for the admin to update/remove rationale?

this is a good question and gets to an important point about moderation of these. my thought: moderation is better done on the frontend by hiding rationales, removing it from chain state does not' remove it from the chain, and i don't think people want to do a DAO vote every time they remove a rationale.

in the future, i imagine that we'll allow DAOs to set an item which indicates what address, or set of addresses, may moderate rationales. then, we can use the [wallet auth system](https://github.com/NoahSaso/cloudflare-worker-cosmos-auth) we're working on to authenticate them, and allow them to moderate. imo, we should be trying to make stopping spam easier than creating it. by charging a TX fee to create spam, and making it effortless to remove it in the UI, i think we have a shot at winning the battle here.

0xekez:
> @JakeHartnell: I'm not sure about is the ability to update the rational

could you expand a little on what about that makes you nervous?

fwiw: i expect that we'll always display the history of rationales next to the current one. if you update it, your previous one will be shown. i've taken care in these changes to expose that data so that it is easily indexable and showable in the frontend.

JakeHartnell:
> > @JakeHartnell: I'm not sure about is the ability to update the rational
> 
> could you expand a little on what about that makes you nervous?
> 
> fwiw: i expect that we'll always display the history of rationales next to the current one. if you update it, your previous one will be shown. i've taken care in these changes to expose that data so that it is easily indexable and showable in the frontend.

On second thought, I think this is fine. :shipit: 

Can have an option to disable it in the UI if need be.

bekauz (approved):

PR #545, opened 2022-11-21
Begin private voting docs

ValarDragon:
Starting some documentation for how I suggest architecting private voting into the DAO DAO design.

This is very WIP. If I get more time available during the week, will write up more on circuit design / other constraints!

[Rendered readme](https://github.com/DA0-DA0/dao-contracts/blob/dev/private-voting/contracts/voting/private-voting/README.md) - gives higher level overview, and where its expected to sit in with other modules

[Rendered Spec](https://github.com/DA0-DA0/dao-contracts/blob/dev/private-voting/contracts/voting/private-voting/SPEC.md) - Intended to go into more details of cryptographic constructions, merkle trees involved, and client side UX. Still have a ways to go on this unfortunately

Code - Not started yet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2524,6 +2524,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "eb9f9e6e233e5c4a35559a617bf40a4ec447db2e84c20b55a6f83167b7e57872"
 
 [[package]]
+name = "private-voting"
+version = "0.1.0"
+dependencies = [
+ "cosmwasm-schema",
+ "cosmwasm-std",
+ "cosmwasm-storage",
+ "cw-multi-test",
+ "cw-storage-plus 0.16.0",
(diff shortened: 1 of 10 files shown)


0xekez:
this is really cool. i like your description of zcash and the idea here. it sounds reasonably intuitive, though i'm unfamiliar with zk proofs so i'm unable to verify it. :)

one note from reading through:

> So the voting power a shielded voter has, must either be frozen at proposal start time..

we do all of our voting power as a function of block height so that this is the case. from your earlier interface, this makes it:

```rust
pub trait VoteModule {
  fn voting_power_by_addr(&self, addr: Addr, height: u64) -> Option<Uint128>
}
```

allowing voting powers to decrease may be fine though, so long as we don't also allow increases.

JakeHartnell:
Privacy is insanely important, but closing this as stale.

PR #548, opened 2022-11-22
Update macros

### LINKED ISSUE #459: ProposalHookMsg variants should be derivable via a macro
In the `proposal-hooks` package we provide an enum with variants that must be implemented by hook receivers. We should provide a macro which will automatically add these to an existing execute message.

### LINKED ISSUE #509: Add "dao" query to voting_query macro
to mirror the proposal_module_query macro

JakeHartnell:
- Closes #509
- Closes #459 

Wound up removing the `info_query` macro as it was used for both proposal and voting modules. It's more and helpful IMO to only have to use one macro to implement a module.

Wound up also implementing a missing `Dao { }` query for the cw721 voting contract. This a good example of why it's nice to have everything in a `voting_query` macro... otherwise the developer has to remember which macros to use.

### DIFF
--- a/contracts/cwd-core/src/msg.rs
+++ b/contracts/cwd-core/src/msg.rs
@@ -4,7 +4,6 @@ use cosmwasm_schema::{cw_serde, QueryResponses};
 use cosmwasm_std::{CosmosMsg, Empty};
 use cw_utils::Duration;
 use cwd_interface::ModuleInstantiateInfo;
-use cwd_macros::{info_query, voting_query};
 
 /// Information about an item to be stored in the items list.
 #[cw_serde]
@@ -39,10 +38,9 @@ pub struct InstantiateMsg {
     /// Instantiate information for the core contract's voting
     /// power module.
     pub voting_module_instantiate_info: ModuleInstantiateInfo,
-    /// Instantiate information for the core contract's
-    /// proposal modules.
-    // NOTE: the pre-propose-base package depends on it being the case
-    // that the core module instantiates its proposal module.
+    /// Instantiate information for the core contract's proposal modules.
+    /// NOTE: the pre-propose-base package depends on it being the case
+    /// that the core module instantiates its proposal module.
     pub proposal_modules_instantiate_info: Vec<ModuleInstantiateInfo>,
 
     /// Initial information for arbitrary contract addresses to be
@@ -50,7 +48,7 @@ pub struct InstantiateMsg {
     /// items map. The value is an enum that either uses an existing
     /// address or instantiates a new contract.
     pub initial_items: Option<Vec<InitialItem>>,
-    /// Implements the DAO Star standard: https://daostar.one/EIP
+    /// Implements the DAO Star standard: <https://daostar.one/EIP>
     pub dao_uri: Option<String>,
 }
 
@@ -117,8 +115,8 @@ pub enum ExecuteMsg {
     /// instantiate info in `to_add` is used to create new modules and
     /// install them.
     UpdateProposalModules {
-        // NOTE: the pre-propose-base package depends on it being the
-        // case that the core module instantiates its proposal module.
+        /// NOTE: the pre-propose-base package depends on it being the
+        /// case that the core module instantiates its proposal module.
         to_add: Vec<ModuleInstantiateInfo>,
         to_disable: Vec<String>,
     },
@@ -133,8 +131,6 @@ pub enum ExecuteMsg {
     },
 }
 
-#[voting_query]
-#[info_query]
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
@@ -185,6 +181,9 @@ pub enum QueryMsg {
         start_after: Option<String>,
         limit: Option<u32>,
     },
+    /// Returns contract version info
+    #[returns(cwd_interface::voting::InfoResponse)]
+    Info {},
     /// Gets all proposal modules associated with the
     /// contract.
     #[returns(Vec<crate::state::ProposalModule>)]
@@ -212,9 +211,18 @@ pub enum QueryMsg {
         start_after: Option<String>,
         limit: Option<u32>,
     },
-    /// Implements the DAO Star standard: https://daostar.one/EIP
+    /// Implements the DAO Star standard: <https://daostar.one/EIP>
     #[returns(crate::query::DaoURIResponse)]
     DaoURI {},
+    /// Returns the voting power for an address at a given height.
+    #[returns(cwd_interface::voting::VotingPowerAtHeightResponse)]
+    VotingPowerAtHeight {
+        address: String,
+        height: Option<u64>,
+    },
+    /// Returns the total voting power at a given block height.
+    #[returns(cwd_interface::voting::TotalPowerAtHeightResponse)]
+    TotalPowerAtHeight { height: Option<u64> },
 }
 
 #[cw_serde]
--- a/contracts/proposal/cwd-proposal-multiple/src/msg.rs
+++ b/contracts/proposal/cwd-proposal-multiple/src/msg.rs
@@ -1,6 +1,6 @@
 use cosmwasm_schema::{cw_serde, QueryResponses};
 use cw_utils::Duration;
-use cwd_macros::{info_query, proposal_module_query};
+use cwd_macros::proposal_module_query;
 use cwd_voting::{
     multiple_choice::{MultipleChoiceOptions, MultipleChoiceVote, VotingStrategy},
     pre_propose::PreProposeInfo,
@@ -133,7 +133,6 @@ pub enum ExecuteMsg {
 }
 
 #[proposal_module_query]
-#[info_query]
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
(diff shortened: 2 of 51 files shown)


contracts/voting/cwd-voting-cw721-staked/src/state.rs
```diff
@@ -15,6 +15,7 @@ pub struct Config {
 }
 
 pub const CONFIG: Item<Config> = Item::new("config");
+pub const DAO_ADDRESS: Item<Addr> = Item::new("dao_address");
```
blue-note, line 18: nit: in other places we name this `dao`

packages/cwd-macros/src/lib.rs
```diff
@@ -109,6 +111,7 @@ fn cwd_interface_path(inside: &str) -> Path {
 /// ```
 #[proc_macro_attribute]
 pub fn voting_query(metadata: TokenStream, input: TokenStream) -> TokenStream {
```
blue-note, line 113: naming nit: `voting_module_query` to match `proposal_module_query`

contracts/cwd-core/src/msg.rs
```diff
@@ -212,9 +211,18 @@ pub enum QueryMsg {
         start_after: Option<String>,
         limit: Option<u32>,
     },
-    /// Implements the DAO Star standard: https://daostar.one/EIP
+    /// Implements the DAO Star standard: <https://daostar.one/EIP>
     #[returns(crate::query::DaoURIResponse)]
     DaoURI {},
+    /// Returns the voting power for an address at a given height.
+    #[returns(cwd_interface::voting::VotingPowerAtHeightResponse)]
+    VotingPowerAtHeight {
```
blue-note, line 219: hmm maybe we could have a `core_module_query` macro as well? feels a little odd to have some modules have these as macros and some not

blue-note:
thanks for cleaning this up! I agree that the number of macros was getting confusing

contracts/cwd-core/src/msg.rs
```diff
@@ -39,18 +38,17 @@ pub struct InstantiateMsg {
     /// Instantiate information for the core contract's voting
     /// power module.
     pub voting_module_instantiate_info: ModuleInstantiateInfo,
-    /// Instantiate information for the core contract's
-    /// proposal modules.
-    // NOTE: the pre-propose-base package depends on it being the case
-    // that the core module instantiates its proposal module.
+    /// Instantiate information for the core contract's proposal modules.
+    /// NOTE: the pre-propose-base package depends on it being the case
+    /// that the core module instantiates its proposal module.
```
0xekez, line 43: this is actually an intentional thing where i try and separate "comments for a developer" from "comments for someone interacting with this contract". wdyt? if you find it confusing, then it probably is

contracts/proposal/cwd-proposal-multiple/src/msg.rs
```diff
@@ -157,32 +155,16 @@ pub enum QueryMsg {
...
-    /// module.
-    #[returns(cwd_voting::pre_propose::ProposalCreationPolicy)]
-    ProposalCreationPolicy {},
-    /// Lists all of the consumers of proposal hooks for this module.
-    #[returns(cwd_hooks::HooksResponse)]
-    ProposalHooks {},
-    /// Lists all of the consumers of vote hooks for this
-    /// module.
-    #[returns(cwd_hooks::HooksResponse)]
-    VoteHooks {},
```
0xekez, line 185: great refactor to make this part of the interface

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
```diff
@@ -45,6 +45,8 @@ pub fn instantiate(
 
     TOTAL_STAKED_NFTS.save(deps.storage, &Uint128::zero(), env.block.height)?;
 
+    DAO_ADDRESS.save(deps.storage, &info.sender)?;
```
0xekez, line 48: how about moving this above the `let owner = // ...` expression so that we can avoid the clone?

contracts/cwd-core/src/msg.rs
0xekez: we totally could just return `env.contract.address` from a `Dao {}` query and be voting module interface compliant.

con: weird looking
pro: consistent, modules can actually treat this like a full blown voting module.

wdyt?

0xekez (commented):
"yay" :) great refactor. love the moving of proposal creation policy and hooks into the proposal interface macro. super important for making sure those modules are compatible with pre-propose modules.

couple little things that i think would be nice to clean up.

contracts/cwd-core/src/msg.rs
JakeHartnell: I thought the core module is unique in that there would likely only be one implementation verses voting or proposal modules where folks will constantly be writing custom ones.

For the sake of clarity, I think it's fine to not use macros here. :shrug:

contracts/cwd-core/src/msg.rs
JakeHartnell: I think these comments are relevant to show in generated code documentation (e.g. docs.rs), which _is_ for developers. : )

contracts/cwd-core/src/msg.rs
JakeHartnell: > we totally could just return `env.contract.address` from a `Dao {}` query and be voting module interface compliant.
> 
> con: weird looking pro: consistent, modules can actually treat this like a full blown voting module.
> 
> wdyt?

Don't have strong opinions here... do think it's a little weird looking... could do that though. Might be cool to use a DAO for voting power... though I think we could do this without the `Dao { }` query?

contracts/voting/cwd-voting-cw721-staked/src/contract.rs
JakeHartnell: Done.

packages/cwd-macros/src/lib.rs
JakeHartnell: Nice, updated.

contracts/voting/cwd-voting-cw721-staked/src/state.rs
JakeHartnell: Made it consistent.

contracts/cwd-core/src/msg.rs
JakeHartnell: > we totally could just return `env.contract.address` from a `Dao {}` query and be voting module interface compliant.
> 
> con: weird looking pro: consistent, modules can actually treat this like a full blown voting module.
> 
> wdyt?

Don't think we want this since it won't behave correctly. The `Dao { }` query is supposed to return the DAO the voting module belongs to. If it's hard coded to `env.contract.address` and another DAO was using that DAO as a voting module, it would return the incorrect value. Not sure we want to encourage this anyway. :joy:

Leaving as is.

0xekez: > and another DAO was using that DAO as a voting module

this is a genius point

contracts/voting/cwd-voting-cw4/src/state.rs
```diff
@@ -16,4 +16,4 @@ pub const TOTAL_WEIGHT: SnapshotItem<Uint128> = SnapshotItem::new(
 );
 
 pub const GROUP_CONTRACT: Item<Addr> = Item::new("group_contract");
-pub const DAO_ADDRESS: Item<Addr> = Item::new("dao_address");
+pub const DAO: Item<Addr> = Item::new("dao");
```
0xekez, line 19: we actually need to leave the key the same here. otherwise, when we upgrade this contract the query will stop working (it'll look up the wrong key). could either add a migrate method to this, or add something along these lines to a comment and keep the key as `dao_address`.

```suggestion
pub const DAO: Item<Addr> = Item::new("dao_address");
```

packages/cwd-macros/src/lib.rs
```diff
@@ -355,13 +310,31 @@ pub fn info_query(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
+            #[returns(u64)]
+            ProposalCount {},
+            /// Gets the current proposal creation policy for this module.
+            #[returns(cwd_voting::pre_propose::ProposalCreationPolicy)]
+            ProposalCreationPolicy {},
+            /// Lists all of the consumers of proposal hooks for this module.
+            #[returns(cwd_hooks::HooksResponse)]
+            ProposalHooks {},
+            /// Lists all of the consumers of vote hooks for thismodule.
+            #[returns(cwd_hooks::HooksResponse)]
```
0xekez, line 336:
```suggestion
            #[returns(::std::primitive::u64)]
            ProposalCount {},
            /// Gets the current proposal creation policy for this module.
            #[returns(::cwd_voting::pre_propose::ProposalCreationPolicy)]
            ProposalCreationPolicy {},
            /// Lists all of the consumers of proposal hooks for this module.
            #[returns(::cwd_hooks::HooksResponse)]
            ProposalHooks {},
            /// Lists all of the consumers of vote hooks for this module.
            #[returns(::cwd_hooks::HooksResponse)]
```

know it's a little verbose, but nice to keep our macros hygienic. unlikely to ever be a problem, but when it is it's pretty hard to debug.

contracts/voting/cwd-voting-cw4/src/state.rs
JakeHartnell: HOLY SHIT you are right... this would not have been fun.

0xekez (approved):

blue-note (approved):

PR #552, opened 2022-11-22
Payroll contract supporting new operations with streams

### LINKED ISSUE #445: payroll contract

deveusss:
closes #445


contracts/external/cw-payroll/src/balance.rs
```diff
@@ -0,0 +1,185 @@
...
+use cosmwasm_schema::cw_serde;
+
+use crate::error::GenericError;
+
+#[cw_serde]
+#[derive(Default)]
+pub struct GenericBalance {
+    pub native: Vec<Coin>,
+    pub cw20: Vec<Cw20CoinVerified>,
+}
```
0xekez, line 12: 1. vectors aren't a great data structure for indexing based on keys.
2. every time we load this from storage, we pay gas for all of the data in it. pretty trivial to DOS the contract, this being the case. check out "attempt two" in this essay for an example of that: https://github.com/DA0-DA0/dao-contracts/discussions/523

how about we move to using cosmwasm maps for these? this might be a nice example to work from: https://github.com/public-awesome/ics721/blob/main/packages/cw-pause-once/src/lib.rs#L23-L26

deveusss: > 1. vectors aren't a great data structure for indexing based on keys.
> 2. every time we load this from storage, we pay gas for all of the data in it. pretty trivial to DOS the contract, this being the case. check out "attempt two" in this essay for an example of that: [The three horsemen of NFT staking and voting power #523](https://github.com/DA0-DA0/dao-contracts/discussions/523)
> 
> how about we move to using cosmwasm maps for these? this might be a nice example to work from: https://github.com/public-awesome/ics721/blob/main/packages/cw-pause-once/src/lib.rs#L23-L26

yeah, agree with you, just we considering not holding much data in the vectors, but still you are right, we need to modify this ones)

0xekez: > that seem to create their own map?

that may have been a confusing explanation :) they don't actually make a `Map`. just trying to suggest that we refactor this to look like that type. for example:

```rust
pub struct GenericBalance {
    pub native: Map<String, Uint128>,
    pub cw20: Vec<String, Uint128>,
}
```

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -1,17 +1,24 @@
...
-    Uint128,
+    from_binary, to_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut, Env,
+    MessageInfo, Order, Response, StdResult, Uint128,
 };
 use cw2::set_contract_version;
-use cw20::{Cw20Contract, Cw20ExecuteMsg, Cw20ReceiveMsg};
+use cw20::{Balance, Cw20CoinVerified, Cw20Contract, Cw20ExecuteMsg, Cw20ReceiveMsg};
 use cw_storage_plus::Bound;
 
 const CONTRACT_NAME: &str = "crates.io:cw20-streams";
```
TrevorJTClarke, line 24: Let's change this to `crates.io:cw-escrow-streams`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -53,13 +60,105 @@ pub fn execute(
...
+    }
+    stream.paused_time = Some(env.block.time.seconds());
+    stream.paused = true;
+    save_stream(deps, id, &stream).unwrap();
+    let response = Response::new()
+        .add_attribute("method", "pause_stream")
+        .add_attribute("paused", stream.paused.to_string())
+        .add_attribute("stream_id", id.to_string())
+        .add_attribute("admin", info.sender)
+        .add_attribute("paused_time", env.block.time.to_string());
```
TrevorJTClarke, line 96: I believe this should actually use same variable: `stream.paused_time`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -53,13 +60,105 @@ pub fn execute(
...
+    if stream.admin != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    remove_stream(deps, id).unwrap();
+    let response = Response::new()
+        .add_attribute("method", "remove_stream")
+        .add_attribute("stream_id", id.to_string())
+        .add_attribute("admin", info.sender)
+        .add_attribute("paused_time", env.block.time.to_string());
```
TrevorJTClarke, line 119: change `paused_time` to `removed_time`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -53,13 +60,105 @@ pub fn execute(
...
+        .may_load(deps.storage, id)?
+        .ok_or(ContractError::StreamNotFound {})?;
+
+    if stream.admin != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
+    if !stream.paused {
+        return Err(ContractError::StreamNotPaused {});
+    }
+    let duration: u128 = (stream.end_time - stream.start_time).into();
```
TrevorJTClarke, line 141: Make sure not `0`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -53,13 +60,105 @@ pub fn execute(
...
+    let paused_duration: u128 = (env.block.time.seconds() - stream.paused_time.unwrap()).into();
+    let rate_per_second = Uint128::from(stream.balance.total_amount() / duration - paused_duration);
```
TrevorJTClarke, line 143: Make sure no over/under flows, and not divide by `0`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -53,13 +60,105 @@ pub fn execute(
...
+    stream.paused = false;
+    stream.rate_per_second = rate_per_second;
```
TrevorJTClarke, line 145: See comment in msg or state -- I don't think we should store `rate_per_second`, can always be computed on request.
So, remove `stream.rate_per_second` but you can leave inside response attribute

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -53,13 +60,105 @@ pub fn execute(
...
+        .add_attribute("paused_time", stream.paused_time.unwrap().to_string())
+        .add_attribute("resume_time", env.block.time.to_string());
+
+    Ok(response)
+}
 pub fn execute_create_stream(
     env: Env,
     deps: DepsMut,
-    config: Config,
+    _info: MessageInfo,
```
TrevorJTClarke, line 161: can be removed?

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -80,59 +182,96 @@ pub fn execute_create_stream(
...
+    // UPD: Fixed
+    if start_time <= block_time && end_time <= block_time {
         return Err(ContractError::InvalidStartTime {});
     }
 
-    let duration: Uint128 = (end_time - start_time).into();
-
-    if balance < duration {
-        return Err(ContractError::AmountLessThanDuration {});
+    if end_time < block_time {
```
TrevorJTClarke, line 193: Change this to be `end_time <= block_time`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -80,59 +182,96 @@ pub fn execute_create_stream(
 
     let block_time = env.block.time.seconds();
     // TODO: Change to support previous dates, within given reason?
-    //       The problem with this is it restricts backdating, 
+    //       The problem with this is it restricts backdating,
     //       so all funds must vest before distribute (for good or bad)
     // TODO: start_time Could default to now?
-    if start_time <= block_time {
+    // UPD: Fixed
+    if start_time <= block_time && end_time <= block_time {
```
TrevorJTClarke, line 189: remove this if block. start time could be `0`, because we want to allow starting to occur in past, if say compensation was started in july 2022 but the stream was created in november 2022 and ends in july 2023.

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -80,59 +182,96 @@ pub fn execute_create_stream(
...
     let stream = Stream {
-        admin: admin.clone(),
+        admin: owner.clone(),
         recipient: recipient.clone(),
         balance,
-        claimed_balance: Uint128::zero(),
+        claimed_balance: GenericBalance::default(),
         start_time,
         end_time,
         rate_per_second,
```
TrevorJTClarke, line 257: remove this

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -70,7 +169,10 @@ pub fn execute_create_stream(
         end_time,
         title,
         description,
+        paused_time,
+        paused,
     } = params;
+
     let owner = deps.api.addr_validate(&admin)?;
```
TrevorJTClarke, line 176: let's stay consistent - use `admin` instead

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -80,59 +182,96 @@ pub fn execute_create_stream(
...
+                })?;
+                msgs.push(msg);
+            }
+        }
+    }
+    let rate_per_second = Uint128::from(total_amount.u128() / duration);
 
     let stream = Stream {
-        admin: admin.clone(),
+        admin: owner.clone(),
```
TrevorJTClarke, line 251: let's stay consistent - use `admin` instead

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -144,12 +283,14 @@ pub fn execute_receive(
     info: MessageInfo,
     wrapped: Cw20ReceiveMsg,
 ) -> Result<Response, ContractError> {
-    let config = CONFIG.load(deps.storage)?;
+    let sender = deps.api.addr_validate(&info.sender.clone().into_string())?;
```
TrevorJTClarke, line 286: Are we 100% this sender will be the appropriate cw20 contract and not just anyone passing variables?

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -188,44 +331,101 @@ pub fn execute_distribute(
...
-    //         recipient: stream.recipient,
-    //     });
-    // }
 
     // TODO: Def change to better comparitor
-    if stream.claimed_balance >= stream.balance {
-        return Err(ContractError::StreamFullyClaimed {});
+    //Fixed
+
+    if stream.recipient != info.sender {
```
TrevorJTClarke, line 338: let's remove this - we want external parties to claim / distribute.

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -188,44 +331,101 @@ pub fn execute_distribute(
...
-    let config = CONFIG.load(deps.storage)?;
-    let cw20 = Cw20Contract(config.cw20_addr);
-    let msg = cw20.call(Cw20ExecuteMsg::Transfer {
-        recipient: stream.recipient.clone().into(),
-        amount: released,
-    })?;
-
-    let res = Response::new()
+    let mut res = Response::new()
         .add_attribute("method", "withdraw")
```
TrevorJTClarke, line 424: change to `distribute`

contracts/external/cw-payroll/src/error.rs
```diff
@@ -34,4 +56,10 @@ pub enum ContractError {
 
     #[error("Numerical overflow")]
```
TrevorJTClarke, line 57: This is a GenericError ;)

contracts/external/cw-payroll/src/msg.rs
```diff
@@ -1,41 +1,43 @@
-use crate::state::GenericBalance;
+use crate::balance::GenericBalance;
 use cosmwasm_schema::{cw_serde, QueryResponses};
 use cosmwasm_std::Uint128;
-use cw20::Cw20ReceiveMsg;
+use cw20::{Cw20ReceiveMsg, Balance};
```
TrevorJTClarke, line 4: remove `Balance` unless used in this file

contracts/external/cw-payroll/src/balance.rs
TrevorJTClarke: After discussion - Let's roll with single token amounts here.
As @0xekez pointed out, a lot of fun edge cases open up with multiple tokens.
Since you can create multiple "streams", we can then force 1 stream to be holding (1 native asset, 1 cw20 asset). 
If there are real use cases to go beyond this, it can be a future upgrade with adequate research & test cov

TrevorJTClarke (changes requested):
Hey @deveusss! 
Lot's of great work here! Threw a ton of nit-picks in here for moving closer to merging into other PR.

Things I didn't review closely:
1. Balance/token stuff - since we discussed changing to single token holdings, will review in depth upon that change.
2. Test coverage - since you said you were deep into adding more and cleaning up. Will review on next pass.

Stoked on this!!!
Cheers!

contracts/external/cw-payroll/src/contract.rs
deveusss: @TrevorJTClarke remove claimed_balance?

contracts/external/cw-payroll/src/contract.rs
deveusss: But when pause happened multiple times, we should recalculate all this pauses, so we need store this value

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -161,71 +277,79 @@ pub fn execute_receive(
...
-    //     });
-    // }
-
-    // TODO: Def change to better comparitor
-    if stream.claimed_balance >= stream.balance {
-        return Err(ContractError::StreamFullyClaimed {});
+    let mut msgs: Vec<CosmosMsg> = vec![];
+    let (available_claims, _) = stream.calc_distribution_rate(env.block.time);
+
+    if !stream.can_ditribute_more() || available_claims.u128() <= 0 {
```
TrevorJTClarke, line 301: change `can_ditribute_more` to `can_distribute_more`

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -161,71 +277,79 @@ pub fn execute_receive(
...
-    let released = vested - stream.claimed_balance;
+        stream
+            .claimed_balance
+            .checked_add_native(&[to_claim.clone()])
+            .unwrap();
+        let msg = CosmosMsg::Bank(BankMsg::Send {
+            to_address: stream.recipient.clone().into(),
+            amount: vec![Coin {
+                denom: native_balance.denom.clone(),
+                amount: to_claim.amount,
```
TrevorJTClarke, line 320: I believe this value to be wrong @deveusss , but please double check my thinking.

You are appropriately checking above for claimed balance & current stream balance, so dont change those.

On this line, `to_claim.amount` doesnt take into account the amount already claimed. So if i hit the `distribute` method multiple times, i can immediately drain the full stream i think.
Line 299 with `available_claims` needs to subtract `stream.claimed_balance` to be fully accurate on line 309.

If im wrong, please help me understand :)
Would also love test coverage for this re-distribute multiple times in a row case.

TrevorJTClarke: Also, noting this is the case for cw20 impl on line 328

contracts/external/cw-payroll/src/msg.rs
```diff
@@ -1,41 +1,32 @@
...
+        id: StreamId, // Stream id
     },
-    // TODO: Add this! :D
-    // TODO: Only called by stream admin
-    // NOTE: Removes any pause from a stream, 
-    //       optionally can shift the start/end time if needed for vesting flexibility
     ResumeStream {
-        id: u64, // Stream id
+        id: StreamId, // Stream id
         start_time: Option<u64>,
```
TrevorJTClarke, line 24: Let's remove `start_time` & `end_time` here. I can think of ways that the admin could be mean for shifting agreed upon params. Better to just have the stream get removed & recreated than cover edge cases on this.

contracts/external/cw-payroll/src/state.rs
```diff
@@ -1,87 +1,109 @@
...
-
     // METADATA
     /// Title of the payroll item, for example for a bug bounty "Fix issue in contract.rs"
     pub title: Option<String>,
     /// Description of the payroll item, a more in depth description of how to meet the payroll conditions
     pub description: Option<String>,
 }
 
+impl Stream {
+    pub(crate) fn can_ditribute_more(&self) -> bool {
```
TrevorJTClarke, line 36: change `can_ditribute_more ` to `can_distribute_more `

PR #554, opened 2022-11-24
Add `NextProposalId` query to proposal module interface.

0xekez:
Doing so pretty significantly simplifies pre-propose modules, and makes less compex to write as we can remove the proposal created hook entirely as part of this change.

This also walks back the addition of pre-propose and hooks to the proposal module interface. I think there is beauty in smallness there, and if someone else implements a proposal module they may not want to use our hook and proposal-gating types.

### DIFF
--- a/contracts/proposal/cwd-proposal-multiple/src/msg.rs
+++ b/contracts/proposal/cwd-proposal-multiple/src/msg.rs
@@ -165,6 +165,18 @@ pub enum QueryMsg {
         start_after: Option<String>,
         limit: Option<u64>,
     },
+    /// Returns the number of proposals that have been created in this module.
+    #[returns(::std::primitive::u64)]
+    ProposalCount {},
+    /// Gets the current proposal creation policy for this module.
+    #[returns(::cwd_voting::pre_propose::ProposalCreationPolicy)]
+    ProposalCreationPolicy {},
+    /// Lists all of the consumers of proposal hooks for this module.
+    #[returns(::cwd_hooks::HooksResponse)]
+    ProposalHooks {},
+    /// Lists all of the consumers of vote hooks for this module.
+    #[returns(::cwd_hooks::HooksResponse)]
+    VoteHooks {},
 }
 
 #[cw_serde]
--- a/contracts/proposal/cwd-proposal-single/src/msg.rs
+++ b/contracts/proposal/cwd-proposal-single/src/msg.rs
@@ -202,6 +202,18 @@ pub enum QueryMsg {
         /// query. If no limit is specified a max of 30 are returned.
         limit: Option<u64>,
     },
+    /// Returns the number of proposals that have been created in this module.
+    #[returns(::std::primitive::u64)]
+    ProposalCount {},
+    /// Gets the current proposal creation policy for this module.
+    #[returns(::cwd_voting::pre_propose::ProposalCreationPolicy)]
+    ProposalCreationPolicy {},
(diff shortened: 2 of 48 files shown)


contracts/proposal/cwd-proposal-multiple/src/msg.rs
```diff
@@ -165,6 +165,18 @@ pub enum QueryMsg {
...
+    ProposalCount {},
+    /// Gets the current proposal creation policy for this module.
+    #[returns(::cwd_voting::pre_propose::ProposalCreationPolicy)]
+    ProposalCreationPolicy {},
+    /// Lists all of the consumers of proposal hooks for this module.
+    #[returns(::cwd_hooks::HooksResponse)]
+    ProposalHooks {},
+    /// Lists all of the consumers of vote hooks for this module.
+    #[returns(::cwd_hooks::HooksResponse)]
+    VoteHooks {},
```
JakeHartnell, line 179: Thought these were implemented by the `proposal_module_query` macro?

contracts/proposal/cwd-proposal-single/src/msg.rs
```diff
@@ -202,6 +202,18 @@ pub enum QueryMsg {
...
+    ProposalCount {},
+    /// Gets the current proposal creation policy for this module.
+    #[returns(::cwd_voting::pre_propose::ProposalCreationPolicy)]
+    ProposalCreationPolicy {},
+    /// Lists all of the consumers of proposal hooks for this module.
+    #[returns(::cwd_hooks::HooksResponse)]
+    ProposalHooks {},
+    /// Lists all of the consumers of vote hooks for this module.
+    #[returns(::cwd_hooks::HooksResponse)]
+    VoteHooks {},
```
JakeHartnell, line 216: Also, thought these would be implemented by the `proposal_module_query` macro... :joy:

packages/cwd-macros/src/lib.rs
```diff
@@ -40,6 +40,12 @@ fn merge_variants(metadata: TokenStream, left: TokenStream, right: TokenStream)
 /// cwd_interface. If we are currently compiling the cwd-interface
 /// crate, `crate::{internal}` is returned. If we are not,
 /// `::cwd_interface::{internal}` is returned.
+///
+/// The this is needed is that cwd_interface both defines types used
```
JakeHartnell, line 44:
```suggestion
/// The reason this is needed is that cwd_interface both defines types used
```

JakeHartnell: Starting to wonder how useful this macro is if it doesn't implement much.

contracts/proposal/cwd-proposal-single/src/msg.rs
JakeHartnell: Ah, I see the other code changes and the note, nevermind.

contracts/proposal/cwd-proposal-multiple/src/msg.rs
JakeHartnell: I understand now. Resolving...

JakeHartnell (approved):
Love this! Greatly simplifies things.

onewhiskeypls (approved):
**utACK**: lgtm!

side thought: @0xekez I've been thinking about this and can't find it entirely helpful, but wdyt about prefixing the cwd-macro names to better distinguish them from the standard cw and rust decorators?

PR #557, opened 2022-11-26
Add `proposer` field to `PendingProposal`.

0xekez:
This is slightly less space efficent than the way it was, but I feel we can sacrifice space efficency a little for safer typing.


JakeHartnell (approved):
Nice improvement.

PR #562, opened 2022-11-26
Cwd proposal delegate

baoskee:
Need design review before I start writing tests! : ) 
Low prio, no hurries. Thought would be fun first contract to write for DaoDao.


contracts/proposal/cwd-proposal-delegate/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+backtraces = ["cosmwasm-std/backtraces"]
+# use library feature to disable all instantiate/execute/query exports
+library = []
+
+[package.metadata.scripts]
+optimize = """docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  cosmwasm/rust-optimizer:0.12.6
+"""
```
0xekez, line 40: because this needs to be built as part of the workspace (you have a cwd-core dep) this won't work / we don't need it
```suggestion
```

contracts/proposal/cwd-proposal-delegate/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+
+[dependencies]
+cosmwasm-schema = "1.1.3"
+cosmwasm-std = "1.1.3"
+cosmwasm-storage = "1.1.3"
```
0xekez, line 45: almost no contracts depend on this, yet it is part of the default template lol
```suggestion
```

contracts/proposal/cwd-proposal-delegate/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+library = []
+
+[package.metadata.scripts]
+optimize = """docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  cosmwasm/rust-optimizer:0.12.6
+"""
+
+[dependencies]
```
0xekez, line 42: all of these can be `cosmwasm-x = { workspace = true } `

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractError> {
+    match msg.id {
+        REPLY_ID_EXECUTE_PROPOSAL_HOOK => {
+            let id = EXECUTE_CTX.load(deps.storage)?;
+            let Delegation {
+                policy_preserve_on_failure,
```
0xekez, line 86: this is sick

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+                ..
+            } = DELEGATIONS.load(deps.storage, id)?;
+
+            if let SubMsgResult::Err(_) = msg.result {
+                // && with `let` expression above gives unstable Rust error
+                if policy_preserve_on_failure {
```
0xekez, line 92:
```suggestion
            if msg.result.is_err() && policy_preserve_on_failure {
```

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+                    return Ok(Response::default()
+                        .add_attribute("execute_failed_but_preserved", id.to_string()));
+                }
+            }
+            // Delete delegation in both success and error case
+            DELEGATIONS.remove(deps.storage, id);
+            Ok(Response::default())
```
0xekez, line 99: nit: nice to have this in an `else` block because it makes it clear from reading that we'll never "fall out" of the if statement above and into this.

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+
+fn query_delegation_count(deps: Deps) -> StdResult<DelegationCountResponse> {
+    let count = DELEGATION_COUNT.load(deps.storage)?;
+    Ok(DelegationCountResponse { count })
+}
+
+fn query_delegation(deps: Deps, delegation_id: u64) -> StdResult<DelegationResponse> {
+    let delegation = DELEGATIONS.load(deps.storage, delegation_id)?;
+    Ok(delegation)
+}
```
0xekez, line 127: probably want to make this a paginated function so that FE can list all of the delegations pending. `packages/cw-paginate` should make that simple

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+}
+
+// MARK: Query helpers
+
+fn query_delegation_count(deps: Deps) -> StdResult<DelegationCountResponse> {
+    let count = DELEGATION_COUNT.load(deps.storage)?;
+    Ok(DelegationCountResponse { count })
+}
```
0xekez, line 122: idea: why expose this at all? all we use it for is a nonce for delegations. let's make it a private state variable and only expose a function for incrementing it. gets us pretty close to correct-by-constructuon!

ex:

```rust
pub(crate) fn advance_current_id(store: &mut dyn Storage) -> StdResult<u64> {
    let CURRENT_ID: Item<u64> = Item::new("current_id");
    let id: u64 = CURRENT_ID.may_load(store)?.unwrap_or_default() + 1;
    CURRENT_ID.save(store, &id)?;
    Ok(id)
}
```

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    delegation: Delegation,
+) -> Result<Response, ContractError> {
+    let Config { admin } = CONFIG.load(deps.storage)?;
+    if info.sender != admin {
+        return Err(ContractError::Unauthorized {});
+    }
+    assert_not_expired(&delegation.expiration, &env.block)?;
```
0xekez, line 141: nice check

contracts/proposal/cwd-proposal-delegate/src/contract.rs
```diff
@@ -0,0 +1,238 @@
...
+    info: MessageInfo,
+    delegation_id: u64,
+) -> Result<Response, ContractError> {
+    let Config { admin } = CONFIG.load(deps.storage)?;
+    if info.sender != admin {
+        return Err(ContractError::Unauthorized {});
+    }
+    // If delegation is irrevocable, return Error
+    let Delegation {
+        policy_irrevocable, ..
```
0xekez, line 161: what's the reason for having this option? i feel things like this can give fake guarantees if the contract is instantiated with a DAO as it's admin. should we be instantiating this without an admin perhap?

contracts/proposal/cwd-proposal-delegate/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("Unauthorized")]
+    Unauthorized {},
+    // Add any other custom errors you like here.
+    // Look at https://docs.rs/thiserror/1.0.21/thiserror/ for details.
```
0xekez, line 12:
```suggestion
```

0xekez (commented):
🫡 cool! super clean. let's land this :)

contracts/proposal/cwd-proposal-delegate/src/contract.rs
baoskee: cool!

contracts/proposal/cwd-proposal-delegate/src/contract.rs
baoskee: good point. The DAO could always remove the proposal module and all delegations would be revocable. However, that would eliminate all their existing delegations as well. There's 2 options to go with this:
1. Rename the policy to `policy_proposal_irrevocable` and add a warning
2. Remove the policy altogether

baoskee:
Will integrate changes and start on the tests! A bit slow on this, been grokking some Cosmos stuff. Need to learn Go, how dreadful

baoskee:
Ok, last commits were:
- fixed Cargo.toml
- made changes for PR
- added pagination query (can remove the single delegation query if that's cleaner)
- made delegation count private
- renamed policy_revocable to `policy_module_irrevocable`. I think that having revocability at the module level is convenient even though it can expose user error (because DAO can remove the module to "revoke" the delegation). For example, the Constitution can technically disband the Supreme Court to prevent a particular judge-made decision to not be implemented (say McCulloch v. Maryland), but that would require high costs plus the dismantling of every other Court decision
- added a bunch of unit tests covering the risks I laid out in README.md 

Lmk.

JakeHartnell:
Closing as stale.

PR #565, opened 2022-11-27
Better token swap ux

JakeHartnell:
Inspired by @NoahSaso: it's better UX for the token swap to be able to instantiate the token swap contract and provide the promised amount in one TX.

This only works for native tokens for now, but Juno will be moving to token factory soon, so that's fine IMO. Could potentially add a factory later if we really want, but IMO... this is good enough.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,9 +48,9 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
 name = "async-io"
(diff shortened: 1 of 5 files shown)


JakeHartnell:
@0xekez from Discord:
>  my five cents here: we can just instantiate the contract out-of-band (button in the action) and then fund it with the action. there's no need to have the DAO instantiate the contract as part of a proposal so long as folks can verify the bytecode

I think this is a good point, as long as the contract doesn't have an admin and can be verified, then all is good. Closing.

PR #568, opened 2022-12-05
Streams linking and synchronisation

deveusss:


contracts/external/cw-payroll/src/error.rs
```diff
@@ -61,9 +61,22 @@ pub enum ContractError {
...
+
+    #[error("Linked stream does not exist.")]
+    LinkedStreamNotFound { stream_id: StreamId },
+
+    #[error("Stream is not linked.")]
+    StreamNotLinked {},
+    #[error("Stream is not an initiator.")]
+    StreamNotInitiator {},
+    #[error("Stream is not detachable.")]
+    StreamNotDetachable {},
```
JakeHartnell, line 81: Nit, consistent spacing between error messages please.

contracts/external/cw-payroll/src/msg.rs
```diff
@@ -82,6 +86,12 @@ pub struct StreamResponse {
     pub paused: bool,
     pub title: Option<String>,
     pub description: Option<String>,
```
JakeHartnell, line 88:
```suggestion
    /// Whether the payroll contract is currently paused
    pub paused: bool,
    /// Human readable title for this contract
    pub title: Option<String>,
    /// Human readable description for this payroll contract
    pub description: Option<String>,
```

Perhaps, not 100% necessary, but makes for nice generated docs.

contracts/external/cw-payroll/src/linking.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        if initiator.admin != info.sender {
+            return Err(ContractError::Unauthorized {});
+        }
+
+        initiator.link_id = None;
+        initiator.is_link_initiator = false;
+        link.link_id = None;
+        link.is_link_initiator = false;
+
```
TrevorJTClarke, line 112: @deveusss we need to add logic to either pause or remove both streams here. I'd prefer the "pause".

The idea is that if there is a bi-directional stream, then either side breaks the link & both streams pause. This way each stream owner must assess if they continue or not.
Think of situation where 2 DAOs decide to do token swap, but then have a disagreement. 1 DAO detachs their side, causing both sides to pause. This makes sure that a bi-directional stream is fair, so that a detach doesnt let another  linked stream continue if it was previously linked.

contracts/external/cw-payroll/src/linking.rs
```diff
@@ -0,0 +1,190 @@
...
+        let response = Response::new()
+            .add_attribute("method", "link")
+            .add_attribute("initiator_id", initiator_id.to_string())
+            .add_attribute("link_id", link_id.to_string())
+            .add_attribute("admin", info.sender.clone());
+
+        Ok(response)
+    }
+
+    fn create_link_delete_msg(
```
TrevorJTClarke, line 124: @deveusss Im worried about this fn & `create_sync_msg` might have security implications.
I'd rather pull this out for now and keep the bidirectional linking only available on same-contract.
We can revisit if people desire multi-contract links.

- [ ] Remove `create_link_delete_msg`
- [ ] Remove `create_sync_msg`

TrevorJTClarke (changes requested):
Hi @deveusss -- great progress!!

See comments below for changes!

I also see we need a few more tests to have coverage of this feature. Here's the top level list of remaining todos:

- [ ] Remove all `is_link_initiator` logic
- [ ] Remove the cross-contract link logic
- [ ] Add test for resuming stream
- [ ] Add tests for invalid linking, invalid detach, valid linking, valid detach (confirm both pause)

Thank you!! We're getting close!

TrevorJTClarke (changes requested):
Hey @deveusss - I think we're there functionality wise, see comments on minor things.

As noted in discord, would love to consider moving away from left/right convention in favor of a fixed length vec. Thoughts?

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -49,39 +50,51 @@ pub fn instantiate(
...
+            stream.paused = true;
+            save_stream(storage, id, &stream).unwrap();
+            Ok(stream)
+        };
+    let stream = pause_stream_local(id, deps.storage).unwrap();
 
-    if stream.admin != info.sender {
-        return Err(ContractError::Unauthorized {});
+    if let Some(link_id) = stream.link_id {
+        pause_stream_local(link_id, deps.storage).unwrap();
```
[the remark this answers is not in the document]
deveusss: Gotcha, so left, right stream pair should be a part of contract state right?

contracts/external/cw-payroll/src/msg.rs
```diff
@@ -9,8 +9,28 @@ pub struct InstantiateMsg {
...
 }
 pub type StreamId = u64;
+pub type StreamIds = Vec<StreamId>;
 pub type ContractResult = Result<Response, ContractError>;
 
+pub(crate) trait StreamIdsExtensions {
+    fn second(&self) -> Option<&StreamId>;
+    fn validate(&self) -> Result<(), ContractError>;
+}
+impl StreamIdsExtensions for StreamIds {
```
TrevorJTClarke, line 19: I like this :)

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -175,21 +169,22 @@ pub fn execute_resume_stream(
...
+            stream.paused_time = None;
             save_stream(storage, id, &stream).unwrap();
             Ok(stream)
         };
 
-    let stream = pause_stream_local(id, deps.storage).unwrap();
+    let stream = resume_stream_local(id, deps.storage)?;
     if let Some(link_id) = stream.link_id {
-        pause_stream_local(link_id, deps.storage).unwrap();
+        resume_stream_local(link_id, deps.storage).unwrap();
```
TrevorJTClarke, line 179: TY!!!

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -961,17 +946,14 @@ mod tests {
         execute(deps.as_mut(), mock_env(), info, msg).unwrap();
 
         let sender = Addr::unchecked("bob").to_string();
-        let left_stream_id: StreamId = 1;
-        let right_stream_id: StreamId = 2;
+        let ids = vec![1, 2];
```
TrevorJTClarke, line 949: this is actually way more readable :)

TrevorJTClarke (approved):
Ummmmmm HELLLLYEAHHHHHHHH - thank you @deveusss  :)

PR #569, opened 2022-12-11
Fix claim_nft MAX_CLAIM out of gas error

de-husk:
Juno [decreased its `max_gas` param](https://wallet.keplr.app/chains/juno/proposals/44) limiting each block to 10000000 gas instead of the old value of 100000000.

I wanted to change our integration tests to support this, but I kept getting an out of gas error sending [this claim_nft tx](https://github.com/DA0-DA0/dao-contracts/blob/fa567797e2f42e70296a2d6f889f341ff80f0695/ci/integration-tests/src/tests/dao_voting_cw721_staked_test.rs#L117-L128).

We need to lower `MAX_CLAIMS` to something that wont lock people out from claiming the outstanding NFTs, keeping in mind that we should have some headroom incase this param is lowered in the future.


.cargo/config
```diff
@@ -3,4 +3,4 @@ wasm = "build --release --lib --target wasm32-unknown-unknown"
 wasm-debug = "build --lib --target wasm32-unknown-unknown"
 t = "test --workspace"
 unit-test = "test --workspace"
-integration-test = "test --package integration-tests -- --ignored --test-threads 1"
+integration-test = "test --package integration-tests -- --ignored --test-threads 1 -Z unstable-options --report-time"
```
de-husk, line 6: I added this so we can see how many seconds each integration test was taking.

justfile
```diff
@@ -36,15 +36,15 @@ deploy-local: download-deps
...
+		-e GAS_LIMIT=10000000 \
 		-e MAX_BYTES=22020096 \
 		-e UNSAFE_CORS=true \
 		-p 1317:1317 \
 		-p 26656:26656 \
 		-p 26657:26657 \
 		-p 9090:9090 \
 		--mount type=volume,source=junod_data,target=/root \
-		ghcr.io/cosmoscontracts/juno:v9.0.0 /opt/setup_and_run.sh {{test_addrs}}
+		ghcr.io/cosmoscontracts/juno:v11.0.0 /opt/setup_and_run.sh {{test_addrs}}
```
de-husk, line 47: decided to upgrade juno version, but this is unrelated to the max gas problem.

contracts/voting/dao-voting-cw721-staked/src/state.rs
```diff
@@ -39,7 +39,7 @@ pub const TOTAL_STAKED_NFTS: SnapshotItem<Uint128> = SnapshotItem::new(
 );
 
 /// The maximum number of claims that may be outstanding.
-pub const MAX_CLAIMS: u64 = 100;
+pub const MAX_CLAIMS: u64 = 70;
```
de-husk, line 42: What do you think we should lower this to?

Here is some data with different values of MAX_CLAIMS:
```
* 100:
    * ERROR
    * gas_used: 57793
    * gas_wanted: 12131747

* 95:
    * ERROR
    * gas_used: 57793
    * gas_wanted: 11533653

* 90: 
    * ERROR
    * gas_used: 57793
    * gas_wanted: 10935558

* 85:
    * ERROR
    * gas_used: 57793
    * gas_wanted: 10337451
   
* 80: 
    * WORKS
    * gas_used: 57793
    * gas_wanted: 9739358

* 70: 
    * WORKS
    * gas_used: 57753
    * gas_wanted: 8543711
```

de-husk:
> # Cosm-Orc Gas Usage
> Contract	Op Name	Gas Used	Old Gas Used	Gas Diff	File
> cw721_base	Instantiate__instantiate_cw721_base	164837	163837	+0.6104%	ci/integration-tests/src/tests/dao_voting_cw721_staked_test.rs:22
> cw721_base	Store__Store	3369402	28006860	-87.9694%	ci/integration-tests/src/helpers/chain.rs:97
> dao_voting_cw721_staked	Execute__claim_nfts	5710869	8103298	-29.5241%	ci/integration-tests/src/tests/dao_voting_cw721_staked_test.rs:120
> dao_voting_cw721_staked	Store__Store	3326008	27639680	-87.9665%	ci/integration-tests/src/helpers/chain.rs:97
> multiple_contracts	Execute__batch_cw721_stake_max_claims	4196439	43191012	-90.2840%	ci/integration-tests/src/tests/dao_voting_cw721_staked_test.rs:251
> dao_core	Store__Store	5628490	47122220	-88.0556%	ci/integration-tests/src/helpers/chain.rs:97
> cw20_base	Store__Store	4141992	34544160	-88.0096%	ci/integration-tests/src/helpers/chain.rs:97
> cw20_stake	Store__Store	3589024	29865200	-87.9826%	ci/integration-tests/src/helpers/chain.rs:97
> cw4_group	Store__Store	2781763	23034530	-87.9235%	ci/integration-tests/src/helpers/chain.rs:97
> cw_admin_factory	Store__Store	2003453	16448830	-87.8201%	ci/integration-tests/src/helpers/chain.rs:97
> cw_token_swap	Store__Store	2371616	19515145	-87.8473%	ci/integration-tests/src/helpers/chain.rs:97
> dao_pre_propose_approval_single	Store__Store	4903233	40985430	-88.0366%	ci/integration-tests/src/helpers/chain.rs:97
> dao_pre_propose_approver	Store__Store	3659835	30464370	-87.9865%	ci/integration-tests/src/helpers/chain.rs:97
> dao_pre_propose_multiple	Store__Store	4233291	35316690	-88.0133%	ci/integration-tests/src/helpers/chain.rs:97
> dao_pre_propose_single	Store__Store	4154927	34653610	-88.0101%	ci/integration-tests/src/helpers/chain.rs:97
> dao_proposal_multiple	Store__Store	6360962	53320060	-88.0702%	ci/integration-tests/src/helpers/chain.rs:97
> dao_proposal_single	Store__Store	6655269	55810350	-88.0752%	ci/integration-tests/src/helpers/chain.rs:97
> dao_voting_cw20_staked	Store__Store	3621797	30142510	-87.9844%	ci/integration-tests/src/helpers/chain.rs:97
> dao_voting_cw4	Store__Store	2670847	22096010	-87.9125%	ci/integration-tests/src/helpers/chain.rs:97
> dao_voting_native_staked	Store__Store	2951790	24473220	-87.9387%	ci/integration-tests/src/helpers/chain.rs:97
> dao_voting_staking_denom_staked	Store__Store	2196893	18085630	-87.8528%	ci/integration-tests/src/helpers/chain.rs:97
> stake_cw20_external_rewards	Store__Store	3097923	25709730	-87.9504%	ci/integration-tests/src/helpers/chain.rs:97
> stake_cw20_reward_distributor	Store__Store	2431608	20071680	-87.8854%	ci/integration-tests/src/helpers/chain.rs:97
> Raw Report for [55cb2f9](https://github.com/DA0-DA0/dao-contracts/commit/55cb2f9299620786104f6bd807f554859784cf4b)

All of our wasm storage costs were decreased a lot, because in juno v10, they lowered the compile cost of wasm file uploads (in order to support the max block gas lowering): https://github.com/CosmosContracts/juno/pull/278/files#diff-0763ee624d1f9f1b6f4475dea92cff0e2883f0d277a978f61a16aea70e2cd490R11

justfile
```diff
@@ -36,15 +36,15 @@ deploy-local: download-deps
 	docker run --rm -d --name cosmwasm \
 		-e PASSWORD=xxxxxxxxx \
 		-e STAKE_TOKEN=ujunox \
-		-e GAS_LIMIT=100000000 \
+		-e GAS_LIMIT=10000000 \
```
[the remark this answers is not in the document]
de-husk: Ideally we would break the build if this drifts. I can do this in another PR.

justfile
de-husk, line 47: We need to upgrade juno version since they made changes to wasm compile gas costs to support the max block gas lowering changes.

.github/workflows/integration_tests.yml
```diff
@@ -47,6 +48,15 @@ jobs:
...
         run: cargo install just || true
 
+      - name: Get mainnet GAS_LIMIT
+        run: echo "MAINNET_GAS_LIMIT=$(curl -s https://juno-rpc.polkachu.com/consensus_params | jq -r '.result.consensus_params.block.max_gas')" >> $GITHUB_ENV
+
+      - name: Mainnet block GAS_LIMIT changed
+        if: ${{ env.MAINNET_GAS_LIMIT != env.GAS_LIMIT }}
+        uses: actions/github-script@v6
+        with:
+          script: core.setFailed(`Integration tests must update GAS_LIMIT from ${process.env.GAS_LIMIT} to ${process.env.MAINNET_GAS_LIMIT}`)
```
0xekez, line 58: so cool that you added this check

ci/integration-tests/src/tests/dao_voting_cw721_staked_test.rs
```diff
@@ -238,12 +242,20 @@ fn cw721_stake_max_claims_works(chain: &mut Chain) {
...
+
+            reqs = vec![];
+        }
     }
 
-    chain
-        .orc
-        .execute_batch("batch_cw721_stake_max_claims", reqs, &user_key)
-        .unwrap();
+    assert_eq!(total_msgs as u64, MAX_CLAIMS * 3);
```
0xekez, line 258: great check 🙏

contracts/voting/dao-voting-cw721-staked/src/state.rs
0xekez: this feels like a very complex expected value calculation that i lack the ability to make. on one hand, bad if NFTs are stuck temporarily in the contract, on the other hand _very_ bad if Juno lowers the block gas limit again and then NFTs are stuck until we can do an upgrade..

i think 70 sounds reasonable. do you know why Juno lowered the block gas limit? knowing that might help us evaluate `P(lower gas limit in the future)`.

0xekez (approved):
amazing PR. love your invariant checks with max gas and the total number of messages sent 🙏

contracts/voting/dao-voting-cw721-staked/src/state.rs
de-husk: Im having a hard time finding more info on why they lowered it. I assume it was for security reasons. There was probably a contract doing something really computationally expensive that was either causing a small halt or maybe threatening to cause a halt.

de-husk: The other thing to keep in mind when we are talking about gas price estimation for these txs (not the actual gas_used that comes back from the tx response), is that we are using some gas price constants that are [bit larger than what most people use](https://github.com/DA0-DA0/dao-contracts/blob/main/ci/configs/cosm-orc/ci.yaml#L6-L7). So we have more headroom for that reason too. 

It's hard to find consistent values for these to compare, but it looks like most people use  a `gas_adjustment` from 1.0 -> 1.5 (we are using 1.3 in cosmjs). And it looks like we are using the [min gas price in dao dao ui](https://github.com/DA0-DA0/dao-dao-ui/blob/development/packages/stateful/components/WalletProvider.tsx#L79), but the recommended 'high' from the [chain registry](https://github.com/cosmos/chain-registry/blob/master/juno/chain.json#L18-L21) is half of what we are using in the integration tests.

de-husk: The other thing to keep in mind is that when we are talking about our gas simulation as opposed to actual gas 
usage, is that we are using a `gas_adjustment` that is a bit higher that what most people use. 

[Ours is 1.5](https://github.com/DA0-DA0/dao-contracts/blob/main/ci/configs/cosm-orc/ci.yaml#L7). It's hard to find consistent values for these to compare, but it looks like most people use  a `gas_adjustment` from 1.0 -> 1.5 (we are using [1.3 in cosmjs for example](https://github.com/cosmos/cosmjs/blob/main/packages/stargate/src/signingstargateclient.ts#L286) in the DAO DAO UI). 

So we have a bit more headroom for that reason too.

0xekez: > So we have a bit more headroom for that reason too

nice :)

> I assume it was for security reasons.

this is my assumption as well. based on that, i think it is OK to make a "leap of faith" here that the new value was tested to be safe and that we don't need to lower again.

i'm good with `70` :)

JakeHartnell (approved):
LGTM!

PR #571, opened 2022-12-13
Add feature flag for cw20 support in dao-core.

0xekez:
This puts cw20 support in the dao-core contract behind a feature flag. The feature is not enabled by default, so for DAOs that want to use cw20s this is a no-op.

As part of this change I have also added lints and tests to CI which run with no default features enabled so we will detect regressions if a future change depends on cw20 support, but does not properly gate itself behind the feature flag.


JakeHartnell (approved):
HELL YEAH. :heart:

.github/workflows/basic.yml
```diff
@@ -29,6 +29,15 @@ jobs:
         env:
           RUST_BACKTRACE: 1
 
+      - name: Run tests w/o default features
+        uses: actions-rs/cargo@v1
```
de-husk, line 33: Just a heads up that we need to stop using `actions-rs/cargo`, I made an issue: https://github.com/DA0-DA0/dao-contracts/issues/572

contracts/dao-core/Cargo.toml
```diff
@@ -12,8 +12,11 @@ crate-type = ["cdylib", "rlib"]
 [features]
 # for more explicit tests, cargo test --features=backtraces
 backtraces = ["cosmwasm-std/backtraces"]
+# enables cw20 treasury management
+cw20 = []
 # use library feature to disable all instantiate/execute/query exports
 library = []
+default = ["cw20"]
```
de-husk, line 19: So, this means that we will need to release 2 different versions of this contract?

We just need to be sure we are careful not to mix up the CHECKSUMs when we do this, and make this clear.

contracts/dao-core/README.md
```diff
@@ -11,3 +11,75 @@ In additon to the wiki spec this contract may also pause. To do so a
...
+        to_add: Vec<String>,
+        to_remove: Vec<String>,
+    },
+	// ...
+}
+```
+
+### Disabling cw20 support
+
+There is an effort on CosmWasm chains to depreciate the cw20 standard
```
de-husk, line 81: NIT: `depreciate` -> `deprecate`

contracts/dao-core/README.md
```diff
@@ -11,3 +11,75 @@ In additon to the wiki spec this contract may also pause. To do so a
...
+in favor of native tokens [generated by the TokenFactory
+module](https://hackmd.io/@reecepbcups/cw20-to-tokenfactory). By
+default, this contract has cw20 support. It may be disabled by
+disabling default features (ex. `cargo build --no-default-features`).
```
de-husk, line 85: Will we be generating the schema / ts files for this no cw20 support? Or is that on the user to figure out if they want that?

de-husk (approved):

contracts/dao-core/Cargo.toml
0xekez: this + your other comment about schema generation are good. my thought is that we can delay working out the specifics of that until we deploy on a chain that doesn't want cw20 support. at that time we can give build instructions for building and verifying without this feature flag enabled.

i've made an issue in rust-optmizer for this so we can see where that goes: https://github.com/CosmWasm/rust-optimizer/issues/107

contracts/dao-core/README.md
0xekez: thanks :)

contracts/dao-core/README.md
0xekez: see response to other comment. thinking we can lazily evaluate this. good question :)

contracts/dao-core/Cargo.toml
de-husk: Sounds good, if its too complicated we can always switch to an instantiation param that just blocks access to these flows if it's set.

PR #586, opened 2022-12-18
Use voting power at time of proposal creation for only_members_execute

### LINKED ISSUE #575: Determine proposal executability based on voting power at time of proposal creation
When `only_members_execute` is true we currently determine membership based on voting power at time of execution, to be consistent with our philosophy that you should know exactly who you're proposing to when creating a proposal, we should determine executability based on voting power at the time the proposal was created.

NoahSaso:
Resolves #575 

This makes it so that `only_members_execute` uses the voting power at time of proposal creation to validate membership during execution instead of at the current time.


0xekez (approved):
nice! thanks

JakeHartnell (approved):
Nice. :heart:

PR #588, opened 2022-12-19
Require unstaking amount to be non-zero.

### LINKED ISSUE #585: Validate that the amount being unstaked is non-zero in native staking
Currently we can get claims with zero associated tokens stuck in the module which is not ideal.

0xekez:
resolves #585


JakeHartnell (approved):
:heart:

NoahSaso (approved):

de-husk (approved):

PR #590, opened 2022-12-19
Check cw20 responds to TokenInfo query during cw20-stake instantiation.

### LINKED ISSUE #581: Validate that cw20 token is configured as staking asset during cw20-stake instantiation
We can do this by querying for the addresse's `TokenInfo` and erroring if the response is invalid.

0xekez:
resolves #581

### DIFF
--- a/contracts/staking/cw20-stake/src/contract.rs
+++ b/contracts/staking/cw20-stake/src/contract.rs
@@ -7,7 +7,7 @@ use cosmwasm_std::{
     MessageInfo, Response, StdError, StdResult, Uint128,
 };
 
-use cw20::Cw20ReceiveMsg;
+use cw20::{Cw20ReceiveMsg, TokenInfoResponse};
 
 use crate::hooks::{stake_hook_msgs, unstake_hook_msgs};
 use crate::math;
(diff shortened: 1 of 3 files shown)


JakeHartnell (approved):
Nice.

contracts/staking/cw20-stake/src/contract.rs
```diff
@@ -72,11 +72,17 @@ pub fn instantiate(
         None => None,
     };
 
+    let token_address = deps.api.addr_validate(&msg.token_address)?;
+    let _: TokenInfoResponse = deps
```
elsehow, line 76: What's going on here? Is this just a side-effecty thing? Would be nice to know what's happening.

I assume the point of this it to error if `InvalidCw20`, but I'm only able to infer that in the context of this PR. A comment here would be nice. It might also explain why this is necessary...

NoahSaso (approved):

contracts/staking/cw20-stake/src/contract.rs
0xekez: thanks for the question. i've added a comment. this is a smoke test that we are dealing with a valid cw20. what's happening here:

1. we make a query against the provided address to see if it responds to the `TokenInfo` query message from the cw20 spec.
2. if the query fails, we transform the error into `InvalidCw20` and bail out.
3. if the query succeeds, we ignore the response as we don't actually care about the contents (hence, the `_` variable).

PR #593, opened 2022-12-19
Add ProposalModuleCount query.

### LINKED ISSUE #577: Expose active and total proposal module count outside of `DumpState` query
The dump state query is allowed to run out of gas as it exists only as a performance optimization in the frontend. Currently we only expose the active and total proposal module count inside that query. To keep that state queryable, we should move that information into a new query.

0xekez:
resolves #577

### DIFF
--- a/contracts/dao-core/schema/dao-core.json
+++ b/contracts/dao-core/schema/dao-core.json
@@ -1772,6 +1772,20 @@
         "additionalProperties": false
       },
       {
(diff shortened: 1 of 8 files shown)


JakeHartnell (approved):
LGTM!

NoahSaso (approved):

de-husk (approved):
Cool! This is a v useful endpoint

PR #594, opened 2022-12-19
Disallow duplicate items during instantiation.

### LINKED ISSUE #576: Deduplicate initial items list during DAO instantiation
It should be an error to provide duplicate item keys during DAO instantiation.

0xekez:
resolves #576

### DIFF
--- a/contracts/dao-core/schema/dao-core.json
+++ b/contracts/dao-core/schema/dao-core.json
@@ -49,7 +49,7 @@
         ]
       },
       "initial_items": {
-        "description": "Initial information for arbitrary contract addresses to be added to the items map. The key is the name of the item in the items map. The value is an enum that either uses an existing address or instantiates a new contract.",
+        "description": "Initial information for arbitrary contract addresses to be added to the items map. The key is the name of the item in the items map. The value is an enum that either uses an existing address or instantiates a new contract.\n\nIt is an error to provide two items with the same key.",
         "type": [
           "array",
           "null"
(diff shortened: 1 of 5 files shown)


JakeHartnell (approved):

contracts/dao-core/schema/dao-core.json
```diff
@@ -49,7 +49,7 @@
         ]
       },
       "initial_items": {
-        "description": "Initial information for arbitrary contract addresses to be added to the items map. The key is the name of the item in the items map. The value is an enum that either uses an existing address or instantiates a new contract.",
+        "description": "Initial information for arbitrary contract addresses to be added to the items map. The key is the name of the item in the items map. The value is an enum that either uses an existing address or instantiates a new contract.\n\nIt is an error to provide two items with the same key.",
```
elsehow, line 52: "It is an error" is a little odd to me. I actually think this whole description could be:

> Initial information for arbitrary contract addresses to be added to the items map. The items map maps the key, a name, to the value, an enum that either uses an existing address or instantiates a new contract.\n\nNote: keys must be unique. Adding an item with an already-existing key will result in an error.

elsehow (approved):
Minor note.

NoahSaso (approved):
noice.

I'm pretty sure items no longer have anything to do with contract addresses or instantiating them. They're just arbitrary data, right? We should probably change the comment/schema to reflect that.

de-husk (approved):

PR #597, opened 2022-12-29
Two-step ownership transfer for staking contracts

0xekez:
this implements a two-step ownership transfer for all of our staking contracts using the [`cw_ownable`](https://crates.io/crates/cw-ownable) crate.

for each contract in `staking/*`:

1. the `manager` is removed
2. `cw_ownable` integrated
3. migration from version one to two is implemented

it is my opinion that having a contract-level owner and CosmWasm-level admin that is a DAO has approximately the same semantics as having an owner and a manager. assuming the admin is a DAO, and the owner an operations multisig:

> the operations multisig has the ability to update unstaking durations, but does not have the ability to update the admin. the admin has the ability to remove the operations multisig and migrate the contract.

that sentence could be reused in a situation where there is an owner, manager, and admin and still be accurate.

### DIFF
--- a/contracts/dao-core/Cargo.toml
+++ b/contracts/dao-core/Cargo.toml
@@ -17,7 +17,6 @@ library = []
 
 [dependencies]
 cosmwasm-std = { workspace = true, features = ["ibc3"] }
-cosmwasm-storage = { workspace = true }
 cosmwasm-schema = { workspace = true }
 cw-storage-plus = { workspace = true }
 cw2 = { workspace = true }
--- a/contracts/staking/cw20-stake-external-rewards/src/contract.rs
+++ b/contracts/staking/cw20-stake-external-rewards/src/contract.rs
@@ -38,19 +38,11 @@ pub fn instantiate(
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
 
     let owner = msg.owner.map(|a| deps.api.addr_validate(&a)).transpose()?;
-    let manager = msg
-        .manager
-        .map(|a| deps.api.addr_validate(&a))
-        .transpose()?;
 
     let reward_token = match msg.reward_token {
         Denom::Native(denom) => Denom::Native(denom),
         Cw20(addr) => Cw20(deps.api.addr_validate(addr.as_ref())?),
     };
-
-    if msg.reward_duration == 0 {
-        return Err(ContractError::ZeroRewardDuration {});
-    }
 
     // Verify contract provided is a staking contract
     let _: cw20_stake::msg::TotalStakedAtHeightResponse = deps.querier.query_wasm_smart(
@@ -60,13 +52,15 @@ pub fn instantiate(
 
     let config = Config {
         owner,
-        manager,
         staking_contract: deps.api.addr_validate(&msg.staking_contract)?,
         reward_token,
     };
     CONFIG.save(deps.storage, &config)?;
 
-    // Non-zero rewards duration checked above.
+    if msg.reward_duration == 0 {
+        return Err(ContractError::ZeroRewardDuration {});
+    }
+
     let reward_config = RewardConfig {
         period_finish: 0,
         reward_rate: Uint128::zero(),
@@ -79,13 +73,6 @@ pub fn instantiate(
             "owner",
             config
                 .owner
-                .map(|a| a.into_string())
-                .unwrap_or_else(|| "None".to_string()),
-        )
-        .add_attribute(
-            "manager",
-            config
-                .manager
                 .map(|a| a.into_string())
                 .unwrap_or_else(|| "None".to_string()),
         )
@@ -125,9 +112,6 @@ pub fn execute(
             execute_update_reward_duration(deps, env, info, new_duration)
         }
         ExecuteMsg::UpdateOwner { new_owner } => execute_update_owner(deps, env, info, new_owner),
-        ExecuteMsg::UpdateManager { new_manager } => {
-            execute_update_manager(deps, env, info, new_manager)
-        }
     }
 }
 
@@ -171,7 +155,7 @@ pub fn execute_fund(
     amount: Uint128,
 ) -> Result<Response<Empty>, ContractError> {
     let config = CONFIG.load(deps.storage)?;
-    if config.owner != Some(sender.clone()) && config.manager != Some(sender.clone()) {
+    if config.owner != Some(sender.clone()) {
         return Err(Unauthorized {});
     };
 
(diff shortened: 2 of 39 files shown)


0xekez:
imo, having a manager and owner as these contracts are currently configured is EXACTLY the same semantically as having an owner and a contract level admin. the way these were used (RAW DAO rest in peace..) would fit the admin / owner model as well.

JakeHartnell (approved):
Nice to see @larry0x's `cw-ownership` in use. :heart:

0xekez:
let's wait for this before landing this PR as we'll have to update the initialization call: https://github.com/steak-enjoyers/cw-plus-plus/pull/7

larry0x:
@0xekez i’ve merged it. Will make a new release a bit later

contracts/staking/cw20-stake-reward-distributor/src/contract.rs
```diff
@@ -34,19 +33,19 @@ pub fn instantiate(
     }
 
     let config = Config {
-        owner: owner.clone(),
         staking_addr: staking_addr.clone(),
         reward_token: reward_token.clone(),
         reward_rate: msg.reward_rate,
     };
     CONFIG.save(deps.storage, &config)?;
+    cw_ownable::initialize_owner(deps.storage, deps.api, &msg.owner)?;
```
blue-note, line 41: does the owner for existing contracts need to be initialized?

0xekez: oh great catch, thanks. i'll add migration logic for this as well

blue-note (approved):

0xekez:
pre-empting some review: i understand that there is some less-than-dry code in these changes, specifically:

1. `execute_update_ownership`
2. tests for ownership transfers
3. migration version checks

the code that is duplicated is so simple and brief, that imo it is more readable just written out each time instead of hiding it behind a level of indirection and an abstraction.

contracts/dao-core/Cargo.toml
```diff
@@ -17,7 +17,6 @@ library = []
 
 [dependencies]
 cosmwasm-std = { workspace = true, features = ["ibc3"] }
-cosmwasm-storage = { workspace = true }
```
JakeHartnell, line 20: lol

contracts/staking/cw20-stake-external-rewards/src/contract.rs
```diff
@@ -37,36 +37,29 @@ pub fn instantiate(
...
 
     let reward_token = match msg.reward_token {
         Denom::Native(denom) => Denom::Native(denom),
         Cw20(addr) => Cw20(deps.api.addr_validate(addr.as_ref())?),
     };
 
-    if msg.reward_duration == 0 {
-        return Err(ContractError::ZeroRewardDuration {});
-    }
-
```
JakeHartnell, line 54: We sure we want to remove this?

JakeHartnell (approved):
Really nice to clean this up. :heart:

contracts/staking/cw20-stake-external-rewards/src/contract.rs
0xekez: moved it down to line 59 :) before we had a comment there saying we checked it above. figured it made more sense to just remove the comment and check the value where we use it.
image

contracts/dao-core/Cargo.toml
0xekez: i do not practice what i preach 😂

contracts/staking/cw20-stake-external-rewards/src/state.rs
```diff
@@ -6,8 +6,6 @@ use cw_storage_plus::{Item, Map};
 
 #[cw_serde]
 pub struct Config {
-    pub owner: Option<Addr>,
```
blue-note, line 9: i'd consider also renaming the config key to "config_v2"? for other migrations we've done this to avoid overriding state just in case

0xekez: great point. thanks. updated

contracts/staking/cw20-stake/src/contract.rs
```diff
@@ -506,34 +473,33 @@ pub fn query_list_stakers(
...
-        pub admin: Addr,
-        pub token_address: Addr,
-        pub unstaking_duration: Option<Duration>,
-    }
+    use cw20_stake_v1 as v1;
 
+    let ContractVersion { version, .. } = get_contract_version(deps.storage)?;
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
     match msg {
-        MigrateMsg::FromBeta { manager } => {
```
blue-note, line 520: are there no more beta migrations left? 

i'm wondering if beta --> v2 is a concern

contracts/staking/cw20-stake/src/contract.rs
```diff
@@ -290,7 +252,7 @@ pub fn execute_unstake(
         }
         Some(duration) => {
             let outstanding_claims = CLAIMS.query_claims(deps.as_ref(), &info.sender)?.claims;
-            if outstanding_claims.len() >= MAX_CLAIMS as usize {
+            if outstanding_claims.len() + 1 > MAX_CLAIMS as usize {
```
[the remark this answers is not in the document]
0xekez: i feel this is slightly more clear in saying:

> add this to the list of claims if adding this to the list of claims would not cause there to be too many outstanding claims.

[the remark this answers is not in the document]
0xekez: this is also so that the code is symmetric with this one: https://github.com/DA0-DA0/dao-contracts/commit/362a8e402a08b5d241262b3d07db0bff6eb30c2d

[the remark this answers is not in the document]
blue-note: that is awesome and indeed more clear

contracts/staking/cw20-stake/src/contract.rs
0xekez: there are none so far as i know. i did some musing on this and think that if you do want to do a DAO upgrade that spans multiple versions, the process will always be:

```
for each version:
  migrate(version)
```

we tag our releases such that doing this isn't too bad. if you need an intermediate code ID, you can always check out the tag and use docker to build a reproducible wasm.

otherwise, doing migrations across multiple versions is quite hard as you may have multiple layers of state changes to apply. for example, to migrate from beta to v2 on this contract you'd need to:

1. add the `manager` state key, do the migration to v1.
2. remove the `manager` state key, do the migration to v2.

this feels like a fair bit to keep track of, when we could require a little extra work during migration and keep these migrations to one-per-major-version.

does that sound reasonable to you?

blue-note: yeah, I think this is reasonable!

PR #602, opened 2023-01-03
WIP cw-payroll

JakeHartnell:
Supports:
- [x] support for both native and cw20 tokens
- [x] vesting schedules (including with cliffs thanks to `wynd-utils`)
- [x] cancellation via an optional contract owner
- [x] queries for listing a DAO's vesting payments
- [x] Optionally stake underlying native assets

WIP:
- [ ] More helpful queries
- [ ] Lots of tests
- [ ] Update readme

PR #608, opened 2023-01-08
cw-vesting with payroll factory

JakeHartnell:
Building off of #602. This simplifies things and minimizes risk by using the factory pattern.

Supports staking, used some code inspired by what we did for mesh security recently, so curious what folks think of rewards calculation, etc.

TODO:
- [x] Factory contract
- [x] Redelegate
- [x] Query list of payroll contracts by instantiator
- [x] Query list of payroll contracts by recipient
- [x] Query vested amount
- [x] MORE TESTS


contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+};
+
+use cw2::set_contract_version;
+use cw_payroll::msg::InstantiateMsg as PayrollInstantiateMsg;
+use cw_utils::parse_reply_instantiate_data;
+
+use crate::error::ContractError;
+use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg};
+
+pub(crate) const CONTRACT_NAME: &str = "crates.io:cw-admin-factory";
```
blue-note, line 14: contract name needs to be changed :)

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub fn instantiate_contract(
+    _env: Env,
+    info: MessageInfo,
+    instantiate_msg: PayrollInstantiateMsg,
+    code_id: u64,
+    label: String,
+) -> Result<Response, ContractError> {
+    // Instantiate the specified contract with owner as the admin.
+    let instantiate = WasmMsg::Instantiate {
```
blue-note, line 55: are there other planned functions of this factory contract other than allowing payroll contracts to be instantiated as their own admins? (I think I remember you saying something about tracking which payroll contracts belong to which DAOs). If that is the case, what is the benefit of using a factory contract vs tracking this information in dao-core?

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: Lol, fixed.

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: We could track the information in DAO core with items....

Have some commits I'm going to push soon that will clarify the direction.

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,183 @@
...
+        INSTANTIATE_CONTRACT_REPLY_ID => {
+            let res = parse_reply_instantiate_data(msg)?;
+            let contract_addr = deps.api.addr_validate(&res.contract_address)?;
+
+            let (recipient, instantiator) = TMP_CONTRACT_INFO.load(deps.storage)?;
+
+            // Save vesting contract payment info
+            vesting_contracts().save(
+                deps.storage,
+                contract_addr.as_str().clone(),
```
blue-note, line 161: any particular reason to store this as a string vs `Addr`?

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractError> {
+    match msg.id {
+        INSTANTIATE_CONTRACT_REPLY_ID => {
+            let res = parse_reply_instantiate_data(msg)?;
+            let contract_addr = deps.api.addr_validate(&res.contract_address)?;
+
+            let (recipient, instantiator) = TMP_CONTRACT_INFO.load(deps.storage)?;
```
blue-note, line 156: is there any case in which TMP_CONTRACT_INFO would not correspond to the contract which has just been instantiated?

contracts/external/cw-payroll-factory/src/state.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+impl<'a> IndexList<VestingContract> for TokenIndexes<'a> {
+    fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<VestingContract>> + '_> {
+        let v: Vec<&dyn Index<VestingContract>> = vec![&self.owner, &self.recipient];
+        Box::new(v.into_iter())
+    }
+}
+
+pub fn vesting_contracts<'a>() -> IndexedMap<'a, &'a str, VestingContract, TokenIndexes<'a>> {
```
blue-note, line 27: this is a cool use of IndexedMap

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,183 @@
...
+        } => {
+            let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
+            let start = start_after.map(Bound::<String>::exclusive);
+
+            // Validate recipient address
+            deps.api.addr_validate(&recipient)?;
+
+            let res: Vec<VestingContract> = vesting_contracts()
+                .idx
+                .recipient
```
blue-note, line 136: why did you decide to make one vesting contract per recipient?

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: We're passing the `recipient` and `instantiator` through to the submessage here. Not 100% sure if this is the best pattern.

Alternately, we could query the newly created contract... which may be better?

contracts/external/cw-payroll-factory/src/state.rs
JakeHartnell: Thanks!

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: Simplifies a lot of things with regard to staking. Also minimizes risk of cross-contaminating funds. Note: I actually tried to make a multi-recipient vesting contract, but it was starting to get messy. There are a bunch of weird edge case around unbonding periods as well.

tl;dr thought it was simpler and safer.

contracts/external/cw-payroll-factory/src/contract.rs
blue-note: I don't think we would be at risk for TMP_CONTRACT_INFO holding incorrect information as in the case of multiple instantiation requests, submessage order will still be preserved, and in the case of an instantiation failure, TMP_CONTRACT_INFO still needs to be set to the correct info before we handle reply success here. querying the created contract would be clearer, but not sure it's entirely necessary

contracts/external/cw-payroll-factory/src/contract.rs
blue-note: I think this factory pattern makes sense -- a payroll contract does not always have to be used by a DAO so I agree they should be kept separate

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -0,0 +1,589 @@
...
+            nonpayable(&info)?;
+
+            // Validate all vesting payment params
+            let checked_params = msg.params.into_checked(deps.as_ref())?;
+
+            // Create new vesting payment
+            let mut vesting_payment = VestingPayment::new(checked_params)?;
+
+            // Set status to unfunded
+            vesting_payment.status = VestingPaymentStatus::Unfunded;
```
blue-note, line 58: small nit on design: I think it's a bit unintuitive that native vesting contracts can be funded immediately but cw20 payments need to be created and then funded in a separate step. I understand the technical reason for why this has to be, but an alternative design might be to just require both native and cw20 payments to be created and funded in two separate steps

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -0,0 +1,589 @@
...
+}
+
+pub fn execute_receive_cw20(
+    _env: Env,
+    deps: DepsMut,
+    info: MessageInfo,
+    receive_msg: Cw20ReceiveMsg,
+) -> Result<Response, ContractError> {
+    // Only accepts cw20 tokens
+    nonpayable(&info)?;
```
blue-note, line 107: i'm imagining a scenario in which the payer does not have all of the funds available for the entire vesting contract upfront. Are they to keep creating new vesting contracts on an ongoing basis in this case? that might cause a large number of contract instantiations (which I'm not sure is a concern in and of itself). What do you think about the ability to re-fund and extend an existing vesting payment? unsure if this complicates the use of curves as well

blue-note:
I think the design is looking good!! I did a first pass, going to do a more detailed look on the next pass

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: Makes things simple and easy to reason about, which is a plus IMO.

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: @0xekez, any preference or thoughts here?

contracts/external/cw-payroll/src/contract.rs
JakeHartnell: I think we can do it with the factory... will try that today.

contracts/external/cw-payroll/src/contract.rs
JakeHartnell: Hmmmmm, could be a nice v2. Actually took out pausing as well because I felt it wasn't essential for the MVP.

Instantiating contracts is cheap.

I think the appeal of the vesting contract from the `recipients` perspective is that they no all the funds for the payment are in the contract. There is a nice piece of mind about that.

JakeHartnell: V2 could support pausing or updating the existing payment.

Unless we can think of a clean way to do it, I think it complicates things a bit. Can always cancel and instantiate a new contract for now.

contracts/external/cw-payroll/src/state.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    //// TODO Do we need this?
+    // pub fn reset_pending_rewards(&mut self) {
+    //     self.rewards.pending -= self.rewards.pending.floor();
+    // }
+
+    /// Turn pending decimal to u128 to send tokens
+    pub fn get_pending_rewards(&self) -> Result<Uint128, ContractError> {
+        // TODO: Find a better way of doing this?
```
JakeHartnell, line 172: @0xekez would love your thoughts here, and also a sanity check on this math. :joy:

JakeHartnell: We have some similar code in the Mesh Security repo, and I think it's messy.

contracts/external/cw-payroll-factory/src/state.rs
0xekez: +1. this is really great.

contracts/external/cw-payroll-factory/src/contract.rs
0xekez: generally speaking, i think it's best to query from the contract in reply > storing state in tmp keys.

while unlikely to happen, this current design is vulnerable to rentrancy where if the vesting contract being instantiated called back into the factory (for some unknown or malicious reason), the tmp variable would get overridden.

contracts/external/cw-payroll/src/contract.rs
```diff
@@ -0,0 +1,599 @@
...
+        .query_delegation(env.contract.address, validator.clone())?;
+
+    // Total rewards we have from this validator
+    let total_accumulated_rewards = &match delegation_query {
+        Some(delegation) => delegation.accumulated_rewards,
+        None => return Err(ContractError::NoDelegationsForValidator {}),
+    };
+
+    // Check to make sure there are rewards
+    if total_accumulated_rewards.is_empty() || total_accumulated_rewards[0].amount.is_zero() {
```
0xekez, line 554: are we guaranteed that the first element in this list will be the native staking denomination? if not, we may want to check that here.

contracts/external/cw-payroll/src/state.rs
```diff
@@ -0,0 +1,395 @@
...
+        self.assert_schedule_vests_amount()?;
+
+        // Check valid recipient address
+        let recipient = deps.api.addr_validate(&self.recipient)?;
+
+        // Check denom
+        let checked_denom = match self.denom {
+            UncheckedDenom::Native(denom) => UncheckedDenom::Native(denom).into_checked(deps)?,
+            UncheckedDenom::Cw20(addr) => UncheckedDenom::Cw20(addr).into_checked(deps)?,
+        };
```
0xekez, line 42:
```suggestion
let checked_denom = self.denom.into_checked(deps)?;
```

are you worried about another denom type being added and not getting a compilation error here? if so, fine to leave the match statement imo. could also check that in a test.

contracts/external/cw-payroll/src/state.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        // If title is included, validate title length (max 280 characters)
```
0xekez, line 44: should we perform this validation for description as well?

contracts/external/cw-payroll/src/state.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /// Asserts the vesting schedule decreases to 0 eventually, 2and is never more than the
+    /// amount being sent. If it doesn't match these conditions, returns an error.
+    pub fn assert_schedule_vests_amount(&self) -> Result<(), ContractError> {
+        self.vesting_schedule.validate_monotonic_decreasing()?;
+        let (low, high) = self.vesting_schedule.range();
+        if low != 0 {
+            Err(ContractError::NeverFullyVested)
+        } else if high > self.amount.u128() {
```
0xekez, line 68: how about requiring that `high == self.amount.u128()`? seems like if the curve would not vest enough that should also be an error

0xekez (commented):
this looks great architecturally. as discussed in Discord, i think we can remove all the complexity of calculating and storing rewards by making the vest receiver the [withdraw address](https://github.com/cosmos/cosmos-sdk/blob/v0.42.4/proto/cosmos/distribution/v1beta1/tx.proto#L29-L37) for staked tokens. frontends can then handle providing an easy way to withdraw those funds.

once that's done i will do a detailed pass over the math. think this is close :)

contracts/external/cw-payroll/src/state.rs
JakeHartnell: Nah, I think it's fine IMO.

contracts/external/cw-payroll/src/contract.rs
JakeHartnell: Not needed anymore with the refactor!

JakeHartnell:
> this looks great architecturally. as discussed in Discord, i think we can remove all the complexity of calculating and storing rewards by making the vest receiver the [withdraw address](https://github.com/cosmos/cosmos-sdk/blob/v0.42.4/proto/cosmos/distribution/v1beta1/tx.proto#L29-L37) for staked tokens. frontends can then handle providing an easy way to withdraw those funds.
> 
> once that's done i will do a detailed pass over the math. think this is close :)

This was a great suggestion! Much cleaner now, and can probably be even cleaner still. Only challenge is test coverage... `cw-multi-test` does not support the `SetWithdrawAddress` distribution message: https://github.com/CosmWasm/cw-multi-test/issues/20

I tried to cover things as best as possible with unit tests, we should figure out what to do until this gets added upstream.

contracts/external/cw-payroll/src/state.rs
JakeHartnell: DELETED IT.

JakeHartnell: :heart:

mikedotexe:
Super psyched. No joke, this is among a handful of smart contracts that can change all of web3 in 2023 💪🏼

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,325 @@
...
+                        })?,
+                        funds: vec![],
+                    })]
+                }
+            };
+
+            Ok(Response::default()
+                .add_attribute("new_payroll_contract", contract_addr)
+                .add_messages(msgs))
+        }
```
webmaster128, line 315: Do I understand correctly that you could move all this code to `instantiate_contract` if you knew the address of the new instance in advance, avoiding the need for `TMP_INSTANTIATOR_INFO`?

I think Instantiate2 coming in CosmWasm 1.2 will give you that.

contracts/external/cw-payroll/src/contract.rs
JakeHartnell: Done!

contracts/external/cw-payroll-factory/src/contract.rs
JakeHartnell: Yes. That would work.

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,549 @@
...
+
+    // Check all params and create vesting payment with status based on denom
+    let vp = match msg.params.denom {
+        UncheckedDenom::Native(ref denom) => {
+            // Validate all vesting payment params
+            let checked_params = msg.clone().params.into_checked(deps.as_ref())?;
+
+            // Check amount sent matches amount in vesting payment
+            if checked_params.amount != must_pay(&info, &checked_params.denom.to_string())? {
+                return Err(ContractError::AmountDoesNotMatch);
```
blue-note, line 42: nit: I believe the error here should be `IncorrectDenom`

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,325 @@
...
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn execute(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
+        ExecuteMsg::Receive(msg) => execute_receive_cw20(env, deps, info, msg),
+        ExecuteMsg::InstantiatePayrollContract {
```
blue-note, line 56: nice! I like instantiating and funding through the factory. nit: call this `InstantiateNativePayrollContract`?

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,549 @@
...
+    dst_validator: String,
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    // Non-payable as this method redelegates funds in the contract
+    nonpayable(&info)?;
+
+    let vesting_payment = VESTING_PAYMENT.load(deps.storage)?;
+
+    // Check status is active
+    if vesting_payment.status != VestingPaymentStatus::Active {
```
blue-note, line 406: we could also use a check that src and dst validators are different

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,549 @@
...
+            if local_denom != denom {
+                return Err(ContractError::NotStakeable);
+            } else {
+                denom
+            }
+        }
+    };
+
+    // Load source validator, subtract redelegated amount
+    let amount_staked_src = STAKED_VESTING_BY_VALIDATOR.load(deps.storage, &src_validator)?;
```
blue-note, line 432: we could map this error to a more informational error that src validator is not delegated to

contracts/external/cw-vesting/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub fn get_vested_amount_by_seconds(&self, time: u64) -> Result<Uint128, OverflowError> {
+        let vesting_funds = self.vesting_schedule.value(time);
+        self.amount.checked_sub(vesting_funds)
+    }
+}
+
+/// A map of vesting payments (ID, VestingPayment)
+pub const VESTING_PAYMENT: Item<VestingPayment> = Item::new("vesting_payment");
+/// A map of staked vesting claims by validator
+pub const STAKED_VESTING_BY_VALIDATOR: Map<&str, Uint128> = Map::new("staked_vesting_by_validator");
```
blue-note, line 134: any reason to not have the key be validated addr?

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,549 @@
...
+
+    Ok(Response::new()
+        .add_attribute("method", "remove_vesting_payment")
+        .add_attribute("owner", info.sender)
+        .add_attribute("removed_time", env.block.time.to_string())
+        .add_messages(msgs))
+}
+
+pub fn execute_distribute(env: Env, deps: DepsMut) -> Result<Response, ContractError> {
+    let vesting_payment = VESTING_PAYMENT.load(deps.storage)?;
```
blue-note, line 234: we need a "is sender the recipient" check here too

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,549 @@
...
+
+    let vesting_payment = VESTING_PAYMENT.load(deps.storage)?;
+
+    // Check status is active
+    if vesting_payment.status != VestingPaymentStatus::Active {
+        return Err(ContractError::NotActive);
+    }
+
+    // Check sender is the recipient of the vesting payment
+    if info.sender != vesting_payment.recipient {
```
blue-note, line 332: nit: it's nice to have these authorization checks happen first as it delivers the most relevant / restrictive error first

JakeHartnell: Feel free to make a commit or PR.

contracts/external/cw-vesting/src/contract.rs
JakeHartnell: Don't think so. That way this can be automated by Croncat for example.

JakeHartnell: Payment always goes to the recipient so it should be fine.

blue-note: makes sense

blue-note (approved):

contracts/external/cw-vesting/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    /// Asserts the vesting schedule decreases to 0 eventually, 2and is never more than the
+    /// amount being sent. If it doesn't match these conditions, returns an error.
+    pub fn assert_schedule_vests_amount(&self) -> Result<(), ContractError> {
+        self.vesting_schedule.validate_monotonic_decreasing()?;
+        let (low, high) = self.vesting_schedule.range();
+        if low != 0 {
+            Err(ContractError::NeverFullyVested)
+        } else if high > self.amount.u128() {
+            Err(ContractError::VestsMoreThanSent)
```
0xekez, line 66: is it ok to vest less than sent? ex: `low == 0 && high < self.amount`

0xekez:
really obscure observation about this contract below. i don't think we need to change anything before merging, saying in the hope it is useful / reveals something i don't understand.

---

let $curve(t)$ be the value of the vesting curve at, and $claimed(t)$ be total claimed up to time $t$.

in this contract, we construct the curve such that $claimable(t) := total - curve(t) - claimed(t)$.

in state, we store:

```rust
#[cw_serde]
pub struct VestingPayment {
    pub amount: Uint128,
    pub claimed: Uint128,
    pub curve: Curve,
}
```

and we maintain that $amount(t) := total - claimed(t)$.

$$
claimable(t) := total - claimed(t) - curve(t)
= amount(t) - curve(t)
$$

this is fine, and works and is readable once your head it around it. though i wonder if we stored:

```rust
pub struct VestingPayment {
    pub total: Uint128,
    pub claimed: Uint128,
    pub curve: Curve,
}
```

this might simplify things a little, as the total is more visible and only the `claimed` field is mutable?

contracts/external/cw-payroll-factory/src/msg.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[cw_ownable]
+#[cw_serde]
+pub enum ExecuteMsg {
+    /// Instantiates a new vesting contract that is funded by a cw20 token.
+    Receive(Cw20ReceiveMsg),
+    /// Instantiates a new vesting contract that is funded by a native token.
+    InstantiateNativePayrollContract {
+        instantiate_msg: PayrollInstantiateMsg,
+        code_id: u64,
```
0xekez, line 19: how about specifying the code ID at instantiation time and allowing the owner to update it? that would allow for some social consensus over what contracts are being used.

an example of this potentially causing problems is the rentrancy issue in here from the temp singleton. using that, a malicious contract could set it's instantiator to itself, or another address it can execute messages on behalf of. if we have some social control over the code ID, we close off that type of attack a little.

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,325 @@
...
+    code_id: u64,
+    label: String,
+) -> Result<Response, ContractError> {
+    // Check sender is contract owner if set
+    let ownership = cw_ownable::get_ownership(deps.storage)?;
+    if let Some(owner) = ownership.owner {
+        if info.sender != owner {
+            return Err(ContractError::Unauthorized {});
+        }
+    }
```
0xekez, line 117: should we also verify that the owner is the `instantiate_msg.owner`?

it feels to me the intent is the same and that not being the case is likely to be bad.

contracts/external/cw-payroll-factory/src/msg.rs
JakeHartnell: Oh, GREAT IDEA!.

JakeHartnell:
> really obscure observation about this contract below. i don't think we need to change anything before merging, saying in the hope it is useful / reveals something i don't understand.
> 
> let curve(t) be the value of the vesting curve at, and claimed(t) be total claimed up to time t.
> 
> in this contract, we construct the curve such that claimable(t):=total−curve(t)−claimed(t).
> 
> in state, we store:
> 
> ```rust
> #[cw_serde]
> pub struct VestingPayment {
>     pub amount: Uint128,
>     pub claimed: Uint128,
>     pub curve: Curve,
> }
> ```
> 
> and we maintain that amount(t):=total−claimed(t).
> 
> claimable(t):=total−claimed(t)−curve(t)=amount(t)−curve(t)
> 
> this is fine, and works and is readable once your head it around it. though i wonder if we stored:
> 
> ```rust
> pub struct VestingPayment {
>     pub total: Uint128,
>     pub claimed: Uint128,
>     pub curve: Curve,
> }
> ```
> 
> this might simplify things a little, as the total is more visible and only the `claimed` field is mutable?

Love this!

contracts/external/cw-payroll-factory/src/contract.rs
```diff
@@ -0,0 +1,325 @@
...
+    instantiate_msg: PayrollInstantiateMsg,
+    code_id: u64,
+    label: String,
+) -> Result<Response, ContractError> {
+    // Check sender is contract owner if set
+    let ownership = cw_ownable::get_ownership(deps.storage)?;
+    if let Some(owner) = ownership.owner {
+        if info.sender != owner {
+            return Err(ContractError::Unauthorized {});
+        }
```
0xekez, line 116: currently if an owner is specified, no cw20 vesting contracts can be instantiated as `info.sender` is the cw20 address here.

0xekez (approved):
to the best of my ability, this does not appear to have any issues. i've pushed some commits which improve vesting validation, make vesting contract code IDs governance controlled, add some tests, and fix a bug which disallowed creating cw20 vests if the factory had an owner set.

it would be nice to increase test coverage and remove usage of the singleton, though i don't want to make perfect the enemy of good here, especially because we have DAOs waiting to use this.

contracts/external/cw-payroll-factory/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name ="cw-payroll-factory"
+version = "2.0.0-beta"
+authors = ["Jake Hartnell"]
```
JakeHartnell, line 4: @0xekez / @blue-note feel free to add your names as well! Thanks for the commits! :heart:

contracts/external/cw-payroll-factory/src/msg.rs
```diff
@@ -0,0 +1,88 @@
...
+    Receive(Cw20ReceiveMsg),
+    /// Instantiates a new vesting contract that is funded by a native token.
+    InstantiateNativePayrollContract {
+        instantiate_msg: PayrollInstantiateMsg,
+        label: String,
+    },
+
+    /// Callable only by the current owner. Updates the code ID used
+    /// while instantiating vesting contracts.
+    UpdateCodeId { vesting_code_id: u64 },
```
JakeHartnell, line 25: Very nice. :heart:

blue-note (approved):

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,556 @@
...
+        // Check stake amount is not greater than vesting payment amount
+        if amount > vp.amount {
+            return Err(ContractError::NotEnoughFunds {});
+        }
+
+        // Check that amount has not already been staked
+        vp.amount.checked_sub(vp.staked_amount + amount)?;
+
+        // Update amounts
+        vp.staked_amount += vp.staked_amount.checked_add(amount)?;
```
0xekez, line 363:
```suggestion
        vp.staked_amount = vp.staked_amount.checked_add(amount)?;
```
unless i'm missing something pretty fundamental here? can we write a test to catch this?

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,556 @@
...
+
+    // Update vesting payment with amount of staked rewards
+    VESTING_PAYMENT.update(deps.storage, |mut vp| -> Result<_, ContractError> {
+        // Check stake amount is not greater than vesting payment amount
+        if amount > vp.amount {
+            return Err(ContractError::NotEnoughFunds {});
+        }
+
+        // Check that amount has not already been staked
+        vp.amount.checked_sub(vp.staked_amount + amount)?;
```
0xekez, line 360:
```suggestion
        if vp.amount < vp.staked_amount + amount {
            Err(ContractError::StakingMoreThanVesting {})
        }
```
this might be more clear if we make it an if statement.

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,556 @@
...
+        }
+
+        // Update amounts
+        vp.staked_amount = vp.staked_amount.checked_sub(amount)?;
+        Ok(vp)
+    })?;
+
+    let denom = deps.querier.query_bonded_denom()?;
+
+    // Create message to delegate the underlying tokens
```
0xekez, line 499:
```suggestion
    // Create message to undelegate the underlying tokens
```

contracts/external/cw-vesting/src/contract.rs
```diff
@@ -0,0 +1,556 @@
...
+    let msg = StakingMsg::Undelegate {
+        validator: validator.clone(),
+        amount: Coin { denom, amount },
+    };
```
0xekez, line 503: what happens once this undelegation completes? does the rewards receiver get the tokens?

seeing as we don't re-increment `vp.amount` here once undelegation completes, i assume yes? in which case i feel i could trivially bypass the vesting by staking all the tokens, and then immediately undelegating them.

contracts/external/cw-vesting/src/contract.rs
JakeHartnell: No, this is an embarrassing mistake on my part.

contracts/external/cw-vesting/src/contract.rs
JakeHartnell: > if this contract receives the undelegated funds, shouldn't we be checking the status of unbonding funds in execute_distribute to make sure we don't just error with 'no tokens' if someone distributes and their tokens are in the unbonding or staked state?

We could do this, but it's actually surprisingly difficult. `cosmwasm_std` does not have support for querying unbonding delegations yet. The delegations query doesn't return any info about it: https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.FullDelegation.html

> if the rewards receiver address receives these unbonded tokens, then i feel it is trivial to bypass vesting entirely by bonding and unbonding the tokens.

It does not, the unbonded tokens need to go to the contract since they may not have fully vested.

While we could return a more helpful error message, I feel the current contract is perhaps "good enough" here as the UI could more than compensate for this by showing tokens are unbonding and that there is nothing to claim.

bekauz (approved):
This is so cool. LGTM!

contracts/external/cw-vesting/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Creating a CW20 Vesting
+A cw20 vesting payment can be created using the cw20 [Send / Receive](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/README.md#receiver) flow. This involves triggering a Send message from the cw20 token contract, with a Receive callback that's sent to the vesting contract.
+
+## Distribute payments
+Vesting payments can be claimed continously at any point after the start time by triggering a Distribute message.
+
+Anyone can call the distribute message, allowing for agents such as [Croncat](https://cron.cat/) to automatically trigger payouts.
+
+## Staking native tokens
+This contract allows for underlying native tokens to be staked.
```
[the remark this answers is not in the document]
JakeHartnell: Adding a bunch more info to the README.md

contracts/external/cw-vesting/src/state.rs
```diff
@@ -0,0 +1,308 @@
...
+        Ok(vesting_payment)
+    }
+
+    pub fn get_vested_amount_by_seconds(&self, time: u64) -> Result<Uint128, OverflowError> {
+        let vesting_funds = self.vesting_schedule.value(time);
+        self.amount.checked_sub(vesting_funds)
+    }
+}
+
+/// A map of vesting payments (ID, VestingPayment)
```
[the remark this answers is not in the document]
JakeHartnell: Lol, will update.

contracts/external/cw-vesting/src/contract.rs
0xekez: > I feel the current contract is perhaps "good enough" here as the UI could more than compensate for this by showing tokens are unbonding and that there is nothing to claim.

thanks for the detailed explanation. i agree with this take. i also recall this being a pain to query from a conversation with @NoahSaso about the DAO treasury cards. validating this in the FE sounds good to me.

contracts/external/cw-vesting/src/contract.rs
0xekez: happens to everyone haha

contracts/external/cw-payroll-factory/src/error.rs
```diff
@@ -0,0 +1,31 @@
...
+    OwnerMissmatch {
+        actual: Option<String>,
+        expected: Option<String>,
+    },
+
+    #[error("{0}")]
+    ParseReplyError(#[from] ParseReplyError),
+
+    #[error("An unknown reply ID was received.")]
+    UnknownReplyID {},
```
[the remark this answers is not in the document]
JakeHartnell: Done.

contracts/external/cw-payroll-factory/Cargo.toml
JakeHartnell: Added.

0xekez (approved):

PR #609, opened 2023-01-08
Voting pow fund distribution

### LINKED ISSUE #266: Implement a DAO liquidation contract.
How does a DAO die? Likely by liquidating it's treasury to governance participants.

This liquidation contract should accept a voting module contract, a height, and some funds (cw20, cw721, or native). It should then disperse these assets to members of the voting contract according to their staked proportion.

For example, if a DAO with 100 Juno and 100 members each with equal voting power would like to die it may send its treasury to the liquidation contract. All members at the time of liquidation would then be able to claim 1 Juno.

bekauz:
Closes #266 

On instantiation,  `funding_period` defines how many blocks there are to fund the distributor contract. After that many blocks, distributor switches into claiming mode. Users can then submit individual cw20/native claims, or just claim all of their share at once.  

User share is determined by their respective fraction of the voting power at the time of instantiating the distributor.
After the distributor contract is funded with tokens from the DAO treasury, each member is entitled to their share of every token in the distributor contract.

The cw20 admin can redistribute any unclaimed tokens. In that case, all performed cw20 and native claims are recorded and subtracted from the cw20 and native balances available for claiming. Any previous user claims are then nullified, and everyone is entitled to their share again.

I feel like this still has potential to improve, but I'm particularly interested in thoughts on:
- Method to withdraw funds out of the distributor contract. Any reason to have it?
- Any alternatives to re-distribution of unclaimed funds. There are many possibilities here, but how likely is it that anything other than redistribution would win the vote?
- Should redistribution work any differently? Perhaps exclude those who did not claim their initial allocation (would speed up the liquidation I suppose)?


0xekez (approved):
this is soooooo slick. i love how functional the code is. have two very small nits, but otherwise lg2m!

contracts/external/cw-fund-distributor/.gitignore
```diff
@@ -0,0 +1,16 @@
...
+
+# Text file backups
+**/*.rs.bk
+
+# macOS
+.DS_Store
+
+# IDEs
+*.iml
+.idea
```
JakeHartnell, line 16: We a  .gitignore in the root directory, probably don't need this one.

contracts/external/cw-fund-distributor/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "cw-fund-distributor"
+version = "0.1.0"
+authors = ["bekauz <bekauz@protonmail.com>"]
+edition = "2021"
+repository = "https://github.com/DA0-DA0/dao-contracts"
+description = "A CosmWasm contract for distributing funds to DAO members"
```
JakeHartnell, line 7:
```suggestion
description = "A CosmWasm contract for distributing funds to DAO members based on voting power."
```

contracts/external/cw-fund-distributor/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Possible use cases may involve:
+- Dissolving a DAO and distributing its treasury to members prior to shutting down
+- Distributing funds among DAO members
+- Funding subDAOs
+
+## Funding Period
+
+Contract is instantiated with a `funding_period` - a time duration that should suffice 
+to move the funds to be distributed into the distributor contract.
```
JakeHartnell, line 15: Why not just instantiate the contract with funds, similar to `cw-vesting`?

Seems like it would be less complicated and easier to setup? https://github.com/DA0-DA0/dao-contracts/pull/608

JakeHartnell: Ah, I see. It makes it so you can send whatever to it to be distributed. Hmmmmmmm...

contracts/external/cw-fund-distributor/src/contract.rs
```diff
@@ -0,0 +1,628 @@
...
+    FUNDING_PERIOD_EXPIRATION.save(deps.storage, &funding_expiration_height)?;
+
+    // validate the contract and save it
+    let voting_contract = deps.api.addr_validate(&msg.voting_contract)?;
+    VOTING_CONTRACT.save(deps.storage, &voting_contract)?;
+
+    let total_power: voting::TotalPowerAtHeightResponse = deps.querier.query_wasm_smart(
+        voting_contract.clone(),
+        &voting::Query::TotalPowerAtHeight {
+            height: Some(env.block.height),
```
JakeHartnell, line 49: We may want to make the distribution height a parameter for the `InstantiateMsg`.

imagine there is a proposal to dissolve a DAO with a large amount assets and a liquid token. Since everyone can see the proposal go live, they can rush to stake their tokens (or acquire some) before the proposal passes and is executed.

Likely not a huge issue, but also easy enough to add an optional `distribution_height` parameter.

contracts/external/cw-fund-distributor/src/contract.rs
```diff
@@ -0,0 +1,628 @@
...
+                .unwrap_or_default();
+            match balance {
+                Some(old_amount) => NATIVE_BALANCES.save(
+                    deps.storage,
+                    coin.denom.clone(),
+                    &old_amount.checked_add(coin.amount).unwrap(),
+                ),
+                None => NATIVE_BALANCES.save(deps.storage, coin.denom.clone(), &coin.amount),
+            }
+            .unwrap();
```
JakeHartnell, line 143: Would be nice to do this without `unwrap()`s.

contracts/external/cw-fund-distributor/src/contract.rs
bekauz: That's a great point. Initially I had it as a parameter but ended up removing it. Will get it back in there!

contracts/external/cw-fund-distributor/src/state.rs
```diff
@@ -0,0 +1,22 @@
+use cosmwasm_std::{Addr, Uint128};
+use cw_storage_plus::{Item, Map};
+use cw_utils::Expiration;
+
+// block height for distribution snapshot
```
JakeHartnell, line 5: For auto generated docs it's best to use `///` here so that the comments show up in generated docs.

contracts/external/cw-fund-distributor/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+pub mod contract;
```
JakeHartnell, line 1:
```suggestion
#![doc = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/README.md"))]

pub mod contract;
```

This includes the readme in generated rust documentation.

JakeHartnell (approved):
Made a few comments and suggestions but otherwise LGTM.

contracts/external/cw-fund-distributor/src/contract.rs
0xekez:
```suggestion
            match balance {
                Some(old_amount) => NATIVE_BALANCES.save(
                    deps.storage,
                    coin.denom.clone(),
                    &old_amount.checked_add(coin.amount).unwrap(),
                ),
                None => NATIVE_BALANCES.save(deps.storage, coin.denom.clone(), &coin.amount),
            }?;
```

bekauz: > Would be nice to do this without `unwrap()`s.

I tried but the issue is with `?` operator only being usable with `Result` or `Option`. I feel like the `checked_add` should never overflow given the `U128` range, but I suppose we could do something like this instead?
```rust
    let mut attributes: Vec<(String, String)> = Vec::new();
    for coin in info.funds {
        if coin.amount > Uint128::zero() {
            let balance = NATIVE_BALANCES
                .may_load(deps.storage, coin.denom.clone())
                .unwrap_or_default();

            let mut new_amount = coin.amount;
            // add any previous balances
            if let Some(previous_amount) = balance {
                new_amount.checked_add(previous_amount)
                    .map_err(|e| ContractError::Std(StdError::from(e)))?;
            };
            NATIVE_BALANCES.save(
                deps.storage,
                coin.denom.clone(),
                &new_amount,
            )?;
        }
    }

```

0xekez: sometimes imperative > functional. this looks great

0xekez:
@bekauz looks like we need to run `cargo update` / `cargo fmt` to get CI to pass. would you mind doing that when you get a chance? if you're too busy i can also push to your branch.

JakeHartnell:
Looks good aside from the merge conflicts. Let's get this in?

JakeHartnell (approved):
LGTM!

PR #618, opened 2023-01-13
Condorcet Voting

0xekez:
this is a proposal module which selects the [Condorcet winner](https://en.wikipedia.org/wiki/Condorcet_winner_criterion#:~:text=The%20Condorcet%20winner%20is%20the,candidates%20in%20a%20plurality%20vote.) from a set of ranked choice ballots. ballots are lists of candidates sorted by order of preference.

this module does not implement the pre-propose and does not support proposal deposits. my hope is that we can land and audit this, focusing on the voting logic, and then follow up with pre-propose.


contracts/proposal/dao-proposal-condorcet/src/msg.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use dao_macros::proposal_module_query;
+
+use crate::config::UncheckedConfig;
+
+pub type InstantiateMsg = UncheckedConfig;
+
+#[cw_serde]
+pub struct Choice {
+    pub msgs: Vec<CosmosMsg<Empty>>,
```
blue-note, line 12: what about choice title and description? 

also todo later or now: we can reconcile this type with the multiple choice option. I like "choice" as a title, simpler

contracts/proposal/dao-proposal-condorcet/src/contract.rs
```diff
@@ -0,0 +1,279 @@
...
+    let config = CONFIG.load(deps.storage)?;
+
+    let id = next_proposal_id(deps.storage)?;
+    let total_power = get_total_power(deps.as_ref(), &dao, None)?;
+
+    if choices.is_empty() {
+        return Err(ContractError::ZeroChoices {});
+    }
+
+    let none_of_the_above = Choice { msgs: vec![] };
```
blue-note, line 121: we'd also want to check if there are at least two choices, and that choice length is below some bound

contracts/proposal/dao-proposal-condorcet/src/contract.rs
```diff
@@ -0,0 +1,279 @@
...
+    let mut choices = choices;
+    choices.push(none_of_the_above);
+
+    let tally = Tally::new(
+        choices.len() as u32,
+        total_power,
+        env.block.height,
+        config.voting_period.after(&env.block),
+    );
+    TALLYS.save(deps.storage, id, &tally)?;
```
blue-note, line 131: super nit: call the state variable TALLY if its one tally?

blue-note (approved):

contracts/proposal/dao-proposal-condorcet/src/contract.rs
0xekez: i think only one choice is actually fine (`!is_empty()` above) because in that case it becomes exactly the same as a `dao-proposal-single` election. down the line we might even want to replace proposal-single with this.

contracts/proposal/dao-proposal-condorcet/src/contract.rs
0xekez: great point. thanks

contracts/proposal/dao-proposal-condorcet/src/m.rs
```diff
@@ -0,0 +1,382 @@
...
+        debug_assert!(x != y);
+        if x < y {
+            self.decrement((y, x), amount)
+        } else {
+            let i = self.index((x, y)) as usize;
+            self.cells[i] = self.cells[i].increment(amount)
+        }
+    }
+
+    pub fn decrement(&mut self, (x, y): (u32, u32), amount: Uint128) {
```
blue-note, line 111: since decrement could also increment, and increment could also decrement, could there be a more aptly named top-level function?

contracts/proposal/dao-proposal-condorcet/src/msg.rs
0xekez: as discussed in Discord: we're going to pass on making proposals human-readable for now. we're interested in writing a secure voting system first. i've added a note to the README to this effect.

contracts/proposal/dao-proposal-condorcet/src/tally.rs
```diff
@@ -0,0 +1,117 @@
...
+    ///  - `vote` a list of candidates sorted in order from most to
+    ///    least favored
+    ///  - `power` the voting power of the voter
+    ///
+    /// Invariants:
+    ///
+    /// - Voter has not already voted.
+    /// - Tally is not expired.
+    pub fn add_vote(&mut self, vote: Vote, power: Uint128) {
+        for (index, preference) in vote.iter().enumerate() {
```
blue-note, line 85: I'm a bit confused, so index is the rank and preference is the candidate? (that seems to be what "`vote` a list of candidates sorted in order from most least favored") is implying

but then iterating through defeats starting from higher ranks on line 89 seems to contradict that?

contracts/proposal/dao-proposal-condorcet/src/m.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    /// Computes statistics about M which are used to determine if a
+    /// proposal has passed or may be rejected early.
+    ///
+    /// Code comments refer to this proof of conditions for early
+    /// rejection:
+    ///
+    /// https://github.com/DA0-DA0/dao-contracts/wiki/Proofs-of-early-rejection-cases-for-Condorcet-proposals
+    pub fn stats(&self, power_outstanding: Uint128) -> Stats {
```
blue-note, line 128: this function is goat

contracts/proposal/dao-proposal-condorcet/src/m.rs
```diff
@@ -0,0 +1,382 @@
...
+            let mut max_negative = Uint128::zero();
+            for row in 0..n {
+                if row != col {
+                    match self.get((col, row)) {
+                        Cell::Positive(p) => {
+                            if p < min_margin {
+                                min_margin = p
+                            }
+                        }
+                        Cell::Negative(n) => {
```
blue-note, line 143: super nit: want to call this something other than n to avoid confusion
                        // with the above n?

contracts/proposal/dao-proposal-condorcet/src/m.rs
```diff
@@ -0,0 +1,382 @@
...
+                            if n > max_negative {
+                                max_negative = n;
+                            }
+                            distance_from_positivity += Uint256::from(n) + Uint256::one();
+                        }
+                        Cell::Zero => distance_from_positivity += Uint256::one(),
+                    }
+                }
+            }
+            if distance_from_positivity.is_zero() {
```
blue-note, line 153: makes sense to return as soon as we've found a positive column because there can only be one (might be good to include that in a comment here)

contracts/proposal/dao-proposal-condorcet/src/tally.rs
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+        self.power_outstanding -= power;
+        self.winner = self.winner();
+    }
+
+    fn winner(&self) -> Winner {
+        match self.m.stats(self.power_outstanding) {
+            Stats::PositiveColumn { col, min_margin } => {
+                if min_margin > self.power_outstanding {
```
blue-note, line 100: this makes sense because if you can't record enough losses to bring this column into the negative, it is an undisputed winner.
            And I think it is proven that there can only ever be one condorcet winner,
                ie only one column with all positive, right?

contracts/proposal/dao-proposal-condorcet/src/tally.rs
0xekez: so given a ballot like

```
[a, b, c]
```

for `c`, this method does:

```
for defeat in 0..2:
  self.m.decrement((c, ballot[defeat]), power)
```

which translates to:

```
self.m.decrement((c, a), power)
self.m.decrement((c, b), power)
```

because of the symmetry properties maintained by the `M` type decrementing `(c, a)` also increments `(a, c)` so we effectively record "c lost to a" and "a beat c" in that single line.

does that help explain what this is doing?

contracts/proposal/dao-proposal-condorcet/src/tally.rs
0xekez: yeah the symmetry means that if there is an all-positive column, then there is an all-negative row (ignoring the x=y line).

contracts/proposal/dao-proposal-condorcet/src/m.rs
0xekez: 🐐

contracts/proposal/dao-proposal-condorcet/src/m.rs
0xekez: i can't really think of one. they both do subtly different thing, and i'm pretty sure they just have to be mutually recursive like this. happy to add one though if you have a suggestion?

contracts/proposal/dao-proposal-condorcet/src/m.rs
0xekez: lol great point i didn't even realize there were two `n` variables in scope

contracts/proposal/dao-proposal-condorcet/src/m.rs
0xekez: good idea. added a comment

contracts/proposal/dao-proposal-condorcet/src/tally.rs
0xekez: > but then iterating through defeats starting from higher ranks on line 89 seems to contradict that?

another way of thinking of this: those ranked higher than you are your "defeats" for that vote

contracts/proposal/dao-proposal-condorcet/src/m.rs
JakeHartnell: IMO, it's fine for now, may just add a quick comment about it?

NoahSaso (approved):
you guys are freaking legends :0

JakeHartnell (approved):
Wow, this is epic! Didn't think we would be so close to on-chain ranked choice voting so soon.

PR #624, opened 2023-01-19
Tidy up toml files for publishing.

0xekez:
This does some basic tidying to our `Cargo.toml` files so that we can publish crates.

Unfortunately it is not possible to [publish a crate with a git dependency](https://github.com/rust-lang/cargo/issues/6738), though we can still publish the good packages.


bekauz (approved):
clean

NoahSaso (approved):

PR #625, opened 2023-01-19
crates.io does not recognize bsd-3.

0xekez:
apparently this does not get checked when doing a publish dry run, so here we are.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -583,7 +583,7 @@ dependencies = [
 
 [[package]]
 name = "cw-denom"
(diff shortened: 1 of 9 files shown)


blue-note (approved):

JakeHartnell (approved):
lol

0xekez:
thanks for the reviews. i'm actually not going to merge this as i'm going to do a larger PR which includes some renaming to get these up on crates.io and avoid name conflicts. sorry to rug you :(

PR #634, opened 2023-01-27
Handle callback messages from preproposal contracts

ismellike:
This would be useful for a prepropose module that wants to propagate a message back to the DAO like setting an item with its address.


0xekez:
thanks for putting this up! could we do this the same way we do it in proposal single and forward the callback to the core module?

https://github.com/DA0-DA0/dao-contracts/blob/403f5cf1613059fdcd987944c6d69e2942f8ff78/contracts/proposal/dao-proposal-single/src/contract.rs#L979-L990

PR #640, opened 2023-02-02
chore(ci): action release-contracts

albttx:
following: [this tweet](https://twitter.com/albttx/status/1621203943773687809) i realized DAO-DAO contracts wan't build though the CI,

Here is a Github action for that

### DIFF
--- a/.github/workflows/release-contracts.yml
+++ b/.github/workflows/release-contracts.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    container: cosmwasm/workspace-optimizer:0.12.11
+    container: cosmwasm/workspace-optimizer:0.12.13
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/release-contracts.yml
+++ b/.github/workflows/release-contracts.yml
@@ -1,0 +1,51 @@
+name: Release contracts
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - 'v*'
+    branches:
+      - main
+      - ci/release-contracts
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    container: cosmwasm/workspace-optimizer:0.12.11
+    steps:
+      - uses: actions/checkout@v3
+
+      # tar is required for cargo cache
+      - run: apk add --no-cache tar
+
+      - name: Set up cargo cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/            
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Compile contracts
+        timeout-minutes: 30
+        run: optimize_workspace.sh .
+
+      - name: Upload contracts
+        uses: actions/upload-artifact@v3
+        with:
+          name: contracts
+          path: artifacts/
+
+      - name: release
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: artifacts/*
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


albttx:
currently running on my fork: https://github.com/albttx/dao-contracts/actions/runs/4077628245/jobs/7026877538

albttx:
need to add something like : https://github.com/xresloader/upload-to-github-release to upload result in released assets

albttx:
Do you want to have this asset build on `main` branch too ?

0xekez:
hey, thanks for this :) @albttx could you help me understand a little what the release flow would become if we had this? where can i find this uploaded .wasm artifacts so i can link to them in release notes?

albttx:
You can find the artifact in the build, check on my fork https://github.com/albttx/dao-contracts/actions/runs/4077628245

We could also use another action to automatically generate the release and upload the artifacts in the released assets

JakeHartnell (approved):
Really cool!

albttx:
Done! you can take a look on my fork the releases

https://github.com/albttx/dao-contracts/releases/tag/v4.1.0-test

The workflow stay the same, you can either tag with the git CLI or via github interface and release, automatically the CI will start and push the assets

albttx:
For the `main` branch, you'll have all the .wasm  files compiled in the pipeline: see https://github.com/albttx/dao-contracts/actions/runs/4084609040

albttx:
@0xekez this is ready, feel free if you have any questions on this

.github/workflows/release-contracts.yml
```diff
@@ -0,0 +1,51 @@
...
+    tags:
+      - 'v*'
+    branches:
+      - main
+      - ci/release-contracts
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    container: cosmwasm/workspace-optimizer:0.12.11
```
0xekez, line 17: i'm a little nervous that this will become out of date when a new workspace-optimizer release is made. ideally we'd be able to synchronize this with the version used in the `workspace-optimize`.

i wonder if we could integrate this with our existing [integrationt test workflow](https://github.com/DA0-DA0/dao-contracts/blob/main/.github/workflows/integration_tests.yml#L61) so to keep that in sync?

for example, the `just integration-test` command there builds the contracts using the `just workspace-optimize` command, so if we moved the upload step here over to the job i think this would get us uploads, and make sure our versions are staying consistent.

albttx: The main issue with `just workspace-optimize` is that it's running a docker container, it's would require to use docker in docker to build...
It's didn't feel an appropriate solution.

Maybe we can use ` cw-optimizoor`, but i believe `cosmwasm/workspace-optimizer` is the prefered tool for compiling, specially on a tag.


IMHO, having the tag is not a big problem... or we can use `:latest` ... but i doubt it's a good solution

albttx:
So what's next ? shoud we merge or not ?

cc: @JakeHartnell @0xekez  ?

JakeHartnell:
> So what's next ? shoud we merge or not ?
> 
> cc: @JakeHartnell @0xekez ?

I'm cool to merge if @0xekez is.

.github/workflows/release-contracts.yml
0xekez: talking with @NoahSaso, having `.wasm` files available from CI is useful and i see the problem with docker within docker.

.github/workflows/release-contracts.yml
```diff
@@ -0,0 +1,51 @@
...
+          name: contracts
+          path: artifacts/
+
+      - name: release
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: artifacts/*
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
0xekez, line 51: @albttx, could you help me understand this line? from reading [the package](https://github.com/softprops/action-gh-release) it seems like this will automatically create releases, or will I still be able to cut them manually? i like to do a manual check of the checksums before releasing so I'm not sure automatically creating releases when tags are pushed is what we want.

0xekez (commented):
apologies for being so slow to get to this and for all the questions. i'm just trying to make sure that we're only storing `.wasm` artifacts that actually match what we want to put on chain.

@albttx what would you think about writing some docs about this? this could go in `CONTRIBUTING.md`. it would be helpful to have in writing:

1. q: when does this action need to be updated? a: whenever we bump the workspace optimizer version used in CI.
2. q: where can one download the artifacts generated by this?

I also have a question about automatic releases. I'm not sure we want to do that per (1) above where we're not guaranteed that this is compiling the using the correct workspace-optimizer version.

.github/workflows/release-contracts.yml
albttx: Hey,
If you want to see an example take a look at my fork https://github.com/albttx/dao-contracts/releases/tag/v4.1.0-test

I gave you permission, so try to generate a new release and you'll see the wasm contracts + checksum will be automatically published into the release.

Yes, every push of a git tag will automatically generate a release, but you can also just generate a release like you used to do! Try on my fork, you'll see, it's life changing! :)

albttx:
@JakeHartnell  @0xekez up ?

.github/workflows/release-contracts.yml
JakeHartnell:
```suggestion
    container: cosmwasm/workspace-optimizer:0.12.13
```

This is the version we have currently in the repo, will bump in the future.

JakeHartnell:
Me like. :heart:

PR #643, opened 2023-02-07
removing empty files

bekauz:
seem to be coming from `clippy` because of "unnecessary variables", however the fix is not as simple as it suggests


0xekez (approved):
🫡

JakeHartnell (approved):

PR #646, opened 2023-02-08
Dao migrator extend multiple proposal modules

Art3miX:
Now migrator is able to migrate DAOs that have the edge case of multiple proposal modules, same length of proposal params should be passed as well.

Please note the changes in the accepted params (wrapped the proposal migration params in another type)


JakeHartnell (approved):
LGTM. :heart:

0xekez (approved):
AMAZING. thank you.

PR #648, opened 2023-02-12
Integration test to measure max NFTs mintable in a single proposal.

0xekez:
based on tuning on my local setup, the current number is 55.

### DIFF
--- a/ci/integration-tests/src/helpers/helper.rs
+++ b/ci/integration-tests/src/helpers/helper.rs
@@ -1,17 +1,20 @@
 use super::chain::Chain;
 use anyhow::Result;
 use cosm_orc::orchestrator::SigningKey;
-use cosmwasm_std::{to_binary, Decimal, Empty, Uint128};
+use cosmwasm_std::{to_binary, CosmosMsg, Decimal, Empty, Uint128};
 use cw20::Cw20Coin;
 use cw_utils::Duration;
 use dao_core::query::DumpStateResponse;
 use dao_interface::{Admin, ModuleInstantiateInfo};
 use dao_voting::{
     deposit::{DepositRefundPolicy, DepositToken, UncheckedDepositInfo},
-    pre_propose::PreProposeInfo,
+    pre_propose::{PreProposeInfo, ProposalCreationPolicy},
     threshold::PercentageThreshold,
     threshold::Threshold,
+    voting::Vote,
 };
+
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -63,3 +63,14 @@ workspace-optimize:
 		--mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
 		--platform linux/amd64 \
 		cosmwasm/workspace-optimizer:0.12.11
+
+workspace-optimize-arm:
```
JakeHartnell, line 67: :heart:

JakeHartnell (approved):
LGTM.

justfile
```diff
@@ -63,3 +63,14 @@ workspace-optimize:
...
+    docker run --rm -v "$(pwd)":/code \
+        --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+        --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+        --platform linux/arm64 \
+        cosmwasm/workspace-optimizer-arm64:0.12.11
+    # rename the wasm files so that they work with cosm-orc
+    # integration test contract names (which are expected to be the
+    # contract name, without the aarch64 postfix).
+    for file in artifacts/*-aarch64.wasm; do mv -- "$file" "${file%-aarch64.wasm}.wasm"; done
```
de-husk, line 76: I tried to handle this situation here:  https://github.com/de-husk/cosm-orc/blob/main/src/orchestrator/cosm_orc.rs#L153-L157

I wonder why it's not working in this case?

ci/integration-tests/src/helpers/helper.rs
```diff
@@ -97,12 +100,194 @@ pub fn create_dao(
...
+	.orc
+	.query(
+	    "dao_proposal_single",
+	    &dao_proposal_single::msg::QueryMsg::ProposalCreationPolicy {}
+	).unwrap()
+	.data()
+	.unwrap()
+    else {
+	panic!("expected pre-propose module")
+    };
```
de-husk, line 130: I like this pattern

de-husk (approved):

justfile
0xekez: possible it was failing for unrelated reasons when i added this. i'll test before merging.

bekauz (approved):

justfile
0xekez: @de-husk confirmed that your solution works and my change was not needed. thanks for pointing this out!

PR #653, opened 2023-02-15
Fix migrator

Art3miX:
Fixed the issue, and added a specific test to the error we were getting

### DIFF
--- a/contracts/external/dao-migrator/src/contract.rs
+++ b/contracts/external/dao-migrator/src/contract.rs
@@ -1,4 +1,4 @@
-use std::env;
+use std::{env, collections::HashSet};
 
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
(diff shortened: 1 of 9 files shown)


contracts/external/dao-migrator/src/contract.rs
```diff
@@ -88,18 +88,23 @@ fn execute_migration_v1_v2(
     }
 
     // List of code ids pairs we got and the migration msg of each one of them.
-    let proposal_pairs: Vec<CodeIdPair> = migration_params
+    let proposal_pairs: Vec<(String, CodeIdPair)> = migration_params
         .proposal_params
```
0xekez, line 92: just a little thing, but could we add a check that this does not have duplicate `addr` values?

0xekez (approved):
looks good! nice work on the alphabetical issue test. would be nice to deduplicate the proposal migration list, and then this is good to go.

JakeHartnell (approved):

contracts/external/dao-migrator/src/contract.rs
Art3miX:
```rust 
let mut uniq = HashSet::new();
if migration_params.proposal_params.iter().all(|(addr, _)| uniq.insert(addr)) {
    return Err(ContractError::DuplicateProposalParams);
}
```

Probably the nicest way I found to check for duplicates, not sure if HashSet is supported on the blockchain (it should in theory)

NoahSaso (approved):
woooo yay let's get this 🍞

contracts/external/dao-migrator/src/contract.rs
0xekez: looks great!

Issue #543, opened 2022-11-19
Make native staked DAO-DAO voting, have SDK hooks that call into it for stake changes


ValarDragon:
DAO DAO voting for natively staked coins can be more powerful, if we make an SDK module that just listens for all delegate, undelegate, and redelegate hooks. Then on each of these hooks, calls a sudo message to a pre-configured contract.

This would let us have a more powerful wrapper for governance! They could now compute the current semantic of "Voting power = stake at end of voting period", with no iteration at end time for instance. (The only iteration needed, would be to reclaim state, but we can have a separate process for that. Espec. if we make a gas token like dynamic)

Or we could switch semantics to "voting power = stake at beginning of voting period", or my favorite "voting duration = k days. `Voting power = sum of stake_at_time_i * voting decision_at_time_i` for times in the final day of voting" (to protect against last second vote attacks)

If we make an SDK module for communicating stake change events during the voting period, then we can unlock all of these capabilities for DAO DAO voting.

JakeHartnell:
> DAO DAO voting for natively staked coins can be more powerful, if we make an SDK module that just listens for all delegate, undelegate, and redelegate hooks. Then on each of these hooks, calls a sudo message to a pre-configured contract.

Woah. This would be amazing. I think it would be useful for many other things as well!

FWIW, we do allow for native staked voting in DAO DAO V2... though I'm not 100% convinced it's the best implementation: https://github.com/DA0-DA0/dao-contracts/tree/main/contracts/voting/cwd-voting-staking-denom-staked

PR #655, opened 2023-02-16
Add Native Token Staking Contract

mccallofthewild:
Drop-in `sdk.Coins` staking contract compatible with cw20-stake. All tests passing. This seems like a cleaner approach than merging the proposal and staking modules.

This makes progress on syncing DAO DAO with the protocol level as referenced in #543. Easiest way to do this while accounting for slashing is by nesting validator staking contracts under a root staking contract. 

I also have something that automatically provisions staking reward distributors for any token. At the protocol level it would make airdrops extremely simple.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2420,6 +2420,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "native-stake"
+version = "2.0.3"
+dependencies = [
+ "anyhow",
+ "cosmwasm-schema",
+ "cosmwasm-std",
+ "cosmwasm-storage",
+ "cw-controllers 0.16.0",
+ "cw-multi-test",
+ "cw-paginate 2.0.3",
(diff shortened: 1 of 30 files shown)


0xekez:
hello! thanks for the contribution. @mccallofthewild would you mind expanding on what you'd like to use this contract for? i am very happy to review, just curious so i can situate it.

we also have a native token staking contract here: https://github.com/DA0-DA0/dao-contracts/tree/main/contracts/voting/dao-voting-native-staked, though it does not have staking hooks so i could see why this would be preferable. have you seen that one? i'd be interested in your thoughts on if that's worth improving, or if we should replace it with this.

---

to fix the lint issue:

```
just lint
```

to fix the typescript generation:

```
just gen
```

JakeHartnell:
Closing because it's stale and we're going to do a different approach.

PR #668, opened 2023-03-06
Update cardinality when redelegating.

0xekez:
Resolves issue one in the cw-vesting Oak Security audit.


JakeHartnell (approved):
Ah, surprised we missed this. smh

Thanks for the fix. :pray:

NoahSaso (approved):

Issue #671, opened 2023-03-06
Add info queries to `cw-vesting`


NoahSaso:
this is my wishlist to improve frontend UX for the vesting contract:

- `vested`: the amount of money that has been unlocked; it may be claimed, staked, unstaking, or any combination of the possible states of the money. i just want the value in the curve that has been vested
- `total`: the total amount of money being vested. amount vested should eventually reach this value
- `duration`: the duration after which `vested` equals `total`, i.e. all money is vested.

making these changes would completely remove the need for the frontend to know what curves are supported, which means we would automatically support vesting contracts in the future with any curve. it also greatly simplifies the math on the frontend, and we all know JS's `Number` math is scary.

i think the first two values should not change even when slashing occurs. i can always subtract slashes from the total when displaying it, but i think it's good to have a distinction between how much the recipient was given vs how much they will actually get.

from those two, i can also calculate the amount that is available to stake with `total - claimed - staked - slashed`.

also, not a big deal, but standardizing all of the durations/timestamps to be in one unit would be nice. second-precision is sufficient.

0xekez:
thank you for this issue my sir 🙏

PR #673, opened 2023-03-07
Improve queryability of vesting contract

0xekez:
In the first commit, this adds queries to access the stake tracker as mentioned by Oak in issue 6. The second commit adds the queries requested by Noah in #671.

Edit: this also removes the typescript generation CI job as dao-dao-ui doesn't use it.


contracts/external/cw-vesting/src/msg.rs
```diff
@@ -193,6 +194,27 @@ pub enum QueryMsg {
     Distributable {
         /// The time, as a unix timestamp in seconds, or none to use
```
NoahSaso, line 195:
```suggestion
        /// The time, as a unix timestamp in nanoseconds, or none to use
```

comment needs updating. i think `Timestamp`s expect `nanoseconds`

contracts/external/cw-vesting/src/vesting.rs
```diff
@@ -420,6 +437,19 @@ impl Vest {
         self.status = Status::Canceled { owner_withdrawable };
         self.vested = Curve::Constant { y: self.vested(t) };
     }
+
+    /// Gets the duration of the vest. For constant curves, `None` is
+    /// returned.
+    pub fn duration(&self) -> Option<Duration> {
```
NoahSaso, line 443: if this will always be a `Duration::Time`, is there a reason to use a `Duration` type instead of a `u64` of the seconds?

on the frontend it adds extra steps (need to check `'time' in duration`) and an extra error handler to make sure it isn't `Duration::Blocks`

NoahSaso (changes requested):
sick! just a question and a comment fix

contracts/external/cw-vesting/src/msg.rs
0xekez: thanks!

contracts/external/cw-vesting/src/vesting.rs
0xekez: 🙏 bec15c70ad7095ab5fef9394618405275f32e2db

great feedback to hear. thanks.

NoahSaso (approved):
sweet 😎 thanks!!

JakeHartnell (approved):
LGTM

Issue #649, opened 2023-02-12
Allow the token swap contract to execute a message on completion


0xekez:
In the case of native tokens being swapped, the `cw-token-swap` contract should be able to execute a message on completion with those funds sent in the `funds` field. in the case of cw20 tokens, the same concept should apply but via a cw20 `ExecuteMsg::Send`.

this would allow the token swap contract to compose with the vesting contract and allow tokens to be swapped into vested ones.

JakeHartnell:
Wow, this is sick. Love it.

PR #676, opened 2023-03-09
Update token swap

Art3miX:
This PR implements #649 feature

## General design:
When funding the contract you can specify what messages you want to execute on completion of the swap.
in case of promised cw20, you can provide any binary message that will be fed as is into `cw20::execute::Send` message.
in case of native token, you can execute any amount of messages from the `AcceptedMessages` enum (bank send and burn, wasm execute and instantiate).

All types moved to `types.rs`.

This allows for a whole bunch of different use cases such as:
1. Fund a different swap contract
2. instantiate vesting contract
3. burn half native tokens and send other half to someone else

## Improvements / questions
1. Currently promised cw20 tokens can only be sent, while native token can do more, should we also add more messages to cw20 cases? We can add Transfer and Burn as well.


contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,399 @@
...
+#[cw_serde]
+pub enum TokenInfo {
+    /// A native token.
+    Native { denom: String, amount: Uint128 },
+    /// A cw20 token.
+    Cw20 {
+        contract_addr: String,
+        amount: Uint128,
+    },
+}
```
0xekez, line 247: How about renaming this to `SwapInfo` and taking the messages to be executed as fields.

```rust
pub enum SwapInfo {
    /// A native token.
    Native { denom: String, amount: Uint128, on_completion: Vec<NativeSendMsg> },
    /// A cw20 token.
    Cw20 {
        contract_addr: String,
        amount: Uint128,
        on_completion: Vec<Cw20SendMsgs>
    },
}
```

This gives us two advantages:

1. When creating a swap agreement, all of the terms are known at instantiation time (as opposed to providing messages during funding).
2. We don't need `ContractError::InvalidSendMsg` as we get correctness by construction here.

The checked version of this could also contain the exact `CosmosMsg` to return, as yielded by `into_checked_cosmos_msgs `.

contracts/external/cw-token-swap/src/contract.rs
```diff
@@ -59,6 +61,7 @@ struct CounterpartyResponse<'a> {
     pub counterparty: CheckedCounterparty,
     pub other_counterparty: CheckedCounterparty,
     pub storage: Item<'a, CheckedCounterparty>,
+    pub other_storage: Item<'a, CheckedCounterparty>,
```
0xekez, line 64: nit: having only the counterparties storage in this struct has some safety guarantees as it means that code with this object can't mutate the other counterparty. not specifying the messages to be executed in `Fund` should mean that we don't need this.

contracts/external/cw-token-swap/src/tests.rs
```diff
@@ -35,6 +37,15 @@ fn cw20_contract() -> Box<dyn Contract<Empty>> {
...
 }
 
+fn cw_vesting() -> Box<dyn Contract<Empty>> {
+    let contract = ContractWrapper::new(
+        cw_vesting::contract::execute,
+        cw_vesting::contract::instantiate,
+        cw_vesting::contract::query,
+    );
+    Box::new(contract)
+}
```
0xekez, line 47: this is an awesome test!

Art3miX:
@0xekez As always thanks for yet another great review!!!

I done the strategy change based on your 2 comments, and modified the tests, here is the new strategy.

We take all msgs we want to execute in the instantiate msg, the msgs are based on the promised token, so you can think of them as "What msgs I need to execute with those funds", so if someone promise native token, you set the `on_completion` field to empty vec and it will send it to the other party, or specify a specific msgs to execute (like init some other contract), same idea for cw20 token.

We validate the msgs by type enforcement and also by making sure the total funds sent in all msgs matches the total funds promised or else it will error.

1 **security concern** with this tho, because the msgs are not sent on funding, the party who instantiate this contract must be a trusted party, or else a malicious party can execute any msg he wants with your funds, and the only way to confirm this will be by querying the statues and verifying the specified msgs.

One thing that helps security is that those msgs are immutable, which also means that if you specify a wrong amount or msg you would need to create a new swap.

Maybe we should provide a msg that locks/invalidate this contract? so if a mistake happened when instantiating, it can be easily closed and refunded to make sure it won't be used in the future by mistake (new issue new PR of course)
Probably should be able to be closed by any party or by the the instantiate sender.

contracts/external/cw-token-swap/src/contract.rs
```diff
@@ -88,6 +89,7 @@ fn get_counterparty<'a>(
 /// escrow funds if both counterparties have funded the contract.
 ///
 /// NOTE: The caller must verify that the denom of PAID is correct.
+#[allow(clippy::too_many_arguments)]
```
0xekez, line 92: do we need this? none of the arguments in the PR have changed. think this might be leftover from the first pass.

contracts/external/cw-token-swap/src/contract.rs
```diff
@@ -112,14 +114,17 @@ fn do_fund(
...
+            .into_send_message(other_counterparty.address.to_string(), false)?;
+
+        msgs.append(
+            &mut other_counterparty
                 .promise
-                .into_send_message(&counterparty.address)?,
-        ]
+                .into_send_message(counterparty.address.to_string(), false)?,
+        );
+        msgs
```
0xekez, line 127: why rewrite this? seems like we have an additional clone this way.

contracts/external/cw-token-swap/src/contract.rs
```diff
@@ -217,7 +222,8 @@ pub fn execute_withdraw(deps: DepsMut, info: MessageInfo) -> Result<Response, Co
     let message = counterparty
         .promise
         .clone()
-        .into_send_message(&counterparty.address)?;
+        .into_send_message(counterparty.address.to_string(), true)?[0]
+        .clone();
```
0xekez, line 226: why do we need to take the first index and clone? should the signature be `into_send_messages`?

contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,420 @@
...
+                    label,
+                }
+                .into(),
+            )),
+        }
+    }
+}
+
+#[cw_serde]
+pub enum Cw20SendMsgs {
```
0xekez, line 135: i'm curious why you chose to break these messages into multiple enums and use an interface, instead of just using one enum? for example:

```
pub enum SomeMessages {
  Cw20(Cw20SendMsgs),
  Native(NativeSendMsgs),
}
```

personally, I am pretty anti traits and OOP style code as I think its less readable and thus more dangerous.

contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,420 @@
...
+}
+
+/// This function returns a vector of CosmosMsgs from the given vector of CompletionMsgs
+/// It verifies the msgs is not empty (else return empty vec)
+/// It verifies the total amount of funds matches the funds sent in all messages
+fn completion_to_cosmos_msgs<T: CompletionMsgsToCosmosMsg>(
+    deps: Deps,
+    msgs: Vec<T>,
+    amount: Uint128,
+    string: &str, // extra data we need (denom or contract addr)
```
0xekez, line 23: is there a way that we could make this an enum or otherwise avoid overloading the `&str` type? see later comment about not using traits which may help with this.

contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,420 @@
...
+                return Err(ContractError::InvalidSendMsg {});
+            }
+            if coins[0].amount.is_zero() {
+                return Err(ContractError::InvalidSendMsg {});
+            }
+            if denom != coins[0].denom {
+                return Err(ContractError::InvalidSendMsg {});
+            }
+            Ok(coins[0].amount)
+        };
```
0xekez, line 86: i think this is a reimplementation of cw-utils [`must_pay`](https://github.com/CosmWasm/cw-utils/blob/main/src/payment.rs#L30-L39). might be good to use that here.

contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    // Verify that total amount of funds matches funds sent in all messages
+    if total_amount != amount {
+        return Err(ContractError::WrongFundsCalculation {});
+    }
+    Ok(cosmos_msgs)
+}
+
+#[cw_serde]
+pub enum NativeSendMsg {
```
0xekez, line 47: all of these variants have a `Vec<Coin>` parameter that we want to be sure to validate. currently, we match and validate on each branch:

```
match type {
   a => validate_coins(a.coins),
   b => validate_coins(b.coins),
}
```

what if we moved the amount to a top level field in a struct.

```
struct NativeSend {
  coins,
  msg,
}
```

then, we don't have to remember to validate in each match arm which makes it less likely we forget!

```
validate_coins(coins)
match msg {
  a => ...,
  b => ...,
}
```

0xekez (commented):
looking good! a few comments on the new trait. I think it would make the code cleaner / safer to use an enum as we only have two variants.

contracts/external/cw-token-swap/src/types.rs
Art3miX: I wish, but the whole point here is that we take all native msgs we allow as is, with the funds parameter, this also tells us how much exactly to send on each msg.
If we do the above, there is zero verification that the coins they sent in `NativeSend `, is the actual coins they set in each msg, so the check will be worthless because we don't verify the actual sent funds in the msgs.

contracts/external/cw-token-swap/src/types.rs
Art3miX: Actually this was done to avoid doing the same checks 2 times and nothing more, I understand it totally added unnecessary confusion/complexity to it, I will remove it.

contracts/external/cw-token-swap/src/types.rs
Art3miX: Removed this function, unneeded.

contracts/external/cw-token-swap/src/types.rs
Art3miX: You are not wrong, but `must_pay` accepts `MessageInfo` and from the looks of it was built specifically for testing funds sent with msgs, which is not the case here (Exactly, kind of).
To use that we either need to:
1. build `MessageInfo` struct with unchecked addr, on every match.
2. or create 1 `mut MessageInfo` struct and modify the funds field on every match.

Both feels hackish and wrong to me, I think having this incapsulated helper function makes it clear what it does and why its there.

contracts/external/cw-token-swap/src/contract.rs
Art3miX: Initially `into_send_message` was returning a cosmosMsg, but our change makes it that `into_send_message` returns a **vector** of cosmosMsgs, so we need to combine the vectors, rather then create new vector from 2 msgs.

Changed it to use `extend` instead of `append` which removes the `&mut from the other_counterparty vector.

For some reason I managed to remove the clone, not sure why I had it in the first place tho, seems not related to the `extend` move.

contracts/external/cw-token-swap/src/types.rs
0xekez: great point :) thanks.

contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,420 @@
...
+                                Ok(cosmos_msg)
+                            })
+                            .collect::<Result<Vec<CosmosMsg>, ContractError>>()?;
+
+                        // Verify that total amount of funds matches funds sent in all messages
+                        if total_amount != amount {
+                            return Err(ContractError::WrongFundsCalculation {});
+                        }
+                        cosmos_msgs
+                    };
```
0xekez, line 243: reading this code i my spidey sense was tingling. first, the iteration here can be done functionally with a fold over a tuple (no need for `mut` _and_ `map`):

```rust
                    let (on_completion, tokens_sent) = on_completion.into_iter().try_fold(
                        (vec![], Uint128::zero()),
                        |(mut messages, total_sent), msg| -> Result<_, ContractError> {
                            let (sent, msg) = msg.into_checked_cosmos_msg(deps, &denom)?;
                            messages.push(msg);
                            Ok((messages, sent + total_sent))
                        },
                    )?;
```

second, only needing to check the length in the `on_completion.len() > 0` case seems strange. the semantics of that suggest that nothing will be done on completion meaning... what? there exist both a bank and cw20 transfer option in the message types, so shouldn't a transfer to the other counterparty be represented as one of those? why have both ways to trigger a transfer to the other counterparty?

this is one of those weird corner cases of code where the logic gets weird and you realize you've got a wrong abstraction. how about making doing nothing on completion not allowed, and expressing the desire to transfer tokens with the transfer completion options?

Art3miX: 1. Thats a neater way of doing it, I like it.
2. I look at it as having a default, at the end of the day we are talking about a swap contract, the default behavior is sending the tokens to the other party, this design allows having a basic swap in place without adding complexity to calling it, so providing an empty vector will result in a default swap behavior.
If you need something more complex, you can provide the msgs you want in the `on_completion` field.

Im not against forcing all msgs on `on_completion`, I just think having this default makes a much better UI.

Thinking about it, I think I can set the default here instead of later, which will make it much clearer that we set default.

0xekez: (2) sounds like a good idea. the default can be to fill with a transfer message, maybe we just do that at instantiation time?

0xekez: might be nice to break this logic into a function as well so it can be reused in the cw20 case

contracts/external/cw-token-swap/src/types.rs
```diff
@@ -0,0 +1,420 @@
...
+
+#[cw_serde]
+pub enum NativeSendMsgs {
+    BankSend {
+        to_address: String,
+        amount: Vec<Coin>,
+    },
+    BankBurn {
+        amount: Vec<Coin>,
+    },
```
JakeHartnell, line 14: Why even have this? Seems like it just risks someone accidentally using it?

PR #681, opened 2023-03-22
Prepare for 2.1 release

0xekez:
- Bumps package versions
- Adds README to stake-tracker
- Adds license field to wormhole and stake-tracker
- Removes dependencies on multi-test and storage-plus forks
- Bumps to latest cw-ownable version

this, #608 and then we'll cut the release!

PR #682, opened 2023-03-24
prepare migrator to v2.1

### LINKED ISSUE #680: Final cleanups for dao-migrator
Just a couple tiny things I think we should clean up before releasing this:

1. I forgot to mention this when we are designing it, but we also need to support migrations for v1 DAOs that are using a legacy cw20-stake contract (found [here](https://discord.com/channels/@me/1018789147839827968/1088169842089136191) on the v0.3.0 tag). The contract is not different between v0.3.0 and v1.0.0 (confirm this tho when doing the implementation), so the migration logic for dao-migrator shouldn't need to change. We just need to confirm it works with a test as we have a couple DAOs in prod using the legacy contract.
2. There is a TODO comment on the `query_state_v1` method. It seems like it has been resolved so we can remove that.

Other than that, `dao-migrator` is looking really good!

Art3miX:
Closes #680

### DIFF
--- a/contracts/external/dao-migrator/src/testing/test_migration.rs
+++ b/contracts/external/dao-migrator/src/testing/test_migration.rs
@@ -33,6 +33,11 @@ pub fn basic_test(voting_type: VotingType, from_core: bool) {
             module_addrs.proposals[0].clone(),
             module_addrs.voting.clone(),
         ),
+        VotingType::Cw20V03 => query_state_v1_cw20(
+            &mut app,
+            module_addrs.proposals[0].clone(),
+            module_addrs.voting.clone(),
+        ),
     };
     //NOTE: We add 1 to count because we create a new proposal in execute_migration
     test_state_v1.proposal_count += 1;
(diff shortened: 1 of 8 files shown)


contracts/external/dao-migrator/src/testing/test_migration.rs
```diff
@@ -81,10 +91,12 @@ fn test_execute_migration() {
     // Test basic migrator (not called from core)
     basic_test(VotingType::Cw20, false);
     basic_test(VotingType::Cw4, false);
+    basic_test(VotingType::Cw20V03, false);
 
     // Test basic migrator (called from core)
     basic_test(VotingType::Cw20, true);
     basic_test(VotingType::Cw4, true);
+    basic_test(VotingType::Cw20V03, true);
 }
```
0xekez, line 100: your test setup is sooo good!

0xekez (approved):

0xekez:
merged this so you should be able to rebase and fix the gas limit error: https://github.com/DA0-DA0/dao-contracts/pull/681

JakeHartnell (approved):

PR #683, opened 2023-03-25
Add boot integration and initial test for cw-token-swap

adairrr:
This change was built with @0xekez and adds a basic test for funding on the cw-token-swap contract.

@CyberHoward


JakeHartnell (approved):
Very nice, would be great to do this across the board? Really liking boot. :heart:

JakeHartnell:
A little outdated now, we should revist. : )

PR #691, opened 2023-04-03
Dao voting cw4 cleanup

### LINKED ISSUE #675: `dao-voting-cw4` maintains duplicate state, does not allow use of existing group
The `dao-voting-cw4` contract maintains duplicate state compared to the `cw4-group` contract, and does not allow the use of an existing group contract. Worth a refactor?

Would be great if the voting contracts shared the same structure and properties, such as being able to create a new contract during instantiation or using an existing one.

JakeHartnell:
Closes #675

### DIFF
--- a/contracts/voting/dao-voting-cw4/src/contract.rs
+++ b/contracts/voting/dao-voting-cw4/src/contract.rs
@@ -1,15 +1,15 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdError, StdResult,
-    SubMsg, Uint128, WasmMsg,
+    to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdResult, SubMsg, WasmMsg,
 };
 use cw2::set_contract_version;
+use cw4::{MemberResponse, TotalWeightResponse};
 use cw_utils::parse_reply_instantiate_data;
 
 use crate::error::ContractError;
-use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg};
-use crate::state::{DAO, GROUP_CONTRACT, TOTAL_WEIGHT, USER_WEIGHTS};
+use crate::msg::{GroupContract, InstantiateMsg, MigrateMsg, QueryMsg};
+use crate::state::{DAO, GROUP_CONTRACT};
 
 pub(crate) const CONTRACT_NAME: &str = "crates.io:dao-voting-cw4";
 pub(crate) const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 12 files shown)


contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -24,124 +25,79 @@ pub fn instantiate(
...
+            };
+
+            let msg = SubMsg::reply_on_success(msg, INSTANTIATE_GROUP_REPLY_ID);
+
+            Ok(Response::new()
+                .add_attribute("action", "instantiate")
+                .add_submessage(msg))
         }
+        GroupContract::Existing { address } => {
+            let group_contract = deps.api.addr_validate(&address)?;
```
bekauz, line 83: could something go wrong here if the address is valid but does not point to a valid `GroupContract`?

JakeHartnell: Yes. @0xekez, any ideas for validation here?

Art3miX (commented):
Looks good! Just got 1 question

What about admin of existing cw4-group? 
When creating new group you set it to be the DAO, maybe this contract should verify that the DAO is the admin or its not set?

JakeHartnell:
> What about admin of existing cw4-group?
> When creating new group you set it to be the DAO, maybe this contract should verify that the DAO is the admin or its not set?

Good question! IMO this should not be enforced. What about things like SubDAOs where a certain groups contract may be used across multiple SubDAOs?

That said, folks should exercise caution when using existing contracts if they don't understand who controls them. Feel this should be more of a UX thing.

contracts/voting/dao-voting-cw4/src/tests.rs
```diff
@@ -138,6 +143,73 @@ fn test_instantiate() {
...
+
+    // Instantiate with existing contract
+    let msg = InstantiateMsg {
+        group_contract: GroupContract::Existing {
+            address: cw4_addr.to_string(),
+        },
+    };
+    let _err = app
+        .instantiate_contract(
+            voting_id,
```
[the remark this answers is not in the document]
JakeHartnell: Yes!  Make a code suggestion and get credit.

contracts/voting/dao-voting-cw4/src/tests.rs
```diff
@@ -138,6 +143,73 @@ fn test_instantiate() {
...
+        )
+        .unwrap();
+
+    // Instantiate with existing contract
+    let msg = InstantiateMsg {
+        group_contract: GroupContract::Existing {
+            address: cw4_addr.to_string(),
+        },
+    };
+    let _err = app
```
Art3miX, line 176:
```suggestion
    app
```

contracts/voting/dao-voting-cw4/src/contract.rs
bekauz: Perhaps it would be nice to use a similar pattern as with `UncheckedDepositInfo` and `CheckedDepositInfo` in `dao-voting` module to validate the group contract

bekauz (approved):

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -24,124 +25,79 @@ pub fn instantiate(
...
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn execute(
+    _deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    _msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    Err(ContractError::NoExecute {})
```
Art3miX, line 100: Complains no tests, I guess you can just add a simple test that trying to execute something errors? 
should be a good reminder in the future to add test if execute msgs are added as well

contracts/voting/dao-voting-cw4/src/contract.rs
Art3miX: Whats the conclusion on this? I think some minimal verification is best to at least try and catch some obvious user errors

Art3miX (approved):

JakeHartnell:
Made some clean up tickets:
- https://github.com/DA0-DA0/dao-contracts/issues/712
- https://github.com/DA0-DA0/dao-contracts/issues/711

PR #697, opened 2023-04-21
Augmented Bond Curves

JakeHartnell:
ABCs are cool! This will try and implement much of what is in this [excellent blog post](https://medium.com/commonsstack/deep-dive-augmented-bonding-curves-b5ca4fad4436).

To do:
- [x] Add max supply option
- [x] Update curve
- [x] Update max supply
- [x] Close bonding curve (only sell)
- [x] MinMax contribution amounts for hatching
- [x] Fix token metadata so that it's set correctly
- [x] Test allowlist
- [x] Test close curve
- [x] Test update curve
- [x] Test update max supply
- [x] Factory contract for creating cw-abc DAOs
- [ ] Curve type validation
- [ ] Move related contracts to `experimental` folder

Bonus:
- [ ] Move curves into a new `cw-curves` package
- [ ] Add more curves (Taylor Series. S-Curve, Piecewise Linear)
- [ ] Extend allowlist options (i.e. keep existing list, but add a query to check if an address is a member of a DAO)


bmorphism:
doing G-d's work, Ser! 👏

JakeHartnell:
For the `cw-curves` package it may be worth looking and the Wynd DAO implementation: https://github.com/wynddao/wynddao/blob/main/packages/utils/src/curve.rs

Probably some useful stuff in there. Would be nice to have a solid curves package.

contracts/external/cw-abc/src/contract.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    PHASE_CONFIG.save(deps.storage, &phase_config)?;
+
+    // TODO don't hardcode this? Make it configurable? Hatch config can be optional
+    PHASE.save(deps.storage, &CommonsPhase::Hatch)?;
+
+    // Initialize owner to sender
+    cw_ownable::initialize_owner(deps.storage, deps.api, Some(info.sender.as_str()))?;
+
+    // Tnstantiate cw-token-factory-issuer contract
```
ismellike, line 89: Spelling of Instantiate

contracts/external/cw-abc/src/msg.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// Update the phase configurations.
+/// These can only be called by the owner.
+#[cw_serde]
+pub enum UpdatePhaseConfigMsg {
+    /// Update the hatch phase configuration
+    Hatch {
+        contribution_limits: Option<MinMax>,
+        // TODO what is the minimum used for?
```
ismellike, line 42: I think paired with an expiration, the minimum could be used to determine if the ABC Hatch phase was successful. 

If expired and the minimum was not met, then the ABC goes into the Closed phase. If expired with the minimum met, then the ABC goes into the Open phase.

contracts/external/cw-abc/src/commands.rs
```diff
@@ -0,0 +1,571 @@
...
+        contract_addr: issuer_addr.to_string(),
+        msg: to_json_binary(&IssuerExecuteMsg::Mint {
+            to_address: info.sender.to_string(),
+            amount: minted,
+        })?,
+        funds: vec![],
+    })];
+
+    // Send funding to fee recipient
+    if funded > Uint128::zero() {
```
ismellike, line 104: This would start to lose state of what's actually in the funding pool.

JakeHartnell:
Closing in favor of #926

PR #698, opened 2023-04-22
Initial phase integration to cw-abc

adairrr:
This change does the following:
- initial phase configuration setup for cw-abc
 - phase configurations and current phase stored in state
 - Hatch phase has an allowlist
 - Basic phase queries
- cw-ownable integration
- Move queries & commands to `queries` and `commands` respectively
- Addition of `funding` pool to the curve state

There are TODOs where functionality is still missing. The burn mechanism was left untouched.

Questions: 
1) Do we want to have a `PreHatch` phase where the configuration can be changed freely? Hatch configuration for the `initialPrice` should NOT be changed after the phase has begun, though the `min` and `max` `initialRaise` should be able to be changed (like in traditional fundraising)
2) Do we want to separate the phase configurations into their own states so that there's not unnecessary loading of state? 
3) Closed phase? Necessary? 
4) Testing - BOOT will be updated to work with custom messages shortly, so should be able to integrate soon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,6 +587,7 @@ dependencies = [
  "cosmwasm-schema",
  "cosmwasm-std",
  "cw-address-like",
+ "cw-ownable",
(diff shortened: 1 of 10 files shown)


JakeHartnell (approved):
Love this direction! :heart:

PR #699, opened 2023-04-23
cw-abc: Updated hatch phase mechanics, donations, queries

adairrr:
This change implements the following:
1) Donation feature for donating tokens directly to the funding pool
- Keeps track of donations and is queryable
2) Initial sell exit-tax implementation 
- Q: do we want hatchers to be able to sell their tokens at all? Right now there is an exit tax functionality but it would allow early investors to dump on later investors. Maybe we should require it to be > 50%?
3) Hatchers are tracked in a `Map` to allow for vesting schedule checking, otherwise their tokens bought afterwards would be stuck in their same vesting schedule
- Queryable
4) Update phase config
5) Initial boot integration and testing infra setup
- TODO: we need to make a PR to the `token-bindings` so that we can have a custom handler for the TokenFactory.

### DIFF
--- a/contracts/external/cw-abc/Cargo.toml
+++ b/contracts/external/cw-abc/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cw-abc"
 version = "0.0.1"
-authors = ["Ethan Frey <ethanfrey@users.noreply.github.com>", "Jake Hartnell"]
+authors = ["Ethan Frey <ethanfrey@users.noreply.github.com>", "Jake Hartnell", "Adair <adairrr@users.noreply.github.com>"]
 edition = { workspace = true }
 description = "Implements an Augmented Bonding Curve"
 license = "Apache-2.0"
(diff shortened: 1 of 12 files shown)


adairrr:
https://github.com/DA0-DA0/dao-contracts/pull/697

contracts/external/cw-abc/Cargo.toml
```diff
@@ -16,6 +16,7 @@ crate-type = ["cdylib", "rlib"]
 backtraces = ["cosmwasm-std/backtraces"]
 # use library feature to disable all instantiate/execute/query exports
 library = []
+boot = ["dep:boot-core"]
```
JakeHartnell, line 19: :heart:

contracts/external/cw-abc/src/integration.rs
```diff
@@ -0,0 +1,38 @@
+use crate::{abc::CurveType, boot::CwAbc};
+use boot_core::{BootUpload, Mock};
+use cosmwasm_std::{Addr, Uint128};
+
+use crate::testing::prelude::*;
+
+type AResult = anyhow::Result<()>; // alias for Result<(), anyhow::Error>
+
+// TODO: we need to make a PR to token factory bindings for the CustomHandler so that messages will actually execute
```
JakeHartnell, line 9: :+1:

adairrr: https://github.com/CosmWasm/token-bindings/issues/7

JakeHartnell (approved):
Me like. :heart:

PR #706, opened 2023-06-01
Support migrating cw20-stake from legacy v0.2.5 directly to v2

NoahSaso:


noahsaso:
No need as the only stuck DAOs can migrate staking contract to v1 and then upgrade to v2 normally.

PR #728, opened 2023-08-18
Token Factory DAOs

JakeHartnell:
Taking a slightly different approach here. This uses Token Factory directly rather than the middleware contract (one less thing), it also gets tests working.

Closes #659 
Closes #664

PR #731, opened 2023-08-20
Token Factory DAOs

### LINKED ISSUE #659: Test `dao-voting-native-staked` against token factory
Token factory is coming to Juno mainnet very soon (it's already on testnet!).
Juno PR: https://github.com/CosmosContracts/juno/pull/368
Module: https://github.com/CosmosContracts/token-factory/tree/main/x/tokenfactory

The [`dao-voting-native-staked`](https://github.com/DA0-DA0/dao-contracts/tree/main/contracts/voting/dao-voting-native-staked) voting contract already supports native denoms, and we likely want to switch token-based DAOs to using token factory tokens instead of cw20s... eventually. Before doing all THAT, we should probably test this thoroughly against the new x/tokenfactory module since new code was added and code scary sdk code scary nightmare nightmare big huge danger time oh no. Let's test it!

### LINKED ISSUE #664: `dao-voting-native-staked` missing features
- [ ] Hooks like cw20-stake and NFT staking
- [ ] Support active threshold
- [ ] Create a native token on Instantiation like cw20 or use existing

JakeHartnell:
Follow up from #728

Closes https://github.com/DA0-DA0/dao-contracts/issues/659
Closes https://github.com/DA0-DA0/dao-contracts/issues/664


contracts/voting/dao-voting-token-factory-staked/src/msg.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// Removes a hook that fires on staking / unstaking
+    RemoveHook { addr: String },
+}
+
+#[active_query]
+#[voting_module_query]
+#[token_query]
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
```
noahsaso, line 98: missing query to retrieve token issuer contract address from state

noahsaso (changes requested):

contracts/voting/dao-voting-token-factory-staked/src/msg.rs
JakeHartnell: That's the `#[token_query]` attribute. :upside_down_face: 

Sometimes I don't like these attributes for messages because they are harder to see.

noahsaso (approved):

PR #738, opened 2023-08-30
Timelock

### LINKED ISSUE #736: Timelock on proposal execution
A good feature to upstream from the Neutron DAO fork is timelock on proposal execution. I.E. when a proposal passes voting there is configurable delay on execution whereby a configurable address (or parent DAO) may vote to veto it.

Their implementation: https://github.com/neutron-org/neutron-dao/tree/448c7c91e85ccd02d13fdaf7cddb66f04abf8ca9/contracts/subdaos/cwd-subdao-timelock-single

There are probably a couple of ways to implement this, but feel the most straight forward is to implement it at the proposal module level (for example, in `dao-proposal-single` and `dao-proposal-multiple`.

JakeHartnell:
Closes #736


JakeHartnell:
Might explore a different approach.

PR #742, opened 2023-09-02
Audit Fixes and Improvements

### LINKED ISSUE #729: Consistent Voting Power Change hook interfaces
Right now, each voting contract has its own interface for voting power change hooks. This is a bit annoying to support if you want to make a generic contract that uses voting power change hooks for any DAO. Let's make this nicer.

JakeHartnell:
Address most of the issues from the recent smart contract audit, also fixes one or two things they didn't catch. Many improvements across the board.

Additionally:
- Closes #729

Major changes from the current `development` version:
- `dao-voting-native-staked` and `dao-voting-token-factory-staked` have been merged into one contract to reduce code duplication. When using an `existing` token, a new `cw_tokenfactory_issuer` is no longer instantiated as it was pointless and it would be impossible to garuntee if the admin would be set correctly.
- The API of `cw_tokenfactory_issuer` has been greatly simplified.
- `cw_tokenfactory_issuer` uses `cw_ownable` to implement two-step ownership transfer, as well as to allow for DAOs to easily reject ownership altogether.
- `dao-voting-cw721-staked` is now consistent with the other voting modules in terms of not taking an `owner`, the owner of a voting module should always be the DAO.
- Advanced features such as `freezing`, `denylist`, and more are disabled by default in `cw_tokenfactory_issuer`. DAOs must explicitly vote to enable them later so everyone can see what they are getting into.
- Hooks are now properly dispatched during staking and unstaking. Additionally, hooks have been merged into a new `dao-hooks` package to allow for better code reuse. A new `DaoHooks` enum contains all hooks used in the codebase to make it easy to write generic contracts leveraging any DAO hook.
- Better validation with more code reuse.
- Unsetting the `BeforeSendHook`, renouncing ownership, and disabling Token Factory admin have been documented and tested.

In addition, many other improvements have been added, including numerous small informational issues (non-critical) mentioned in the audit.


contracts/external/cw-tokenfactory-issuer/tests/cases/set_before_update_hook.rs
cheddar27, line 1: Rename file from "set_before_update_hook.rs" to "set_before_send_hook.rs"

JakeHartnell: Done!

contracts/voting/dao-voting-native-staked/src/contract.rs
```diff
@@ -66,10 +66,18 @@ pub fn instantiate(
     CONFIG.save(deps.storage, &config)?;
     DAO.save(deps.storage, &info.sender)?;
 
+    // Validate denom by checking supply
+    let supply: Coin = deps.querier.query_supply(msg.denom.to_string())?;
+    println!("supply {:?}", supply);
```
cfkelly18: The fix looks good to me! There is just a println remaining in line 71 that looks like its left over from debugging

contracts/voting/dao-voting-native-staked/src/contract.rs
```diff
@@ -66,10 +66,18 @@ pub fn instantiate(
...
+    if Uint128::is_zero(&supply.amount) {
+        return Err(ContractError::InvalidDenom {});
+    }
+
+    // Validate active threshold
```
cfkelly18: The fix looks good to me! There is just a println remaining in line 71 that looks like its left over from debugging

JakeHartnell: Silly me! Thanks for catching that.

noahsaso (approved):
damn beast.

test-contracts/dao-proposal-hook-counter/src/contract.rs
```diff
@@ -73,6 +79,28 @@ pub fn execute_proposal_hook(
...
+        StakeChangedHookMsg::Stake { .. } => {
+            let mut count = STAKE_COUNTER.load(deps.storage)?;
+            count += Uint128::new(1);
+            STAKE_COUNTER.save(deps.storage, &count)?;
+        }
+        StakeChangedHookMsg::Unstake { .. } => {
+            let mut count = STAKE_COUNTER.load(deps.storage)?;
+            count += Uint128::new(1);
+            STAKE_COUNTER.save(deps.storage, &count)?;
+        }
```
jcr-oaksec, line 98: Both `Stake` and `Unstake` adding to the counter look like a mistake unless you are tracking something like "total number of operations". Can you please check?

contracts/voting/dao-voting-cw721-staked/src/contract.rs
```diff
@@ -655,31 +664,55 @@ pub fn reply(deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractE
...
+                    .querier
+                    .query_wasm_smart(collection_addr, &Cw721QueryMsg::NumTokens {})?;
+
+                // Chec the count is not greater than supply
+                if count > Uint128::new(supply.count.into()) {
+                    return Err(ContractError::InvalidActiveCount {});
+                }
+            }
+            Ok(Response::new())
+        }
```
jcr-oaksec, line 715: This only partially remediates issue 7; there can still be a supply of zero NFTs if an `ActiveThreshold::Percentage` is in place.

test-contracts/dao-proposal-hook-counter/src/contract.rs
JakeHartnell: Total number of operations.

Issue #737, opened 2023-08-29
Allow Admin to Pause / Unpause SubDAO


JakeHartnell:
Parent DAOs should probably be allowed to pause / unpause SubDAOs.

ismellike:
Can be marked as closed #744

PR #744, opened 2023-09-07
Allow admin to pause/unpause subdao

ismellike:
Also fixed some typos and removed unused msg.rs from dao-dao-core

#737


contracts/dao-dao-core/src/tests.rs
```diff
@@ -1150,6 +1150,52 @@ fn test_admin_permissions() {
...
+    assert_eq!(paused, PauseInfoResponse::Unpaused {});
+
+    // DAO cannot unpause itself
+    let res = app.execute_contract(
+        core_with_admin_addr.clone(),
+        core_with_admin_addr.clone(),
+        &ExecuteMsg::Unpause {},
+        &[],
+    );
+    assert!(res.is_err());
```
JakeHartnell: Would swap the order of these two tests and add one more:
- Check the DAO cannot unpause itself
- Check a random person can't pause it
- Check the admin can unpause successfully

JakeHartnell (commented):
Only one small comments on tests and then we can get this in!

Also, ignore the lints, have a fix for those.

Sometimes new rust versions break CI.

contracts/dao-dao-core/src/contract.rs
```diff
@@ -174,6 +184,21 @@ pub fn execute_pause(
...
 
+pub fn execute_unpause(deps: DepsMut, sender: Addr) -> Result<Response, ContractError> {
+    let admin = ADMIN.load(deps.storage)?;
+
+    // Only the admin may call this method.
+    if sender != admin {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    PAUSED.remove(deps.storage);
```
noahsaso, line 191:
```suggestion
    PAUSED.remove(deps.storage)?;
```

i think this line is missing an unwrap?

noahsaso (changes requested):
sick! ~i think i found one line that needs fixing~, and it seems like you need to format (`cargo fmt --all -- --check`)

also, it wouldn't hurt to add a test that random person cannot unpause as well, not just the DAO itself

update: i was wrong, no fix needed other than the format

contracts/dao-dao-core/src/contract.rs
noahsaso: oh nvm i'm wrong

contracts/dao-dao-core/src/contract.rs
```diff
@@ -159,8 +167,10 @@ pub fn execute_pause(
     sender: Addr,
     pause_duration: Duration,
 ) -> Result<Response, ContractError> {
-    // Only the core contract may call this method.
-    if sender != env.contract.address {
+    let admin = ADMIN.load(deps.storage)?;
+
+    // Only the core contract or admin may call this method.
+    if sender != env.contract.address && sender != admin {
```
noahsaso: hmm, maybe only the admin should be able to call this message (non-SubDAOs have admin set to self). are there other messages that allow both the core contract and the admin?

ismellike: Ok I see the pattern is usually just checking for the sender being the dao-core, and the admin will execute through the `ExecuteAdminMsgs`.
I'll keep that same pattern and also allow the ExecuteAdminMsgs through when paused.

noahsaso: actually, i think it makes sense that *only* the admin can call unpause, because ofc the SubDAO can't unpause itself normally.

hm that brings up another question actually:

should the parent be able act as the SubDAO if the SubDAO is paused?

this feels like an arbitrary design decision i don't know how to make. @bekauz @JakeHartnell any thoughts?

noahsaso: i sent a message in the discord to solicit some community input:
https://discord.com/channels/895922260047720449/895926354728603659/1202665603446603847

ismellike: Actually, Unpause should only be called directly by the admin. A dao can unpause itself if admin, and only a parent dao should be able to unpause the subDao.

noahsaso: right well a DAO can't unpause itself ever, because if it's paused, it can't execute messages. that's why i think it's safe to just check the admin.

noahsaso: ok, i think a parent DAO may want to execute messages on behalf of a SubDAO when it's paused for security purposes. maybe the SubDAO is being taken over and the DAO needs to rescue funds or update membership.

ismellike: We'd have to refactor the way authorization works. Right now it's Admin -> ExecuteAdminMsgs, which calls DAO methods from DAO, because the actions are only checking to see if the sender is the dao itself. The msg loses some context.

Maybe we get rid of ExecuteAdminMsgs and create a helper method for checking authorization at each call based off admin, paused, and dao?

noahsaso: why would we have to refactor that? what you said earlier about allowing `unpause` and `execute_admin_msgs` through while paused should be fine, right? as long as both `unpause` and `execute_admin_msgs` just check that sender is the admin.

ismellike: Because `execute_admin_msgs` is just passing the msgs through to be executed as the DAO itself using `.add_message(msgs)`. 

Take for example a paused DAO where we allow admin messages:
Admin messages pass through just fine.
The msgs are now attempting to be executed with `info.sender` being `env.contract.address`, but the message will now not be of type execute_admin or unpause and fail mesage check at the beginning.

noahsaso: ah i see the confusion. `.add_message(msgs)` is the actual chain-level "i want to execute this message as myself" in a contract. those do not get routed through execute (unless of course the parent DAO tries to tell the SubDAO to execute a message on itself, such as updating core info). so i'm 99% sure that allowing both `unpause` and `execute_admin_msgs` will do what we want.

that actually brings up another use case: what if the admin wants to execute messages on the SubDAO while it's paused?

now i'm thinking that when a SubDAO is paused, the admin should be allowed to send any message it would normally be allowed to send (i.e. entirely skip the paused check if the admin is the sender), and any other sender errors out when paused. i believe this is safe because it will still prevent proposal modules from executing anything, and it will allow the admin to execute anything they want, including unpause.

also, to allow the admin to tell the SubDAO to update its core info while it's paused, the SubDAO needs to be able to execute messages on itself too. so really i think both the admin and the SubDAO can skip the paused check entirely. this still prevents proposal modules from executing anything, so it should have the same intended effect, while allowing the admin to do things as the SubDAO *and* tell the SubDAO to do things on itself (like `UpdateConfig`)

does this make sense? i may be missing something, there are so many layers here D:

noahsaso: @bekauz pls sense-check everything i just said ^

contracts/dao-dao-core/src/contract.rs
```diff
@@ -106,10 +106,18 @@ pub fn execute(
...
+        }
+        _ => {
+            // Check if the DAO is currently paused
+            if let Some(expiration) = PAUSED.may_load(deps.storage)? {
+                if !expiration.is_expired(&env.block) {
+                    return Err(ContractError::Paused {});
+                }
+            }
         }
     }
```
bekauz: how about this?
```suggestion
    // No actions can be performed while the DAO is paused except for unpause.
    if let Some(expiration) = PAUSED.may_load(deps.storage)? {
        if !expiration.is_expired(&env.block) && !matches!(msg, ExecuteMsg::Unpause {}) {
            return Err(ContractError::Paused {});
        }
    }

```

contracts/dao-dao-core/src/contract.rs
```diff
@@ -159,7 +168,6 @@ pub fn execute_pause(
     sender: Addr,
     pause_duration: Duration,
 ) -> Result<Response, ContractError> {
-    // Only the core contract may call this method.
```
bekauz: why remove this comment?

contracts/dao-dao-core/src/contract.rs
bekauz: first, i'm a little paranoid about how this can potentially impact the veto flow. could we get some tests done to cover pausing/unpausing dynamics with a veto committee subdao?

same goes for the `execute_admin_msgs` discussion. we should have tests to cover those cases, because even though this is a straightforward pr, it impacts other flows.

bekauz (commented):
overall lgtm!

that being said, it would be great to add some extended testing for this. as mentioned in comment, imho adding some tests that would combine the pausing/unpausing logic with a veto committee would be a good way to make sure we aren't introducing some potential chaos here.

bekauz:
also now that i think of it, i'm not sure how i feel about the fact that pausing / unpausing can be done by different parties. don't have a solution off the top of my head right now, but something to think about.

contracts/dao-dao-core/src/contract.rs
JakeHartnell: Yeah, would keep it.

ismellike: Sure thing had just removed for consistency with other methods

JakeHartnell: IMO more comments is always good.

JakeHartnell:
> also now that i think of it, i'm not sure how i feel about the fact that pausing / unpausing can be done by different parties. don't have a solution off the top of my head right now, but something to think about.

Had similar thoughts initially, but eventually warmed up to it as a nice change that Neutron introduced. This feature is most relevant in the context of SubDAOs. _For DAOs without an admin, nothing changes_. As a parent DAO is in full control of its SubDAOs, it makes sense that that control should also extend to pausing and unpausing a SubDAO.

For example, the Juno Charter has departments that may become inactive due to lack of members or activity. Rather than having to do all the work of moving around treasuries, vesting, etc., the SubDAO can simply enter a paused state until the parent DAO decides what to do with it.

The Parent DAO (`admin`) can already pause a SubDAO, I think it is useful and safest to allow them to unpause it. To be clear there is only one party that can unpause it, and that is the Parent DAO. Otherwise, nothing changes.

contracts/dao-dao-core/src/contract.rs
ismellike: Ahh ok, I was confused about the admin telling the subdao to execute on itself, and it makes sense that we're really trying to lock down the proposal modules from further executing.

ismellike: Would it make sense to only check the paused state in the `execute_proposal_hook` method?

contracts/dao-dao-core/src/contract.rs
ismellike: Based off Noah's comment to allow the DAO and admin to ignore paused check: 
https://github.com/DA0-DA0/dao-contracts/pull/744#discussion_r1476221069

bekauz:
that makes sense!

contracts/dao-dao-core/src/contract.rs
bekauz: imho that would be a little more confusing for someone not familiar with our repo, wdyt?

ismellike: I think it makes the most sense tbh. We avoid checking if sender is admin or the core_contract on each execute, and we are more direct with what is attempting to be accomplished.

noahsaso:
this is really interesting; i have some thoughts! sorry for the long ass brain dump but i want to make sure we know exactly what we're doing, and this is me working it out out loud 😁

both allowing the parent DAO and DAO itself to bypass the pause, AND moving the paused check to `execute_proposal_hook`, effectively change the behavior of what pause *is*

before: "lock contract, frozen until expiration or forever if no expiration set"
after: "prevent proposals from being able to execute anything"

interestingly, proposals can still be created and voted on. they just won't be executable until the DAO is unpaused. i think this is fine, and we can disable proposal creation in the UI while paused to enforce the expected behavior that you shouldn't be creating proposals in a paused DAO.

our goals (what we want from pause):
- ensure DAO or parent DAO can pause the DAO in case of emergency
- ensure DAO or parent DAO can indefinitely lock the DAO, likely upon DAO death
- ensure parent DAO can unpause a paused DAO
- ensure parent DAO can operate on the DAO while paused (in order to rescue assets, update membership, etc.)

all of these are achieved by both allowing the parent and DAO itself to call execute while paused (and no one else), AND moving the pause logic to the proposal execution hook.

something else to consider though: authz allows an account to execute on behalf of another account. if the DAO has authorized (authz granted) a wallet the authority to do something as the DAO, there is *no way* for us to prevent action being taken while paused, since the wallet authz executes as the DAO itself. we cannot prevent the wallet from updating membership, for example, since the voting module would allow that if it comes from the DAO when authz executed by the wallet. we also couldn't prevent the wallet from calling messages on the DAO itself, IFF we allow the DAO to execute messages on itself (which we have to allow if we want to let the parent tell the DAO to execute messages while paused, like to add/remove a voting or proposal module). now, it'd be very silly for a DAO to give generic authz power on or as the DAO to a wallet—that effectively sets them as admin, so maybe it's not a case we really need to design for, but we should definitely know how it will interact with this pause logic. most likely, DAO-given authz will be for very specific actions, like trading on a DEX or managing staking. in these instances, we can't do anything about them anyways, and the DAO/parent needs to know to revoke any outstanding authz grants when they go to pause. that is something the UI can help solve, and there's nothing the contracts can do about it either way. additionally, a DAO may want to give a trusted wallet or DAO (maybe a security SubDAO) the ability to pause/unpause using authz, so in all cases we must allow both the parent DAO and DAO itself to bypass the pause and execute things, as long as the proposal hook is blocked.

so the question really is: **should pausing simply mean halting the governance process (by disallowing proposal module execution)?** i think that sounds right, but now i'm curious what @0xekez had in mind at the beginning when putting the auth logic at the top level like this. since there was no ability to unpause early and no ability for the parent to execute while paused, it acts as a complete emergency freeze until the expiration (or indefinitely if no expiration). if we want to give the parent DAO the power to execute as the DAO and on the DAO, which it seems like we do, we are changing what paused means—and we need to allow the parent DAO to execute at least `execute_admin_msgs` as well as the DAO itself to execute anything (so the parent DAO can use `execute_admin_msgs` to update config, voting modules, etc.).

it seems that both keeping the paused logic in the top-level execute function (with the addition that the parent and DAO itself can still execute anything) and simply moving the paused logic to the `execute_proposal_hook` function have the same security guarantees.

wdyt @ismellike @JakeHartnell @bekauz ?

bekauz:
> it'd be very silly for a DAO to give generic authz power on or as the DAO to a wallet—that effectively sets them as admin, so maybe it's not a case we really need to design for

really great point about authz. completely flew over my head.

my immediate thought after reading through this again would be to design this flow with the assumption that DAO gave generic (or some more restrictive but still critical) authz power to some wallet. while it's unlikely and unwise, i totally echo this:
> we should definitely know how it will interact with this pause logic

@NoahSaso could we revoke all authorizations on pause? should we? just pondering here.

bekauz:
one more doomer thought regarding the authz revoking above:

the number of things that we need to consider when adding features like this makes me somewhat paranoid of getting to the point where adding something new will require us explicitly addressing numerous other features previously added _without considering something like this_. not sure if this constitutes feature creep, internal dependency hell (is that a thing?), or something else, but i think you get what i mean. veto feature was giving me similar feelings at the start, but ended up being quite encapsulated.

noahsaso:
> @NoahSaso could we revoke all authorizations on pause? should we? just pondering here.

nope 🤩

i don't think we can force the DAO to revoke all authorizations on pause without adding some way to register all of the authorizations that already exist. even if we did that, i suspect we'd run into gas limit issues if too many authorizations had been granted, thereby preventing pausing.

instead, this is something we should surface in the UI when attempting to pause. we can display all outstanding authorizations to make it clear that certain executions will still be possible, and even prompt to add actions to revoke them all in the same proposal.

> one more doomer thought regarding the authz revoking above:
> 
> the number of things that we need to consider when adding features like this makes me somewhat paranoid of getting to the point where adding something new will require us explicitly addressing numerous other features previously added _without considering something like this_. not sure if this constitutes feature creep, internal dependency hell (is that a thing?), or something else, but i think you get what i mean. veto feature was giving me similar feelings at the start, but ended up being quite encapsulated.

yeah... i think we are choosing to play the game on hard mode here. we can't avoid this problem if we want to add features. we should strongly seek to solve problems without modifying the core contract when possible, but pause has to be done directly like this. i mean, pause should really be the only feature that has to be implemented so bluntly like this. most other features have to do with proposal and voting modules.

we should be paranoid. we're building smart contracts that give the users full power to change anything about the contract and its relationship to other accounts. most other contracts aren't susceptible to this because they don't allow generic message executions, but we're basically creating a passthrough contract with some wrapper logic. i think this is the best we can do 🤷🏻‍♂️

noahsaso:
alright @ismellike, i think we're good to go on moving the paused check back to the beginning of the execution function, and skipping it if the sender is the admin or the DAO itself :D

contracts/dao-dao-core/src/contract.rs
```diff
@@ -106,10 +106,15 @@ pub fn execute(
...
     msg: ExecuteMsg,
 ) -> Result<Response, ContractError> {
-    // No actions can be performed while the DAO is paused.
+    // Check if the DAO is paused
     if let Some(expiration) = PAUSED.may_load(deps.storage)? {
         if !expiration.is_expired(&env.block) {
-            return Err(ContractError::Paused {});
+            // If paused, then only allow messages from the Admin or DAO itself
+            if info.sender != env.contract.address
+                && Some(info.sender.clone()) != ADMIN.may_load(deps.storage)?
```
noahsaso:
```suggestion
                && info.sender.clone() != ADMIN.load(deps.storage)?
```

`ADMIN` should always be defined, so it's safe to just use `ADMIN.load` here i'm pretty sure

noahsaso (changes requested):
nice! one more thing

noahsaso:
seems like you need to recreate the schema and then the linter will pass

noahsaso (approved):
SICK GREAT WORK GREAT DISCUSSIONS EVERYONE

Issue #684, opened 2023-03-26
DAO Managed Deny-list for Voting


JakeHartnell:
**Use case:** A SubDAO manages a large portion of the treasury, how to they guarantee that they won't use those tokens to vote and take over the DAO?

Solution would be to have a DAO manage vote deny list, which would give members on this list zero voting power. For example, a Development Foundation could be added this deny list to give DAO members assurance that they won't use their tokens in governance again. Probably has other uses as well.

PR #746, opened 2023-09-08
Voting power limits for dao-voting-native-staked

ismellike:
Added 2 new macros for limitable voting modules
Added an error for LimitExceeded when attempting to stake more than allowed Added test case for limits
Solution is optimistic hence the map instead of snapshotmap for limits, so reducing limit after a user has already staked will not be affected #684 Also fixed some clippy errors


contracts/voting/dao-voting-native-staked/src/contract.rs
```diff
@@ -125,6 +127,13 @@ pub fn execute_stake(
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
     let amount = must_pay(&info, &config.denom)?;
+    let limit = LIMITS.may_load(deps.storage, &info.sender)?;
+
+    if let Some(limit) = limit {
+        if amount > limit {
+            return Err(ContractError::LimitExceeded { limit });
+        }
+    }
```
JakeHartnell: Ok, have to think about this a bit...

JakeHartnell: Wondering if staking is the best place to enforce limits... :thinking: 

Like otherwise the tokens are liquid and a malicious actors sends them to another address that's not limited?

I thinking Voting Power limits should be enforced in the `VotingPowerAtHeight` query. However, this has made me realize there are actually two kinds of limits here:
- Voting Power limits
- Staking limits

Not sure how useful staking limits are... given sybils.

Anyway, will think more about this.

ismellike: This was definitely the simpler approach.
I agree that it should actually be implemented through the VotingPowerAtHeight query (and also update TotalPowerAtHeight query accordingly) to be able to avoid sybils with the unstaking_duration. 
I think this could work:

`pub const LIMITS: SnapshotMap<&Addr, Uint128> = SnapshotMap::new(
    "limits",
    "limits__checkpoints",
    "limits__changelog",
    Strategy::EveryBlock,
);`

For a user's voting power, we get the lesser of limit or voting_power if a limit exists.
For the total power, we get the original total_power and subtract all reduced_power (voting_power - limit when voting_power > limit).

JakeHartnell: Ok, think the way we should do it in this feature is enforce it in the `VotingPowerAtHeight` query. Just feels like there are less edge cases. Staking limits may also make sense at some point, but would hold back on that for now as I feel there are too many easy ways around that.

ismellike: The `TotalPowerAtHeight` query needs to be adjusted as well right? How would you imagine this being handled if yes?

PR #748, opened 2023-09-11
Use indexed map for proposal modules

ismellike:
Added an optional query parameter 'include_disabled' on ProposalModule that is false by default 
Removed ActiveProposalModule query
Added a migrate path FromV2

### DIFF
--- a/contracts/dao-dao-core/src/contract.rs
+++ b/contracts/dao-dao-core/src/contract.rs
@@ -637,13 +637,11 @@ pub fn query_proposal_modules(
         )
     };
 
-    let items: Vec<_> = if let Some(l) = limit {
-        items_iter.take(l as usize).collect::<StdResult<Vec<_>>>()?
+    if let Some(l) = limit {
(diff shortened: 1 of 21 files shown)


contracts/dao-dao-core/src/contract.rs
```diff
@@ -675,10 +667,7 @@ pub fn query_dump_state(deps: Deps, env: Env) -> StdResult<Binary> {
     let admin = ADMIN.load(deps.storage)?;
     let config = CONFIG.load(deps.storage)?;
     let voting_module = VOTING_MODULE.load(deps.storage)?;
-    let proposal_modules = PROPOSAL_MODULES
-        .range(deps.storage, None, None, cosmwasm_std::Order::Ascending)
-        .map(|kv| Ok(kv?.1))
-        .collect::<StdResult<Vec<ProposalModule>>>()?;
+    let proposal_modules = query_proposal_modules(deps, None, None, None)?;
```
ismellike, line 668: Only active proposal modules are listed here now

PR #749, opened 2023-09-13
Add error to FailedProposalExecution response

ismellike:
Minimal information is shown when a proposal execution fails, and this could give users a better clue of what went wrong. It should probably be used in all reply_on_error handles.


JakeHartnell (approved):
Nice and simple! I like it.

JakeHartnell (changes requested):
Actually, can we not use `unwrap_err`? Fine if there is some default, but in general, we try not to use `unwrap` in smart contracts.

JakeHartnell (approved):
THANKS!

PR #750, opened 2023-09-14
Remaining Audit Fixes & Factory Pattern

JakeHartnell:
Addresses a few of the remaining audit fixes, but more importantly solves a blocker for both Stargaze NFT DAOs as well as potentially future Token Factory DAOs. The Stargaze team requested that NFT creation honor would honor the fairburn mechanism.

This is made possible by the factory pattern: creating Token or NFT DAO tokens by calling into a factory contract that does validation as well as returning required information to implement the voting module. 

For example, as a user I want to sell a collection on Stargaze and use that collection and the proceeds from it to create and fund a DAO (with the DAO as the creator of the collection).

This "factory pattern" feature is intended to work with minters such as Stargaze minters or the WIP Augmented Bonding Curve tokens (or potentially other tokens such as LP tokens) which require a more complicated setup process. For these types of projects, the token is created through another contract that manages its issuance, the factory pattern is designed to support this. Factory contracts MUST set `TokenFactoryCallback` or `NftFactoryCallback` in the response.

An important note on security: as validation SHOULD happen in the factory contracts, it's important to rely only trusted factory contracts.

Overview of changes:
- `ModuleInstantiateInfo` now takes a `funds` argument so that factories that require funds (such as one using Stargaze fairburn)  can be supported.
- sg721 support has been removed in favor of the factory pattern
- A new test contract `dao-test-custom-factory` has been created for testing factory patterns as well as demonstrating how to build a custom factory.
- `test-contracts` have been moved inside the `contracts/test` folder so they can be built with the `workspace-optimizer`
- Documentation and comments about this feature have been added
- Fix inaccurate error message for threshold validation
- Improve reuse of active threshold logic in `dao-voting-cw721-staked`, and fix a bug whereby one could update the threshold to be greater than supply


noahsaso (approved):
NOICE

PR #752, opened 2023-09-19
Single Choice and Multiple Choice Veto

### LINKED ISSUE #736: Timelock on proposal execution

JakeHartnell:
Closes #736


contracts/external/dao-migrator/Cargo.toml
```diff
@@ -43,7 +44,13 @@ cw20-stake-v1                 = { workspace = true, features = ["library"] }
 cw-core-interface-v1          = { package = "cw-core-interface", version = "0.1.0", git = "https://github.com/DA0-DA0/dao-contracts.git", tag = "v1.0.0" }
 cw4-voting-v1                 = { package = "cw4-voting", version = "0.1.0", git = "https://github.com/DA0-DA0/dao-contracts.git", tag = "v1.0.0" }
 cw20-v1                       = { version = "0.13", package = "cw20" }
-cw4-v1                       = { version = "0.13", package = "cw4" }
+cw4-v1                       =  { version = "0.13", package = "cw4" }
```
0xJayak: extra empty space

contracts/pre-propose/dao-pre-propose-approver/src/tests.rs
```diff
@@ -14,7 +14,7 @@ use dao_pre_propose_approval_single::{
     state::PendingProposal,
 };
 use dao_pre_propose_base::{error::PreProposeError, msg::DepositInfoResponse, state::Config};
-use dao_proposal_single as cps;
+use dao_proposal_single as dps;
```
0xJayak: I would put cps to dps change into a dedicated PR which will make this PR smaller and more readable. Happy to handle it if you don't mind.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -255,6 +258,77 @@ pub fn execute_propose(
...
 }
 
+pub fn execute_veto(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    proposal_id: u64,
+) -> Result<Response, ContractError> {
+    let mut prop = PROPOSALS
+        .may_load(deps.storage, proposal_id)?
```
0xJayak, line 272: What's is the reason for `may_load` instead of `load` here? Should be better just end here if no proposal found no?

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -255,6 +258,77 @@ pub fn execute_propose(
...
+        .ok_or(ContractError::NoSuchProposal { id: proposal_id })?;
+
+    match prop.status {
+        Status::Timelocked { expires } => {
+            match prop.timelock {
+                Some(ref timelock) => {
```
0xJayak: Alternatively, the nesting can be reduced with `if let`, e.g.:

`if let Status::Timelocked { expires } = prop.status {
    return Err(ContractError::NotPassed {});
}`

Which would also have a benefit of handling the error sooner.

contracts/pre-propose/dao-pre-propose-approver/src/tests.rs
JakeHartnell: Please! A separate PR for this would be nice.

0xJayak: Done, see #757

contracts/proposal/dao-proposal-single/src/contract.rs
0xJayak: Here is the a snippet which works for me. But I see the `may_load` pattern is the standard in the source code in the other places.

`let mut prop = PROPOSALS
        .load(deps.storage, proposal_id)
        .or(Err(ContractError::NoSuchProposal { id: proposal_id }))?;`

contracts/proposal/dao-proposal-single/src/contract.rs
0xJayak: Sorry, got it wrong above, but could be refactored like this to reduce one level of nesting.

` let expires = match prop.status {
        Status::Timelocked { expires } => expires,
        _ => return Err(ContractError::NotPassed {})
    };`

contracts/pre-propose/dao-pre-propose-approver/src/tests.rs
JakeHartnell: Merged! Thank you!

packages/dao-voting/src/timelock.rs
```diff
@@ -0,0 +1,102 @@
...
+    TimelockedExpired {},
+
+    #[error("Only vetoer can veto a proposal.")]
+    Unauthorized {},
+}
+
+#[cw_serde]
+pub struct Timelock {
+    /// The time duration to delay proposal execution for
+    pub delay: Timestamp,
```
JakeHartnell: Change to duration.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -255,6 +256,111 @@ pub fn execute_propose(
...
+                    let proposal_status_changed_hooks = proposal_status_changed_hooks(
+                        PROPOSAL_HOOKS,
+                        deps.storage,
+                        proposal_id,
+                        old_status.to_string(),
+                        prop.status.to_string(),
+                    )?;
+
+                    // Add prepropose / deposit module hook which will handle deposit refunds.
+                    let proposal_creation_policy = CREATION_POLICY.load(deps.storage)?;
```
JakeHartnell: We should consider adding a new policy `NO_REFUND_ON_VETO`? Technically might be an extra case.

JakeHartnell: Maybe rename. : )

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -255,6 +256,111 @@ pub fn execute_propose(
...
+                // If timelock is not configured throw error. This should never happen.
+                None => Err(ContractError::TimelockError(TimelockError::NoTimelock {})),
+            }
+        }
+        Status::Timelocked { expires } => {
+            match prop.timelock {
+                Some(ref timelock) => {
+                    // Check if the proposal timelock has expired, vetoer cannot veto
+                    // after expiration
+                    timelock.check_is_expired(env.block.time, expires)?;
```
JakeHartnell: Not sure if this warrants an option... open to ideas.

contracts/proposal/dao-proposal-single/src/contract.rs
bekauz: I think it will be easier to add this after or with the prop-multi veto pr?

bekauz:
@NoahSaso @JakeHartnell I think it's mostly ready. I'll give it another pass in a few days for the final cleanup.

One question though: do we want to accept votes when prop is `Timelocked` or `Vetoed`? atm voting is allowed until the prop `expiration` (not the timelock expiration but the prop itself) is active. Somewhat similar logic to how we allow voting when prop is passed early before expiration so that any members late for voting can still cast their votes to express their opinion.

JakeHartnell:
> One question though: do we want to accept votes when prop is Timelocked or Vetoed? atm voting is allowed until the prop expiration (not the timelock expiration but the prop itself) is active. Somewhat similar logic to how we allow voting when prop is passed early before expiration so that any members late for voting can still cast their votes to express their opinion.

Hmmmm, on the fence about this... guess we should make it consistent?

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -477,6 +481,7 @@ pub fn execute_vote(
     // cause a different one. This then serves to allow
     // for better tallies of opinions in the event that a
     // proposal passes or is rejected early.
+    // benskey TODO: check for potential timelock issues
```
JakeHartnell: TODO : )

contracts/proposal/dao-proposal-single/README.md
```diff
@@ -59,3 +59,49 @@ handling a hook.
...
+
+If the vetoer address is another DAO, this duration should be carefully
+considered because of the DAO voting period.
+
+### `vetoer`
+
+Vetoer (`String`) is the address allowed to veto the proposals that are in
+`Timelocked` state.
+
+Vetoer address can be updated via a regular proposal config update.
```
JakeHartnell, line 95: Helpful note:
```
If you want the `vetoer` role to be shared between multiple organizations or individuals, a [cw1-whitelist](https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw1-whitelist) contract address can be used to allow multiple accounts to veto the prop.
```

contracts/proposal/dao-proposal-single/src/contract.rs
JakeHartnell: Great.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -855,56 +968,53 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
...
-        MigrateMsg::FromV1 {
-            close_proposal_on_execution_failure,
-            pre_propose_info,
-        } => {
+        MigrateMsg::FromV2 { timelock } => {
             // `CONTRACT_VERSION` here is from the data section of the
             // blob we are migrating to. `version` is from storage. If
             // the version in storage matches the version in the blob
             // we are not upgrading.
+            // TODO fix this check!
```
JakeHartnell: Remove?

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -855,56 +968,53 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
...
                 return Err(ContractError::AlreadyMigrated {});
             }
 
-            // Update the stored config to have the new
-            // `close_proposal_on_execution_falure` field.
-            let current_config = v1::state::CONFIG.load(deps.storage)?;
+            let parsed_version = Version::from_str(&version)
+                .map_err(|_| ContractError::MigrationVersionError {})?;
+            // migration is only possible from 2.0.0
+            if parsed_version < Version::new(2, 0, 0) {
```
JakeHartnell: Also error if it's greater than the current version?

noahsaso:
> > One question though: do we want to accept votes when prop is Timelocked or Vetoed? atm voting is allowed until the prop expiration (not the timelock expiration but the prop itself) is active. Somewhat similar logic to how we allow voting when prop is passed early before expiration so that any members late for voting can still cast their votes to express their opinion.
> 
> Hmmmm, on the fence about this... guess we should make it consistent?

yeah i'm on the fence too.

on one hand, it feels weird that some may get to vote (and have their vote recorded/displayed) while others who did not vote before it was vetoed (pre-expiration) do not get to express an opinion.. but on the other hand, it feels meaningless to vote after a veto, so i foresee some people not voting while others vote anyway, leading to an outcome that is not representative of the members' opinions.

maybe it shouldn't be veto-able until after the prop is decided (i.e. it is early-passed because revoting is disabled, or it expires). or alternatively, votes could be deleted if it's vetoed early (that feels pretty weird too though). wdyt about that?

Cargo.toml
```diff
@@ -116,13 +117,21 @@ dao-voting-cw721-staked = {path = "./contracts/voting/dao-voting-cw721-staked",
...
+voting-v1 = { package = "dao-voting", version = "0.1.0" }
+stake-cw20-v03 = { package = "stake-cw20", version = "0.2.6" }
+
+# v2 dependencies. used for state migrations
+cw-utils-v2 = { package = "cw-utils", version = "0.16" }
+dao-dao-core-v2 = { package = "dao-dao-core", version = "2.2.0" }
+dao-interface-v2 = { package = "dao-interface", version = "2.2.0" }
+dao-proposal-single-v2 = { package = "dao-proposal-single", version = "2.2.0" }
+dao-pre-propose-single-v2 = { package = "dao-pre-propose-single", version = "2.2.0" }
+voting-v2 = { package = "dao-voting", version = "2.2.0" }
```
noahsaso: should we name these more explicitly (like `v220`)? surely we will have future migration dependencies that will require renaming these

contracts/external/dao-migrator/Cargo.toml
```diff
@@ -27,13 +27,14 @@ cw20             = { workspace = true }
 dao-interface    = { workspace = true }
 
 dao-dao-core               = { workspace = true, features = ["library"] }
-dao-voting             = { workspace = true }
-dao-proposal-single    = { workspace = true, features = ["library"] }
+# dao-voting             = { workspace = true }
+# dao-proposal-single    = { workspace = true, features = ["library"] }
```
noahsaso: should these be removed?

contracts/external/dao-migrator/examples/schema.rs
```diff
@@ -1,11 +1,10 @@
 use cosmwasm_schema::write_api;
-use dao_proposal_single::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg};
+use dao_migrator::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
```
noahsaso, line 2: lmao

contracts/proposal/dao-proposal-single/src/testing/contracts.rs
```diff
@@ -39,16 +39,16 @@ pub(crate) fn cw20_stake_contract() -> Box<dyn Contract<Empty>> {
...
+// pub(crate) fn v2_proposal_single_contract() -> Box<dyn Contract<Empty>> {
+//     let contract = ContractWrapper::new(
+//         dao_proposal_single_v2::contract::execute,
+//         dao_proposal_single_v2::contract::instantiate,
+//         dao_proposal_single_v2::contract::query,
+//     )
+//     .with_reply(dao_proposal_single_v2::contract::reply)
+//     .with_migrate(dao_proposal_single_v2::contract::migrate);
+//     Box::new(contract)
+// }
```
noahsaso: is this still needed?

contracts/proposal/dao-proposal-single/src/state.rs
```diff
@@ -55,6 +58,8 @@ pub struct Config {
     /// remain open until the DAO's treasury was large enough for it to be
     /// executed.
     pub close_proposal_on_execution_failure: bool,
+    /// Optional time delay on proposal execution
+    pub timelock: Option<Timelock>,
```
noahsaso:
```suggestion
    /// Optional time delay on proposal execution during which the configured vetoer can veto the proposal.
    pub timelock: Option<Timelock>,
```

my initial thought was this comment could be improved, but i think `timelock` is just quite vague in general. what if we just call it `veto`? and the new status enum variant can be `VetoTimelock` instead of `Timelock`.

i think in general the `timelock` naming scheme could use a refactor to `veto` since that's really what it is

contracts/proposal/dao-proposal-single/src/proposal.rs
```diff
@@ -61,14 +70,20 @@ impl SingleChoiceProposal {
...
+        match self.status {
+            Status::Open if self.is_passed(block) => match &self.timelock {
+                // if prop is passed and time lock is configured,
+                // calculate lock expiration and set status to `Timelocked`.
+                Some(timelock) => Status::Timelocked {
+                    expiration: timelock.delay.after(block),
+                },
+                // Otherwise the proposal is simply passed
+                None => Status::Passed,
+            },
```
noahsaso: this doesn't ever return `passed` if the configured timelock delay has expired. i feel like it would make sense that this switches to `passed` instead of returning `timelocked` with an `expiration` field in the past. currently this requires the expiration to be explicitly checked everywhere. on the other hand, leaving it as timelocked allows for more granular errors when attempting to execute or veto. wdyt?

contracts/proposal/dao-proposal-single/src/msg.rs
```diff
@@ -199,26 +212,12 @@ pub enum QueryMsg {
...
-        /// failure. For example, with this enabled a proposal to send 5
-        /// tokens out of a DAO's treasury with 4 tokens would be closed when
-        /// it is executed. With this disabled, that same proposal would
-        /// remain open until the DAO's treasury was large enough for it to be
-        /// executed.
-        close_proposal_on_execution_failure: bool,
+    FromV2 {
         /// This field was not present in DAO DAO v1. To migrate, a
         /// value must be specified.
         ///
```
noahsaso: comment no longer accurate for v2 timelock migration

contracts/proposal/dao-proposal-single/src/msg.rs
```diff
@@ -199,26 +212,12 @@ pub enum QueryMsg {
 
 #[cw_serde]
 pub enum MigrateMsg {
-    FromV1 {
```
noahsaso: is there a reason to remove the v1 migration? is it difficult to support both?

contracts/proposal/dao-proposal-single/src/v2_state.rs
```diff
@@ -0,0 +1,129 @@
...
+        no: v2.no,
+        abstain: v2.abstain,
+    }
+}
+
+pub fn v2_status_to_v3(v2: voting_v2::status::Status) -> Status {
+    match v2 {
+        voting_v2::status::Status::Open => Status::Open,
+        voting_v2::status::Status::Rejected => Status::Rejected,
+        voting_v2::status::Status::Passed => todo!(),
```
noahsaso: `todo!()` needs to be replaced

contracts/proposal/dao-proposal-single/src/v2_state.rs
```diff
@@ -0,0 +1,129 @@
...
+            };
+        }
+
+        status_conversion!(Status::Open);
+        status_conversion!(Status::Closed);
+        status_conversion!(Status::Executed);
+        status_conversion!(Status::Rejected);
+        status_conversion!(Status::ExecutionFailed);
+        // TODO test passed status conversion
+        // status_conversion!(Status::Passed);
```
noahsaso: merp

packages/dao-voting/src/timelock.rs
```diff
@@ -0,0 +1,73 @@
...
+    Unauthorized {},
+}
+
+#[cw_serde]
+pub struct Timelock {
+    /// The time duration to delay proposal execution for.
+    pub delay: Duration,
+    /// The address able to veto proposals.
+    pub vetoer: String,
+    /// Whether or not the vetoer can excute a proposal early before the
```
noahsaso:
```suggestion
    /// Whether or not the vetoer can execute a proposal early before the
```

contracts/proposal/dao-proposal-single/src/contract.rs
noahsaso: seems like `may_load(..)?.ok_or(..)` does the same thing 🤷 just a style difference i guess

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -255,6 +258,114 @@ pub fn execute_propose(
...
+                    )?;
+
+                    Ok(Response::new()
+                        .add_attribute("action", "veto")
+                        .add_attribute("proposal_id", proposal_id.to_string())
+                        .add_submessages(proposal_status_changed_hooks)
+                        .add_submessages(proposal_completed_hooks))
+                }
+                // If timelock is not configured throw error
+                None => Err(ContractError::TimelockError(TimelockError::NoTimelock {})),
```
noahsaso: would it be better to check this one time at the beginning of the function instead of matching inside each status branch? if so, you could also move the `check_is_vetoer` calls to the top.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -255,6 +258,114 @@ pub fn execute_propose(
...
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    proposal_id: u64,
+) -> Result<Response, ContractError> {
+    let mut prop = PROPOSALS
+        .may_load(deps.storage, proposal_id)?
+        .ok_or(ContractError::NoSuchProposal { id: proposal_id })?;
+
+    match prop.status {
```
noahsaso, line 287: i believe this function is missing a `prop.update_status(&env.block);` call at the beginning before checking its status. the status stored in the proposal state may not be up to date based on the latest voting outcome if the voting period did not end early (like if revoting is allowed or quorum is not met). both the `execute` and `close` functions have this pattern where they update before reading the status, so i assume it is necessary here too.

for example: if the proposal is rejected due to lack of participation (quorum not met), this veto call will still see the proposal as open and try to veto it (if it has not yet been closed). calling `update_status` should ensure that it is up to date before it is checked.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -273,18 +384,43 @@ pub fn execute_execute(
             &config.dao,
             Some(prop.start_height),
         )?;
-        if power.is_zero() {
+        let vetoer_call = match config.timelock {
```
noahsaso:
```suggestion
        let vetoer_call = match prop.timelock {
```

use `timelock` from proposal instead of config

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -273,18 +384,43 @@ pub fn execute_execute(
...
     prop.update_status(&env.block);
-    if prop.status != Status::Passed {
-        return Err(ContractError::NotPassed {});
+    let old_status = prop.status;
+    match prop.status {
+        Status::Passed => (),
+        Status::Timelocked { expiration } => {
+            // we can only end up in `Timelocked` state if `timelock`
+            // is not `None`
+            if let Some(ref timelock) = prop.timelock {
```
noahsaso: this comment is true, but everywhere else we check anyways and error if not, so might as well do it here to be insanely careful

packages/dao-hooks/src/all_hooks.rs
```diff
@@ -14,6 +14,8 @@ pub enum DaoHooks {
     MemberChangedHook(MemberChangedHookMsg),
     /// Called when NFTs are staked or unstaked.
     NftStakeChangeHook(NftStakeChangedHookMsg),
+    /// Pre-propose hooks
+    PreProposeHook(PreProposeHookMsg),
```
noahsaso, line 18: why is this being added here now? is it related to veto somehow?

packages/dao-pre-propose-base/src/execute.rs
```diff
@@ -286,19 +286,28 @@ where
         // bizare has happened. In that event, this message errors
         // which ought to cause the proposal module to remove this
         // module and open proposal submission to anyone.
-        if new_status != Status::Closed && new_status != Status::Executed {
+        if new_status != Status::Closed
+            && new_status != Status::Executed
+            && new_status != Status::Vetoed
+        {
             return Err(PreProposeError::NotClosedOrExecuted { status: new_status });
```
noahsaso: wdyt about updating this error to be `NotClosedExecutedOrVetoed`? or just `NotCompleted`? with a similar change in the error string of course

noahsaso (changes requested):
this is sooooo cool, super excited for this!!!!! and great job.

i have a handful of questions and a few bugs i believe need correcting. also, two discussion points (that are both covered by conversations i opened in this review, just surfacing them here for visibility):
1. i think we should rename timelock entirely to veto. timelock is ambiguous and somewhat distracting. i think it makes sense for the status to be `VetoTimelock` as that describes the state of a proposal, but the config option being called `veto` makes way more sense to me than `timelock`. thoughts?
2. should the status of proposals once the veto timelock expires revert back to passed? or should it stay timelocked with an expired expiration field? i'm not really sure. i mentioned some pros and cons in the conversation where i thought of it.

packages/dao-hooks/src/all_hooks.rs
JakeHartnell: Consistency. "All hooks" should contain all hooks.

packages/dao-pre-propose-base/src/execute.rs
JakeHartnell: I'm a fan of `NotCompleted`. You or @bekauz want to pick that up?

contracts/proposal/dao-proposal-single/src/contract.rs
JakeHartnell: Agreed.

contracts/proposal/dao-proposal-single/src/testing/contracts.rs
JakeHartnell: At some point it will be when we want to do contract migrations and tests, but let's kill it for now.

contracts/external/dao-migrator/Cargo.toml
JakeHartnell: We need to think about a good migration pattern going forward... these are needed if we want the migrator to support `v1 -> v2`, `v2 -> v3`, etc... but in favor of removing them for now.

Cargo.toml
JakeHartnell: Migration needs serious thought... we need a good pattern if we're going to move fast.

JakeHartnell:
> this is sooooo cool, super excited for this!!!!! and great job.
> 
> i have a handful of questions and a few bugs i believe need correcting. also, two discussion points (that are both covered by conversations i opened in this review, just surfacing them here for visibility):

Thanks for the awesome and helpful bug discovery and comments!

> 1. i think we should rename timelock entirely to veto. timelock is ambiguous and somewhat distracting. i think it makes sense for the status to be `VetoTimelock` as that describes the state of a proposal, but the config option being called `veto` makes way more sense to me than `timelock`. thoughts?

Yes. 100%.

> 2. should the status of proposals once the veto timelock expires revert back to passed? or should it stay timelocked with an expired expiration field? i'm not really sure. i mentioned some pros and cons in the conversation where i thought of it.

Hmmmmmm, this is a really good question... going to think about this one while addressing low hanging fruit.

contracts/proposal/dao-proposal-single/src/contract.rs
JakeHartnell: Very good catch.

JakeHartnell: Very good catch.

contracts/proposal/dao-proposal-single/src/contract.rs
bekauz: good one, will add!

packages/dao-pre-propose-base/src/execute.rs
bekauz: yep!

contracts/proposal/dao-proposal-single/src/testing/contracts.rs
JakeHartnell: Deleted.

contracts/external/dao-migrator/Cargo.toml
JakeHartnell: Done.

contracts/proposal/dao-proposal-single/src/testing/contracts.rs
noahsaso: > At some point it will be when we want to do contract migrations and tests, but let's kill it for now.

wait we should definitely be doing contract migration and tests. i'm confused, we've been doing migrations and tests the whole time right? then we shouldn't delete this.

contracts/external/dao-migrator/Cargo.toml
noahsaso: > We need to think about a good migration pattern going forward... these are needed if we want the migrator to support `v1 -> v2`, `v2 -> v3`, etc... but in favor of removing them for now.

well if they're for migrating then we should not remove them right?

contracts/proposal/dao-proposal-single/src/contract.rs
bekauz: done! nice one

contracts/proposal/dao-proposal-single/src/contract.rs
bekauz: checking and returning an err now

contracts/proposal/dao-proposal-single/src/contract.rs
bekauz: `may_load()` returns the data if it's there, none if it's not there, or an error if there are any parsing issues.

`load()` returns the data if it's there, or an error in case there is no data or there are any parsing issues.

in some cases it doesn't really matter, but in most cases it's nice to distinguish whether there is nothing stored from an error.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -855,56 +952,58 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
 
     match msg {
-        MigrateMsg::FromV1 {
-            close_proposal_on_execution_failure,
-            pre_propose_info,
-        } => {
```
JakeHartnell: Let's restore this.

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -855,56 +952,58 @@ pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, Co
...
+        MigrateMsg::FromV2 { veto } => {
+            let version_2 = Version::new(2, 0, 0);
+
```
JakeHartnell: Let's kill this for now. Will support migrations at the next major version.

If a DAO really wants to use this, they can just swap out a new proposal module with veto from the previous one they were using.

Cargo.toml
```diff
@@ -116,13 +116,23 @@ dao-voting-cw721-staked = {path = "./contracts/voting/dao-voting-cw721-staked",
...
+stake-cw20-v03 = { package = "stake-cw20", version = "0.2.6" }
+
+
+# v2 dependencies. used for state migrations
+cw-utils-v2 = { package = "cw-utils", version = "0.16" }
+dao-dao-core-v2 = { package = "dao-dao-core", version = "2.2.0" }
+dao-interface-v2 = { package = "dao-interface", version = "2.2.0" }
+dao-proposal-single-v2 = { package = "dao-proposal-single", version = "2.2.0" }
+dao-pre-propose-single-v2 = { package = "dao-pre-propose-single", version = "2.2.0" }
+voting-v2 = { package = "dao-voting", version = "2.2.0"}
```
noahsaso: are these still needed now that we reverted the v2 migration?

contracts/external/dao-migrator/Cargo.toml
```diff
@@ -43,7 +42,14 @@ cw20-stake-v1                 = { workspace = true, features = ["library"] }
...
 cw20-v1                       = { version = "0.13", package = "cw20" }
-cw4-v1                       = { version = "0.13", package = "cw4" }
+cw4-v1                        = { version = "0.13", package = "cw4" }
+
+# v2 migration
+dao-dao-core-v2 = { workspace = true }
+dao-interface-v2 = { workspace = true }
+cw-utils-v2 = { workspace = true }
+dao-proposal-single-v2 = { workspace = true, features = ["library"] }
+voting-v2 = { workspace = true}
```
noahsaso: are these still needed now that we reverted the v2 migration?

noahsaso (changes requested):
sick so so close

noahsaso (approved):

PR #763, opened 2023-11-10
Fix tests from Massv

Dat-Andre:


Dat-Andre:
Reverted the bumps on [osmosis-std, osmosis-test-tube and prost because the Integration Tests fail

JakeHartnell:
Merged this into #764

PR #766, opened 2023-11-14
Add missing query to dao-pre-propose-approver

noahsaso:
The `dao-pre-propose-approver` contract does not currently provide any query to determine which pre-propose ID corresponds with a given proposal. In other words, you cannot check which pending proposal an approval proposal was made for.

This PR adds a query `PendingProposalIdForApprovalProposalId` which takes the `id` of the approval proposal being voted on by the approval DAO and returns the `id` of the pending proposal in the `dao-pre-propose-single` contract on the DAO waiting for proposal approval.

### DIFF
--- a/contracts/pre-propose/dao-pre-propose-approver/src/contract.rs
+++ b/contracts/pre-propose/dao-pre-propose-approver/src/contract.rs
@@ -188,6 +188,9 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
             QueryExt::PreProposeApprovalContract {} => {
                 to_binary(&PRE_PROPOSE_APPROVAL_CONTRACT.load(deps.storage)?)
             }
+            QueryExt::PendingProposalIdForApprovalProposalId { id } => {
+                to_binary(&PROPOSAL_IDS.load(deps.storage, id)?)
+            }
         },
         _ => PrePropose::default().query(deps, env, msg),
     }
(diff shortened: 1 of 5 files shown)


contracts/pre-propose/dao-pre-propose-approver/src/contract.rs
```diff
@@ -188,6 +188,9 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
             QueryExt::PreProposeApprovalContract {} => {
                 to_binary(&PRE_PROPOSE_APPROVAL_CONTRACT.load(deps.storage)?)
             }
+            QueryExt::PendingProposalIdForApprovalProposalId { id } => {
+                to_binary(&PROPOSAL_IDS.load(deps.storage, id)?)
```
bekauz: Nitpicking outside of the pr scope but I would rename that `PROPOSAL_IDS` map to something a little more obvious, like `PROPOSAL_ID_TO_PRE_PROPOSE_ID_MAP`

bekauz: Shall we do a `may_load` here to avoid errors in case it's not set?

bekauz (commented):
just a few comments. not 100% sure about renaming. what do you think about returning an `Option<u64>` instead?

contracts/pre-propose/dao-pre-propose-approver/src/contract.rs
JakeHartnell: Agree with the `may_load`.

JakeHartnell (approved):
Nice! Agree with the comment above about `may_load` for the query, but otherwise LGTM.

contracts/pre-propose/dao-pre-propose-approver/src/contract.rs
noahsaso: > Shall we do a `may_load` here to avoid errors in case it's not set?

sounds good

bekauz (approved):
lets go!

PR #767, opened 2023-11-15
Added missing completed proposal queries to approval-single

noahsaso:
The `dao-pre-propose-approval-single` contract does not store pending proposals once they have been approved or rejected. This makes it impossible to retrieve past decisions and connect approval proposals used in `dao-pre-propose-approver` to existing proposals.

This PR adds queries to check whether or not a given approval ID is pending, and allows querying completed proposals. If approved, it also includes the created proposal ID. It also adds a query to get the completed pre-propose ID from the created proposal ID.


JakeHartnell (approved):
LGTM

PR #770, opened 2023-11-16
Remove requirement that vesting payment owner matches factory

noahsaso:
This removes the requirement that vesting payments must be created with the same owner as the payroll factory. This allows for creating vesting payments that are cancellable by someone else, or multiple accounts, while still restricting who is allowed to create new vesting payments on that factory.


JakeHartnell (approved):
lgtm

PR #776, opened 2023-11-29
 Auto Pass Props if User Controls 100% of Voting power

nickf12:
Hi @JakeHartnell 
This should close #775

### DIFF
--- a/contracts/proposal/dao-proposal-single/src/contract.rs
+++ b/contracts/proposal/dao-proposal-single/src/contract.rs
@@ -1,7 +1,7 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    to_binary, Addr, Binary, CosmosMsg, Deps, DepsMut, Empty, Env, MessageInfo, Order, Reply,
+    to_json_binary, Addr, Binary, CosmosMsg, Deps, DepsMut, Empty, Env, MessageInfo, Order, Reply,
     Response, StdResult, Storage, SubMsg, WasmMsg,
 };
 use cw2::{get_contract_version, set_contract_version, ContractVersion};
@@ -183,14 +183,6 @@ pub fn execute_propose(
         &dao_interface::msg::QueryMsg::VotingModule {},
     )?;
 
-    // Get Proposer voting power
-    let power = get_voting_power(
-        deps.as_ref(),
-        proposer.clone(),
-        &config.dao,
-        Some(env.block.height),
-    )?;
-
     // Voting modules are not required to implement this
     // query. Lacking an implementation they are active by default.
     let active_resp: IsActiveResponse = deps
@@ -208,7 +200,6 @@ pub fn execute_propose(
 
     let proposal = {
         // Limit mutability to this block.
-        // If proposer has voting power = total_power. AutoPass proposal
         let mut proposal = SingleChoiceProposal {
             title,
             description,
@@ -219,11 +210,7 @@ pub fn execute_propose(
             threshold: config.threshold,
             total_power,
             msgs,
-            status: if power == total_power {
-                Status::Passed
-            } else {
-                Status::Open
-            },
+            status: Status::Open,
             votes: Votes::zero(),
             allow_revoting: config.allow_revoting,
         };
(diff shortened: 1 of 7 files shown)


noahsaso:
can you rebase on development? i think this PR is undoing some of the recent fixes

nickf12:
Hi @NoahSaso,
I think this was the previous PR that I closed and rebased my branch.
I don't see any commit behind my branch so not sure what I need to rebase.

nickf12:
Rebased! Please let me know if there is any other conflict

noahsaso:
hey @nickf12 i'm concerned about the tests being modified. instead of modifying tests, could you add new tests that specifically cover this feature? that way we can be 100% confident that this doesn't break anything.

also, @JakeHartnell @bekauz, i'm curious what you think about the following:
1. this currently results in a passed proposal with 0 votes cast. do you think this should auto-submit a yes vote? i'm leaning towards it should.
2. i'm a bit uncomfortable setting the status to `Passed` right away as i'm worried about forgotten hooks and future logic. IMO it'd be *much* safer and future-proof to leave the existing `execute_propose` code as-is and then call the `execute_vote` function at the end if the proposer has all voting power, which takes care of adding the ballot, updating the status, and firing the proposal status change and new vote hooks. if we take that route instead, there should probably be a special case added to update the proposal into passed manually when `allow_revoting` is enabled, since just casting a vote wouldn't be sufficient to pass the proposal.
4. this feature should be enabled by a config setting, like `auto_pass_when_only_member_proposes`.

bekauz:
first, this needs to be rebased again.

i also share @NoahSaso concerns about modifying the tests and bypassing the `Open` prop state. imho it would be a little weird seeing a proposal that passed without any votes.

if we do want this i think some things need to be considered:
- what if revoting is enabled?
- is it ok to potentially execute a prop before min_voting_period expires?
- is the `proposer_power == total_power` check the right one or should we check the threshold as well?
- what if veto is configured?

noahsaso:
> first, this needs to be rebased again.
> 
> i also share @NoahSaso concerns about modifying the tests and bypassing the `Open` prop state. imho it would be a little weird seeing a proposal that passed without any votes.
> 
> if we do want this i think some things need to be considered:
> 
> * what if revoting is enabled?
> * is it ok to potentially execute a prop before min_voting_period expires?
> * is the `proposer_power == total_power` check the right one or should we check the threshold as well?
> * what if veto is configured?

yeah, great points. i forgot about veto and min_voting_period. which is kind of exactly my point, this feels like an idea that sounds nice in theory but is going to make things very complicated very quickly, to a concerning degree, and make future feature changes harder.

also, if we can solve something in the frontend without making a smart contract change, i'm going to make a generalization (which i.. generally.. avoid) and claim we should always implement it via the frontend. we can solve this by adding a vote message to the same transaction as the proposal creation message. technically if someone submitted a proposal the instant before, the vote could be cast to the wrong proposal, but that's very low probability and safer than implementing this in the smart contract IMO

bekauz:
i think adding an option to automatically cast a vote on proposal creation would make more sense.

proposals could be configured to allow atomic prop-then-vote. if that's the case, the proposer can choose to opt into it and cast their vote or remain neutral and just create a proposal.

this way we are just combining two actions that already exist, so no changes to the existing status flow should be needed.

noahsaso:
> i think adding an option to automatically cast a vote on proposal creation would make more sense.
> 
> proposals could be configured to allow atomic prop-then-vote. if that's the case, the proposer can choose to opt into it and cast their vote or remain neutral and just create a proposal.
> 
> this way we are just combining two actions that already exist, so no changes to the existing status flow should be needed.

i love this solution, brilliant. people have asked for this in the past.

bekauz:
> proposals could be configured to allow atomic prop-then-vote

on a second thought, is this even needed? if for any reason a dao chose not want to enable this feature, the proposer could just create the prop and cast their vote on it anyways.

noahsaso:
> > proposals could be configured to allow atomic prop-then-vote
> 
> on a second thought, is this even needed? if for any reason a dao chose not want to enable this feature, the proposer could just create the prop and cast their vote on it anyways.

oh yeah, i guess there's no need to implement a config option for this.. nice

noahsaso:
hey, bump on this @nickf12, do you think you could change this PR to just add a parameter to `propose` so that the proposer can optionally cast a vote on proposal creation?

noahsaso:
closing in favor of #799

PR #778, opened 2023-12-04
Update contracts to v2.4.0 with audit veto fix

noahsaso:
This updates all the contract versions to v2.4.0. It also includes a fix from the veto audit.


JakeHartnell (approved):
:tada:

PR #780, opened 2023-12-11
Fix vetoer bypassing only_members_execute flag

noahsaso:
This fixes a minor bug found during the Oak Security veto audit in which the vetoer could bypass the `only_members_execute` configuration, executing a proposal even if they are not a member of the DAO, after the veto timelock expires.


JakeHartnell (approved):
LGTM!

Issue #710, opened 2023-06-14
Enable optional vesting of staking rewards in `cw-vesting`


JakeHartnell:
User story: core team of a project also wants to vest staking rewards to demonstrate alignment to the project.

Issue #677, opened 2023-03-14
Allow `cw-vesting` to register multiple slashes at once


NoahSaso:
Right now, the vesting contract allows registering one slash at a time. Because slashes will likely be registered via DAO proposals, and because DAO proposal actions map 1:1 with CosmWasm messages, it would greatly improve the frontend UX to allow registering multiple slashes in one message. Often there will only be one slash registered at a time, but in the event there is more than one, we don't want to confuse the user by adding multiple actions that all look the same (or worse, adding a dropdown to select which slash to register...).

PR #782, opened 2023-12-12
Extend `cw-vesting` to allow for staking in DAOs

### LINKED ISSUE #779: `cw-vesting` should allow for staking in DAOs
Currently, `cw-vesting` supports staking tokens natively on Cosmos chains, we need something similar but designed for staking in token DAOs.

As a user who receives vested tokens from a DAO, I would like to be able to stake those tokens to participate in governance.

JakeHartnell:
Closes #779

This topic keeps coming up in spaces and I wanted to explore how difficult it would be. This PR is a strawman solution (couldn't help myself :joy:). DO NOT MERGE.

Some observations:
- Need to restrict which staking contracts a vesting contract can stake with as that could be a way to sneak out unvested funds if the staking contract is not trusted.
- Not supporting the ability to propose drastically simplifies things as we don't have to worry about deposits getting refunded.
- Not supporting cw20 greatly simplifies things, but we may want to add support for that for use with the $DAO token?
- Doesn't really impact core logic too much and so should be relatively safe and cheep to audit.
- Frontend would certainly require work with this approach... but hard to imagine an approach where that wouldn't be the case.


noahsaso:
Features derived from list above and discussion in Monday call:
- Mutable allowlist of contracts the staking contract can interact with (should be instantiated with the voting module / staking contract)
- Ability to passthrough wasm messages to the allowlisted contracts, so that the owner can send returned deposits back to themself for example
- Flag to enable or disable staking (maybe just via the allowlist being empty or not)
- Multiplier for voting power derived from remaining vesting
- Require proposal deposits to be already vested (i.e. liquid and transferrable)

JakeHartnell:
Related issues we may also want to consider when working on this:
- https://github.com/DA0-DA0/dao-contracts/issues/710
- https://github.com/DA0-DA0/dao-contracts/issues/677

JakeHartnell:
Limits on voting power for vested tokens could be achieved with #746.

noahsaso:
Closing as this is not the design we are going to pursue to add voting support for vesting tokens.

PR #786, opened 2023-12-17
Fix voting module native token proposal deposit on instantiate

### LINKED ISSUE #765: Cannot setup proposal deposit on token DAO creation when creating a new Token Factory
When creating a new token factory DAO, using `voting_module_token` as the denom for the proposal deposit fails because `UncheckedDepositInfo`'s `into_checked` function assumes that the `voting_module_token` is a cw20. This function needs to be updated to allow specifying which token type the voting module uses so it can validate accordingly.

The workaround is to create the DAO first and then setup a proposal deposit after.

noahsaso:
Resolves #765 

This allows setting up a proposal deposit using the newly created voting module token on DAO instantiation.

### DIFF
--- a/packages/dao-dao-macros/src/lib.rs
+++ b/packages/dao-dao-macros/src/lib.rs
@@ -152,16 +152,16 @@ pub fn voting_module_query(metadata: TokenStream, input: TokenStream) -> TokenSt
 }
 
 /// Adds the necessary fields to an enum such that it implements the
-/// interface needed to be a voting module with a token.
+/// interface needed to be a voting module with a cw20 token.
 ///
 /// For example:
 ///
 /// ```
-/// use dao_dao_macros::token_query;
+/// use dao_dao_macros::cw20_token_query;
 /// use cosmwasm_schema::{cw_serde, QueryResponses};
 /// use cosmwasm_std::Addr;
 ///
-/// #[token_query]
+/// #[cw20_token_query]
 /// #[cw_serde]
 /// #[derive(QueryResponses)]
 /// enum QueryMsg {}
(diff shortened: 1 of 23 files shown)


packages/dao-dao-macros/src/lib.rs
```diff
@@ -210,6 +210,65 @@ pub fn token_query(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
+/// ```
+#[proc_macro_attribute]
+pub fn native_token_query(metadata: TokenStream, input: TokenStream) -> TokenStream {
+    merge_variants(
+        metadata,
+        input,
+        quote! {
+        enum Right {
+            #[returns(::dao_interface::voting::DenomResponse)]
+            Denom {}
```
JakeHartnell, line 265: This is great. `dao-voting-token-staked` DAOs do have a "Token Contract", `cw-tokenfactory-issuer` which is used for interacting with the Token Factory token and provides some extra features.

Now it's possible to get that from the denom, but requires having to parse it.

Don't feel too strongly here, just wanted to mention it. Maybe `cw20_token_query` could be just `token_contract` query? Then you could have both?

JakeHartnell (approved):
LGTM! Wish we had caught this during the first audit.

Made a small note, but don't feel too strongly about it.

packages/dao-dao-macros/src/lib.rs
noahsaso: i think one of us is misunderstanding the other 😅

the `token_contract` in `dao-voting-token-staked` and `dao-voting-cw20-staked` are fundamentally different. the `voting_module_token` variant is used exclusively when instantiating with a proposal deposit, as the pre-propose module setting up the proposal deposit needs to query to find the denom that the voting module uses (either native denom or cw20 contract) to validate into a `CheckedDenom` enum and save it in the config. this has nothing to do with the token capabilities, it just needs to retrieve the actual token being used during instantiation so that it can save the right denom for future proposal deposit validations.

JakeHartnell: Never mind me, I see the query is still there in the `msg.rs` file.

PR #788, opened 2023-12-18
Support both osmosis and CW tokenfactory msg standards

noahsaso:
This adds support for the `/cosmwasm.tokenfactory` msg type URLs to `cw-tokenfactory-issuer`, now supporting both `/cosmwasm.` and `/osmosis.`. The feature flag `osmosis_tokenfactory` or `cosmwasm_tokenfactory` control which version of the contract gets built.

It also removes dependency on the `token_bindings` package which uses `CosmosMsg::Custom(TokenFactoryMsg)` because we now use stargate (`CosmosMsg::Stargate`) messages instead, which are supported on all chains since msgs are protobuf encoded. This is critical to supporting both token the `cosmwasm.` and `osmosis.` tokenfactory standards.

### DIFF
--- a/packages/cw-tokenfactory-types/Cargo.toml
+++ b/packages/cw-tokenfactory-types/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "cw-tokenfactory-types"
+authors = ["Noah Saso <noah@daodao.zone>"]
+description = "A package for the handling of variou tokenfactory types."
+edition = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+version = { workspace = true }
+
+[features]
+default = ["osmosis_tokenfactory"]
+# use the /cosmwasm.tokenfactory... msg types
+cosmwasm_tokenfactory = []
+# use the /osmosis.tokenfactory... msg types
+osmosis_tokenfactory = []
+
+[dependencies]
+cosmwasm-std = { workspace = true }
+cosmwasm-schema = { workspace = true }
+dao-interface = { workspace = true }
+osmosis-std = { workspace = true }
+osmosis-std-derive = { workspace = true }
+prost = { workspace = true }
+prost-types = { workspace = true }
+schemars = { workspace = true }
+serde = { workspace = true }
+serde-cw-value = { workspace = true }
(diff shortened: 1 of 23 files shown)


packages/cw-tokenfactory-types/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "cw-tokenfactory-types"
+authors = ["Noah Saso <noah@daodao.zone>"]
+description = "A package for the handling of variou tokenfactory types."
```
JakeHartnell:
```suggestion
description = "A package for the handling of various tokenfactory types."
```

packages/cw-tokenfactory-types/src/cosmwasm.rs
```diff
@@ -0,0 +1,281 @@
+use std::convert::TryFrom;
+use std::convert::TryInto;
+
+use osmosis_std_derive::CosmwasmExt;
+
+// see https://github.com/notional-labs/wasmd/blob/v0.30.0-sdk469.4/proto/cosmwasm/tokenfactory/v1beta1/tx.proto
```
JakeHartnell, line 6: Why are we using the types in a fork that probably won't be maintained?

JakeHartnell: Who is using these?

noahsaso: migaloo uses `/cosmwasm.tokenfactory...` types. it's not that specific fork necessarily, that's just a link to a tag that has the protobufs. the msg types are the exact same except some of the osmosis msgs have additional fields, but mostly only the type URL differs. migaloo will probably eventually migrate over to the osmosis types, but for now they are on the cosmwasm version 🤷 i don't think it's a big deal to support both, and since they are behind feature flags, all this does is make it compilable for a wider set of chains

JakeHartnell (approved):
LGTM

packages/cw-tokenfactory-types/src/cosmwasm.rs
JakeHartnell: Oh, sounds good then. Maybe add a comment about which chain it's for? :shipit:

PR #793, opened 2024-01-04
Fix clippy errors from latest rust stable

JakeHartnell:
Fixes clippy errors that have come up with the latest rust stable release. No functionality changes.


JakeHartnell:
Silly codecov...

PR #797, opened 2024-01-31
Proposal Incentives & Voting Incentives

ismellike:
**Proposal Incentives Contract (dao-proposal-incentives):**

A new contract has been added to implement incentives for making successful proposals in a DAO. This contract allows DAO's to offer rewards for successful proposals, which can be configured as native or CW20 tokens. The rewards are determined based on the configuration at the proposal's start time.

A new query has been added to the proposal modules for `GenericProposalInfo`, which we can call from a received hook to get a proposal's `proposer` and `start_time`.

A new query has been added to the DAO core for `ProposalModule`, which we can call from the incentives contract to check authorization of the sender as a proposal module of the DAO.

**Voting Incentives Contract (dao-voting-incentives):**

The voting incentives contract is adapted from the proposal incentives contract.

When a vote hook is received, its proposal is queried to make sure the proposal occurred after the contract was instantiated (`config.start_height`). If the user has already voted, then we ignore the vote. Otherwise, we increment `config.total_vote_count` and the user's votes. From this, we can determine how much a user will be awarded after the voting incentives' expiration. If no users have voted during the voting incentives' period, then the funds will be sent to the contract's owner on expiration.

Anyone can call the expire function to set the amount available for distribution. Users can then call `ExecuteMsg::Claim{}` to receive their rewards.


ismellike:
@JakeHartnell

contracts/external/dao-proposal-incentives/src/execute.rs
```diff
@@ -0,0 +1,110 @@
...
+        // Validate the message is coming from a proposal module of the owner (DAO)
+        deps.querier.query_wasm_smart::<ProposalModule>(
+            owner,
+            &dao_interface::msg::QueryMsg::ProposalModule {
+                address: info.sender.to_string(),
+            },
+        )?;
+
+        // Check prop status and type of hook
+
```
JakeHartnell, line 38:
```suggestion
```

JakeHartnell:
@ismellike, the proposal incentives contract looks great!

contracts/external/dao-voting-incentives/src/tests.rs
```diff
@@ -0,0 +1,470 @@
...
+                description: "Testing".to_string(),
+                msgs: vec![],
+                proposer: None,
+            }),
+            &[],
+        )
+        .unwrap();
+    let results = vote_yes_on_proposal(&mut context, 3u64);
+    for (i, result) in results.iter().enumerate() {
+        assert!(result.is_ok());
```
ismellike, line 447: This actually panics, because the hooks simultaneously return an error of expiration.
When the error is processed in the proposal module, it tries to remove by index on a list that has changed already.

```
thread 'tests::test_hooks' panicked at packages/cw-hooks/src/lib.rs:60:26:
removal index (is 1) should be < len (is 1)
```

VoteHook of index 0 is successfully removed at index 0, but VoteHook of index 1 tries to be removed at index 1 of a list of length 1.

ismellike: I added a new method in cw-hooks for `remove_hook_by_index_from_reply` that takes a new parameter `removed_hook_indexes: Item<Vec<u64>>`. 

We can adjust the given index to the actual index from this new tracked state on the proposal modules: `REMOVED_VOTE_HOOKS_BY_INDEX` or `REMOVED_PROPOSAL_HOOKS_BY_INDEX`

contracts/external/dao-voting-incentives/src/msg.rs
```diff
@@ -1,42 +1,56 @@
...
+    #[returns(Config)]
     Config {},
-    /// Returns the rewards for the given address.
-    #[returns(cosmwasm_std::Uint128)]
+    /// Returns the claimable rewards for the given address.
+    #[returns(RewardResponse)]
     Rewards { address: String },
+    /// Returns the expected rewards for the given address
+    #[returns(RewardResponse)]
+    ExpectedRewards { address: String },
```
JakeHartnell: What is the difference between these two queries? They seem the same to me... maybe I am missing something.

ismellike: Rewards is meant to be used after expiration, because the balance is set in config. The value refers to the actual claimable amount.

ExpectedRewards is meant to be used before the voting incentives are expired, because the vote counts and balance can still change.

JakeHartnell: Hmmmm, a little confusing. Maybe others have thoughts, but perhaps we could combine these and use a custom response type to indicate whether incentives are expired and the amount that is expected vs claimable?

Open to other ideas.

ismellike: Yeah, I think the combination would be a good idea, and I could add an `is_claimable` flag to the response.

contracts/external/dao-proposal-incentives/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name ="dao-proposal-incentives"
+authors = ["Jake Hartnell <no-reply@no-reply.com>", "Gabe Lopez <ismellike@users.noreply.github.com>"]
+description = "A contract that implements incentives for voting in a DAO."
```
JakeHartnell:
```suggestion
description = "A contract that implements incentives for creating a successful proposal in a DAO."
```

JakeHartnell:
Can we rebase? In the README for these contracts make sure we note that they aren't audited.

Issue #759, opened 2023-11-02
Auto Pass Props if User Controls 100% of Voting power


JakeHartnell:
As a user setting up DAOs before adding other members, it would be quicker to auto-pass a proposal if I control 100% of the voting power.

In the `dao-proposal-single` contract, if a user has 100% of the voting power auto-pass the proposal.

noahsaso:
closing in favor of #798 per the discussion in #776

PR #715, opened 2023-07-02
Bounties Contract

JakeHartnell:
Closes #714

PR #799, opened 2024-02-01
Cast vote on proposal creation

### LINKED ISSUE #798: Allow auto casting a vote on proposal creation
per the discussion in https://github.com/DA0-DA0/dao-contracts/pull/776, it's a better idea to allow auto-casting a vote on proposal creation than to allow auto-passing a prop based on the proposer's voting power.

auto-casting a vote will take care of auto-passing and enable more use cases. for example, the proposer might want to abstain.

noahsaso:
Resolves #798 

This allows the proposer to cast a vote when creating a proposal, which also takes care of instant-passing proposals when the member with majority voting power proposes something (#759). It will fail if the proposer is not a member (when anyone may propose is enabled).

It works by simply chaining the vote execution function after the proposal creation. This makes it future proof by respecting all other relevant features (such as `min_voting_period` and `allow_revoting`, which would have been ignored by #776).


contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -254,11 +252,35 @@ pub fn execute_propose(
...
+                },
+                Attribute {
+                    key: "rationale".to_string(),
+                    value: vote.rationale.unwrap_or_else(|| "_none".to_string()),
+                },
+            ],
+        )
+    } else {
+        (vec![], vec![])
+    };
```
JakeHartnell: Nice and simple, love it.

JakeHartnell (approved):
Wonderful! Really like this approach and that it saves a whole transaction!

contracts/proposal/dao-proposal-multiple/src/contract.rs
```diff
@@ -263,11 +251,42 @@ pub fn execute_propose(
...
+                },
+            ],
+        )
+    } else {
+        (vec![], vec![])
+    };
+
     Ok(Response::default()
         .add_submessages(hooks)
+        .add_submessages(vote_hooks)
```
bekauz, line 285: curious about one thing here: given that we basically combine the propose & vote actions into one, why not submit the vote hooks as a regular (fire and forget) message? in that case if proposal creates successfully, but the vote fails (e.g. voting on a non-existent option), the proposal would still be created fine. somewhat mimicking the current way to propose and vote.

_fwiw i also think this should be a submessage though_

contracts/proposal/dao-proposal-multiple/src/contract.rs
```diff
@@ -429,16 +448,20 @@ pub fn execute_vote(
...
         .add_submessages(change_hooks)
         .add_submessages(vote_hooks)
         .add_attribute("action", "vote")
-        .add_attribute("sender", info.sender)
+        .add_attribute("sender", sender)
         .add_attribute("proposal_id", proposal_id.to_string())
         .add_attribute("position", vote.to_string())
+        .add_attribute(
+            "rationale",
+            rationale.unwrap_or_else(|| "_none".to_string()),
```
bekauz, line 463:
```suggestion
            rationale.unwrap_or_else(|| "none".to_string()),
```

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -157,25 +152,28 @@ pub fn execute(
 pub fn execute_propose(
     deps: DepsMut,
     env: Env,
-    sender: Addr,
-    title: String,
-    description: String,
-    msgs: Vec<CosmosMsg<Empty>>,
-    proposer: Option<String>,
+    info: MessageInfo,
```
bekauz: any reason for switching from `sender: Addr` to `info: MessageInfo` here?

bekauz (approved):
lgtm! awesome pr, only have a few small questions

contracts/proposal/dao-proposal-multiple/src/contract.rs
noahsaso: i do this to match `_none` in `dao-proposal-single`. i kinda like that it distinguishes it a bit from a normal rationale. `_none` is slightly less likely to be a rationale submitted by someone, whereas `none` could be used (like `n/a` or `no rationale`)

contracts/proposal/dao-proposal-single/src/contract.rs
noahsaso: nope, i think i did that before i realized the vote sender needs to be the proposer, and never undid it. will fix! nice catch.

contracts/proposal/dao-proposal-multiple/src/contract.rs
noahsaso: hmm, i don't think it would be good practice to partially succeed a message. in my mind, if the user intends to submit a proposal and vote on it, it should only succeed if what they intended to happen actually happens. there might be a specific reason/workflow they've built that requires proposing and voting to occur atomically. if you allow votes to fail, you're essentially allowing invalid inputs to be submitted.

also, voting fails anyways in the `execute_vote` function call above this part. these submessages are just the hooks that are triggered after voting succeeds.

JakeHartnell:
CI for integrations testing is failing, but there is a fix for that here: https://github.com/DA0-DA0/dao-contracts/pull/715

noahsaso:
> CI for integrations testing is failing, but there is a fix for that here: #715

sir that is a bounties contract 😂

can you split up the CI fix into a separate pr pls <3

contracts/proposal/dao-proposal-multiple/src/contract.rs
bekauz: ah got it, cool!

JakeHartnell:
Le fix for CI: https://github.com/DA0-DA0/dao-contracts/pull/801

noahsaso:
> Le fix for CI: #801

thank you!

noahsaso:
hey @JakeHartnell seems the CI is failing for a cache issue? how do we resolve it D:

JakeHartnell:
> hey @JakeHartnell seems the CI is failing for a cache issue? how do we resolve it D:

Yolo merge? Ran test tube locally and everything passed so 99% sure it will work...

PR #800, opened 2024-02-06
Put cw-vesting staking features behind a feature flag

noahsaso:
Neutron does not support the staking module, and the `cw-vesting` contract attempts to execute a staking-related message on instantiate for the native token. This means that the vesting contract does not currently work for NTRN on Neutron.

This PR adds a `staking` feature flag to `cw-vesting` that explicitly enables the execution variants that have to do with staking, in addition to the withdraw address message on instantiate.


bekauz (approved):

JakeHartnell (approved):
LGTM!

Issue #781, opened 2023-12-11
Support `active_threshold` in `dao-voting-cw4`


JakeHartnell:
As a DAO, we would like to have SubDAOs where there needs to be a minimum number of members for the SubDAO to be active.

We support `active_threshold` for NFTs and token-based DAOs, might as well make it supported in membership / multisig based DAOs.

theghostmac:
Hi @JakeHartnell,

I have just added this in my fork, and will make a PR in a moment. However, I want to verify that there will be need to:
- Query SubDAOs based on active/inactive status.
- Enforce rules for active/inactive SubDAOs and implement checks based on the status, maybe during proposal creation or voting etc to ensure only active SubDAOs can participate.
- Adding and removing SubDAO members
- Automatically toggle a SubDAO status based on amount of members >= active threshold.

Just a checklist we may have to include to this new feature...

JakeHartnell:
To be clear, the contract that needs to be modified is `contracts/voting/dao-voting-cw4.` There are examples of implementing active threshold in `contracts/voting/dao-voting-cw721-staked` and `contracts/voting/dao-voting-token-staked`, use those for reference. Both of those examples, show the queries that need to be implemented, namely a `QueryMsg::IsActive` query.

Many of the features you listed are not actually needed for this as these features are handled via the voting module contract (i.e. `dao-voting-cw4`).

PR #808, opened 2024-03-12
[feature] Updated SubDao functionality to have members count and an active threshold

theghostmac:
This pull request addresses the issue #781 "Support active_threshold in dao-voting-cw4" by updating the SubDao functionality to include a member count and an active threshold. 

This ensures that SubDAOs can only be considered active if they meet the minimum required member count specified by the active threshold.

### DIFF
--- a/contracts/dao-dao-core/schema/dao-dao-core.json
+++ b/contracts/dao-dao-core/schema/dao-dao-core.json
@@ -1317,19 +1317,33 @@
       "SubDao": {
         "type": "object",
         "required": [
-          "addr"
+          "active_threshold",
+          "addr",
+          "member_count"
         ],
         "properties": {
+          "active_threshold": {
+            "description": "The minimum number of members required for the SubDAO to be considered active.",
+            "type": "integer",
+            "format": "uint32",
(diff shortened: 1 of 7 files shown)


theghostmac:
Can I redirect this PR to a separate branch? Seeing that `development` is the default..

JakeHartnell:
> This pull request addresses the issue https://github.com/DA0-DA0/dao-contracts/issues/781 "Support active_threshold in dao-voting-cw4" by updating the SubDao functionality to include a member count and an active threshold.

This PR is not exactly what I had in mind when I wrote the ticket which is my bad as I should have included more detail. The contract that needs to be modified is `contracts/voting/dao-voting-cw4`. There are examples of implementing active threshold in `contracts/voting/dao-voting-cw721-staked` and `contracts/voting/dao-voting-token-staked`, use those for reference.

Additionally, we try not to touch the core module unless we absolutely have to due to audit costs.

This PR just adds some metadata around this, but the core-active threshold logic is in the voting module contracts. Going to close for now, but happy to find time to chat about this issue if you want to take it on!

PR #810, opened 2024-03-18
Add admin auth to cw-admin-factory

noahsaso:
This adds the ability to set an admin on `cw-admin-factory`, which restricts the contract self-admin instantiate execution message to be executable only by the admin. If no admin is set, the contract can be executed by anyone, and there can never be an admin set again.

### DIFF
--- a/contracts/external/cw-admin-factory/src/contract.rs
+++ b/contracts/external/cw-admin-factory/src/contract.rs
@@ -46,7 +46,6 @@ pub fn execute(
             code_id,
             label,
         } => instantiate_contract(deps, env, info, msg, code_id, label),
-        ExecuteMsg::UpdateAdmin { admin } => execute_update_admin(deps, info, admin),
     }
 }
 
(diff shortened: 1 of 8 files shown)


contracts/external/cw-admin-factory/src/contract.rs
```diff
@@ -39,17 +45,26 @@ pub fn execute(
             instantiate_msg: msg,
             code_id,
             label,
-        } => instantiate_contract(env, info, msg, code_id, label),
+        } => instantiate_contract(deps, env, info, msg, code_id, label),
+        ExecuteMsg::UpdateAdmin { admin } => execute_update_admin(deps, info, admin),
```
JakeHartnell: Nit: current pattern in most of our contracts is that we use `cw-ownable` or Oak / some other auditor will have comments about not having a two-step admin transfer process.

noahsaso: yeah hm, that just felt like overkill here. i'm also open to removing the ability to update at all, and just make this an immutable contract. wdyt?

noahsaso: yeah, let's keep it stupid simple. i'm making it immutable

contracts/external/cw-admin-factory/src/contract.rs
```diff
@@ -39,17 +45,25 @@ pub fn execute(
...
+    // If admin set, require the sender to be the admin.
+    if let Some(admin) = ADMIN.load(deps.storage)? {
+        if admin != info.sender {
+            return Err(ContractError::Unauthorized {});
+        }
+    }
+
     // Instantiate the specified contract with factory as the admin.
     let instantiate = WasmMsg::Instantiate {
         admin: Some(env.contract.address.to_string()),
```
bekauz, line 69: if admin is set, could it be desirable to use that address instead of the factory as the wasm instantiate msg admin?

bekauz (approved):
overall lgtm, just one clarifying question about the `instantiate_contract` method!

contracts/external/cw-admin-factory/src/contract.rs
noahsaso: in the reply, this contract changes the admin (using the fact that it is the current admin) to the DAO's address. that's the point of the contract :)

contracts/external/cw-admin-factory/src/contract.rs
JakeHartnell: > yeah, let's keep it stupid simple. i'm making it immutable

👑

contracts/external/cw-admin-factory/src/contract.rs
bekauz: forgot to get back on this, gotcha. then all makes sense!

PR #818, opened 2024-04-03
ABC Improvements

ismellike:
#697 Extension of this PR

- [X] Use a map for HATCHER_ALLOWLIST
- [X] Clean up state when going from Hatch -> Open
- [x] Various spell-checking fixes
- [x] Allow an existing tokenfactory token - reverted
- [x] Implement a circuit breaker
- [x] Add a new method to donate into the reserve [from comment](https://github.com/DA0-DA0/dao-contracts/pull/818#discussion_r1555016419) - reverted
- [x] Add transparency around initial token config - reverted
- [x] Fix logic around auto-sending funding pool [from comment](https://github.com/DA0-DA0/dao-contracts/pull/697#discussion_r1555019234)
- [x] Move curves into cw-curves package
- [x] Allow hatcher allowlist by DAO membership


contracts/external/cw-abc/src/test_tube/integration_tests.rs
```diff
@@ -514,3 +519,151 @@ fn test_update_curve() {
...
+    let builder = TestEnvBuilder::new();
+    instantiate_msg.supply.token_info = TokenInfo::Existing {
+        denom: "junox".to_string(),
+    };
+    let result = builder.setup(&app, instantiate_msg.clone());
+    assert!(result.is_err());
+}
+
+#[test]
+fn test_existing_token() {
```
JakeHartnell: For existing tokens, I think we need more tests to ensure no invariants. For example, if I use an existing token of which I have 1 million tokens, then the curve opens up and there are only enough funds to burn 500k... what happens if I try to sell the full amount? Just get zero for the rest of it?

One of the advantages of bonding curve based tokens is that tokens can't go to zero, but using existing tokens breaks this. Maybe not the end of the world, but we should think of test cases around this if we're going to support it.

JakeHartnell: Also should be queries to know if there was a pre-mint, that's important context for users.

ismellike: My understanding of this was that everyone could still sell, but people would start buying in deeper in the curve during the hatch phase. This would lead to less reserve relative to supply in the Open phase (compared to starting from 0 supply). 

This could be fine from the user-perspective if we introduce a new function for donation into the reserve instead of the funding pool. This would increase the reserve relative to supply and - for DAO's with profit being distributed back in - could fix this "ponzi-like" situation.  

Definitely a good idea to be able to show as much context as possible though. I will try to think of some specific test-cases as well.

ismellike: After studying this some more, it's definitely not a good idea to have an existing token supply. DAO's wanting to use an ABC for liquidity should start as a membership-based DAO to instantiate the cw-abc and then update their voting module. I'm going to revert some of these changes here.

contracts/voting/dao-voting-token-staked/src/msg.rs
```diff
@@ -3,6 +3,7 @@ use cosmwasm_std::{Binary, Uint128};
 use cw_utils::Duration;
 use dao_dao_macros::{active_query, native_token_query, voting_module_query};
 use dao_interface::token::NewTokenInfo;
+#[allow(unused_imports)]
```
JakeHartnell, line 6: Why were these needed?

contracts/external/cw-abc/src/commands.rs
```diff
@@ -283,32 +273,111 @@ pub fn execute_close(deps: DepsMut, info: MessageInfo) -> Result<Response, Contr
...
+    Ok(Response::new()
+        .add_attribute("action", "donate")
+        .add_attribute("donor", info.sender)
+        .add_attribute("amount", payment)
+        .add_attribute("total_donation", total_donation)
+        .add_messages(msgs))
+}
+
+/// Withdraw funds from the funding pool (only callable by owner)
+pub fn withdraw(
```
JakeHartnell, line 258: Do we have any tests for this?

JakeHartnell (approved):
LFG!

contracts/voting/dao-voting-token-staked/src/msg.rs
ismellike: This cleans up the build output, because it doesn't recognize the Response is actually used in the QueryResponses:
image

contracts/external/cw-abc/src/commands.rs
ismellike: Adding some now

JakeHartnell:
Running `just gen` will update the contract schemas and fix those linting errors.

ismellike:
Shoutout to @BHIIKTOR btw for noticing unnecessary packages and aliases after moving things to cw-curves package

ismellike:
I still want to squeeze in a couple of things.

- [x] Buy and sell quotes (how many tokens will a user receive from x amount)
- [x] MinMax contribution limit override by hatcher config

ismellike:
this is ready 🔥

PR #819, opened 2024-04-06
Add active threshold functionality to CW4 voting module, closes #781

### LINKED ISSUE #781: Support `active_threshold` in `dao-voting-cw4`

theghostmac:
### Changelog

- Added new custom error message to error.rs, `InvalidThreshold`
- Added the `query_is_active` function and corresponding `Query::IsActive`
- Added the `execute_update_active_threshold` function and corresponding `ExecuteMsg::UpdateActiveThreshold`
- Added the new optional `Config` and updated the `instantiate` function to make use of it.

### Author
GhostMac <theghostmac>

### DIFF
--- a/contracts/voting/dao-voting-cw4/src/contract.rs
+++ b/contracts/voting/dao-voting-cw4/src/contract.rs
@@ -252,7 +252,7 @@ pub fn query_is_active(deps: Deps) -> StdResult<Binary> {
 
     let is_active = match active_threshold {
         ActiveThreshold::AbsoluteCount { count } => {
-            total_weight.weight >= count.to_string().parse::<u64>().unwrap()
+            Uint128::new(total_weight.weight as u128) >= count
         }
         _ => false, // Should never happen as percentage is not supported
     };
--- a/contracts/voting/dao-voting-cw4/src/contract.rs
+++ b/contracts/voting/dao-voting-cw4/src/contract.rs
@@ -1,16 +1,24 @@
+use std::ops::{Div, Mul};
+
+use cosmwasm_std::{
+    Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdResult, SubMsg,
+    to_json_binary, Uint128, Uint256, WasmMsg,
+};
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
-use cosmwasm_std::{
-    to_json_binary, Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdResult, SubMsg,
-    Uint128, WasmMsg,
-};
-use cw2::{get_contract_version, set_contract_version, ContractVersion};
+use cw2::{ContractVersion, get_contract_version, set_contract_version};
 use cw4::{MemberListResponse, MemberResponse, TotalWeightResponse};
 use cw_utils::parse_reply_instantiate_data;
 
+use dao_interface::voting::IsActiveResponse;
+use dao_voting::threshold::{
+    ActiveThreshold, assert_valid_percentage_threshold
+    ,
+};
+
 use crate::error::ContractError;
 use crate::msg::{ExecuteMsg, GroupContract, InstantiateMsg, MigrateMsg, QueryMsg};
-use crate::state::{DAO, GROUP_CONTRACT};
+use crate::state::{ACTIVE_THRESHOLD, Config, CONFIG, DAO, GROUP_CONTRACT};
 
 pub(crate) const CONTRACT_NAME: &str = "crates.io:dao-voting-cw4";
 pub(crate) const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -25,6 +33,20 @@ pub fn instantiate(
     msg: InstantiateMsg,
 ) -> Result<Response, ContractError> {
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let config: Config = if let Some(active_threshold) = msg.active_threshold {
+        Config {
+            active_threshold_enabled: true,
+            active_threshold: Some(active_threshold),
+        }
+    } else {
+        Config {
+            active_threshold_enabled: false,
+            active_threshold: None,
+        }
+    };
+
+    CONFIG.save(deps.storage, &config)?;
 
     DAO.save(deps.storage, &info.sender)?;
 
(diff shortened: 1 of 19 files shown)


JakeHartnell:
Needs a few things:
- Fix the build
- Add tests for `active_threshold` and being able to update it
- Make sure CI passes
- Clean up

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -26,6 +34,20 @@ pub fn instantiate(
...
+        Config {
+            active_threshold_enabled: true,
+            active_threshold: Some(active_threshold),
+        }
+    } else {
+        Config {
+            active_threshold_enabled: false,
+            active_threshold: None,
+        }
+    };
```
JakeHartnell: Would do this a bit differently, see how it's done in `dao-voting-token-staked`: https://github.com/DA0-DA0/dao-contracts/blob/3ab701747d6d5571cfee7f5a78d8fe39ffa3a8f7/contracts/voting/dao-voting-token-staked/src/contract.rs#L79

contracts/voting/dao-voting-cw4/src/state.rs
```diff
@@ -1,5 +1,19 @@
...
+/// The minimum amount of users for the DAO to be active
+pub const ACTIVE_THRESHOLD: Item<ActiveThreshold> = Item::new("active_threshold");
+
+#[cw_serde]
+pub struct Config {
+    pub active_threshold_enabled: bool,
+    pub active_threshold: Option<ActiveThreshold>,
+}
+
+pub const CONFIG: Item<Config> = Item::new("config");
```
JakeHartnell: We don't really need this as if `ACTIVE_THRESHOLD` is not set, it's not enabled. There is no other config needed for this contract.

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -164,12 +245,48 @@ pub fn query_total_power_at_height(deps: Deps, env: Env, height: Option<u64>) ->
...
+            let group_contract = GROUP_CONTRACT.load(deps.storage)?;
+            let total_weight: TotalWeightResponse = deps.querier.query_wasm_smart(
+                &group_contract,
+                &cw4_group::msg::QueryMsg::TotalWeight { at_height: None },
+            )?;
+            to_json_binary(&IsActiveResponse {
+                active: total_weight.weight >= count.into(),
+            })
+        }
+        Some(ActiveThreshold::Percentage { percent, .. }) => {
```
JakeHartnell: Should have added a note on this, but `ActiveThreshold::Percentage` is not really needed in the case of a membership DAO. Unlike NFTs or Tokens, there is no notion of a `total_supply` or staking.

This is good because we can simplify the contract!

On instantiate, we should only allow for `ActiveThreshold` to be defined in terms of absolute count, and throw an error if someone submits a percentage.

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -108,12 +130,70 @@ pub fn instantiate(
...
+    if info.sender != dao {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    if let Some(active_threshold) = &new_active_threshold {
+        // Pre-validation before state changes
+        match active_threshold {
+            ActiveThreshold::Percentage { percent } => {
+                assert_valid_percentage_threshold(*percent)?;
+            }
```
JakeHartnell: I think we should not allow `ActiveThreshold::Percentage` for this contract, so we should throw an error here.

contracts/voting/dao-voting-cw4/src/contract.rs
JakeHartnell: We don't need the `Config` here, just saving the `ACTIVE_THRESHOLD` state if `active_threshold` is provided.

Additionally, we should validate the `active_threshold`, it should:
- Be an `ActiveThreshold::AbsoluteCount`
- Greater than zero

JakeHartnell: Don't forget to save `ACTIVE_THRESHOLD` here.

theghostmac:
> Needs a few things:
> 
> * Fix the build
> * Add tests for `active_threshold` and being able to update it
> * Make sure CI passes
> * Clean up

I have updated the contract with everything you pointed out. Build passes. Moving on to the tests.

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -26,6 +29,18 @@ pub fn instantiate(
...
+        Config {
+            active_threshold: Some(active_threshold),
+        }
+    } else {
+        Config {
+            active_threshold: None,
+        }
+    };
+
+    CONFIG.save(deps.storage, &config)?;
```
JakeHartnell: This config struct is really not necessary. You forget to save `ACTIVE_THRESHOLD` here.

contracts/voting/dao-voting-cw4/src/state.rs
```diff
@@ -1,5 +1,17 @@
...
+
+/// The minimum amount of users for the DAO to be active
+pub const ACTIVE_THRESHOLD: Item<ActiveThreshold> = Item::new("active_threshold");
+
+#[cw_serde]
+pub struct Config {
+    pub active_threshold: Option<ActiveThreshold>,
+}
+
+pub const CONFIG: Item<Config> = Item::new("config");
```
JakeHartnell: This is the same info as `ACTIVE_THRESHOLD`. Delete Config struct.

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -164,12 +217,48 @@ pub fn query_total_power_at_height(deps: Deps, env: Env, height: Option<u64>) ->
...
+        None => to_json_binary(&IsActiveResponse { active: true }),
+    }
+}
+
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
+    let config: Config = CONFIG.load(deps.storage)?;
+    // Update config as necessary
+    CONFIG.save(deps.storage, &config)?;
+
```
JakeHartnell: Not needed.

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -164,12 +217,48 @@ pub fn query_total_power_at_height(deps: Deps, env: Env, height: Option<u64>) ->
...
+                &cw4_group::msg::QueryMsg::TotalWeight { at_height: None },
+            )?;
+            let percentage_base = Uint256::from(10u64).pow(percent.decimal_places()); // Ensure correct power base for scaling
+            let required_weight = Uint256::from(total_weight.weight)
+                .multiply_ratio(percent.atomics(), percentage_base); // Correct ratio calculation
+
+            to_json_binary(&IsActiveResponse {
+                active: Uint256::from(total_weight.weight) >= required_weight,
+            })
+        }
```
JakeHartnell: `ActiveThreshold::Percent` should NOT be used with this contract. Do you see that you're just comparing total_weight.weight` to `total_weight.weight`?

JakeHartnell: Please read my earlier comments about not allowing `ActiveThreshold::Percent`.

contracts/voting/dao-voting-cw4/src/contract.rs
JakeHartnell: Don't forget to validate that it's not a percent.

JakeHartnell:
> > Needs a few things:
> > 
> > * Fix the build
> > * Add tests for `active_threshold` and being able to update it
> > * Make sure CI passes
> > * Clean up
> 
> I have updated the contract with everything you pointed out. Build passes. Moving on to the tests.

Close, but see above comments.
- Get rid of `CONFIG`, it's duplicate state for no reason
- Make sure to save `ACTIVE_THRESHOLD` on instantiation and validate that it's NOT `ACTIVE_THRESHOLD::Percent`
- Get rid of `ACTIVE_THRESHOLD::Percent` in the `IsActive` query, it does nothing right now and we should not allow it
- Fix linting errors and tests, CI should pass

contracts/voting/dao-voting-cw4/src/contract.rs
```diff
@@ -164,10 +238,28 @@ pub fn query_total_power_at_height(deps: Deps, env: Env, height: Option<u64>) ->
...
+    let active_threshold = ACTIVE_THRESHOLD.load(deps.storage)?;
+    let group_contract = GROUP_CONTRACT.load(deps.storage)?;
+    let total_weight: TotalWeightResponse = deps.querier.query_wasm_smart(
+        group_contract,
+        &cw4_group::msg::QueryMsg::TotalWeight { at_height: None },
+    )?;
+
+    let is_active = match active_threshold {
+        ActiveThreshold::AbsoluteCount { count } => {
+            total_weight.weight >= count.to_string().parse::<u64>().unwrap()
```
JakeHartnell: So the active threshold `count` here is `Uint128`. Rather than doing this weird `count.to_string().parse::<u64>().unwrap()` (converting to string and then parsing that into `u64`). It's cleaner to just convert the `total_weight.weight` `u64` into `Uint128`. Don't loose any fidelity that way either (though unlike will be using very large numbers in this context).

Something like:
```suggestion
            Uint128::new(total_weight.weight as u128) >= count
```

As a bonus, this gets rid of an `unwrap()` which we generally don't like to use in smart contracts.

PR #824, opened 2024-04-27
bump `cosmwasm-std` to `1.5.4`

bekauz:
following the latest [cw security advisory](https://github.com/CosmWasm/advisories/blob/main/CWAs/CWA-2024-002.md), we bump the version of `cosmwasm-std` from `1.5.0` to `1.5.4`.

given that `overflow-checks = true` flag is enabled, we are only affected by:
> - `Uint{64,128}::pow` / `Int{64,128}::pow`
> - `Int{64,128}::neg` (the unary negation operator -). The only value that can overflow is `Int{64,128}::MIN`.

affected contracts (_potentially_):
- `cw20-stake-external-rewards`
- `dao-voting-cw20-staked`
- `dao-voting-cw721-staked`
- `dao-voting-token-staked`
- `dao-voting`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,9 +257,9 @@ dependencies = [
 
 [[package]]
 name = "bnum"
-version = "0.8.1"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


JakeHartnell (approved):
LGTM

PR #827, opened 2024-05-08
Update optimizer cmd to v0.15.1

ismellike:
Updates optimizer and renames just workspace-optimize to just optimize for simplicity


JakeHartnell (approved):
LGTM

PR #829, opened 2024-05-08
Fix unwrap_or_default attr values

### LINKED ISSUE #828: Response attributes with null/empty values
If a response's attribute value is set to the default string (null/empty), the WasmMsg will error out with an error like such:

```
Error {
    context: "Error executing WasmMsg:\n  sender: dao\n  Execute { contract_addr: \"contract0\", msg: {\"extension\":{\"msg\":{\"update_token_role\":{\"token_id\":\"1\",\"role\":null}}}}, funds: [] }",
    source: EmptyAttributeValue {
        key: "role",
    },
}
```

Example from cw721-roles with test:
https://github.com/DA0-DA0/dao-contracts/blob/3ab701747d6d5571cfee7f5a78d8fe39ffa3a8f7/contracts/external/cw721-roles/src/contract.rs#L329

```
app.execute_contract(
        Addr::unchecked(DAO),
        cw721_addr.clone(),
        &ExecuteMsg::Extension {
            msg: ExecuteExt::UpdateTokenRole {
                token_id: "1".to_string(),
                role: None,
            },
        },
        &[],
    )
    .unwrap_err();
```

Some cases like the proposal execute error logging cannot be defaulted, because it is set up to always contain an err. (should fix anyway though)
https://github.com/DA0-DA0/dao-contracts/blob/3ab701747d6d5571cfee7f5a78d8fe39ffa3a8f7/contracts/proposal/dao-proposal-single/src/contract.rs#L1084

ismellike:
Closes #828


JakeHartnell:
Ah, really good catch. Just some linting errors, then LGTM.

JakeHartnell (approved):
Nice catch!

noahsaso (approved):
thanks 🙏

PR #831, opened 2024-05-16
Integrate cw-orch

Kayanski:
This PR aims at adding cw-orch inside Dao-Dao. 
Tasks to fill : 
- [x] Add a cw-orch-interface package with all the contract interfaces.
- [x] Add the derive macros on all execute and query msgs
- [x] Write tests using cw-orch syntax
- [ ] Osmosis Test Tube tests
- [ ] Include artifacts in there (at least in crates.io release).


ismellike:
Minor request 
It would be great to use the `id = <default contract id>` syntax for the contract interfaces. 
https://orchestrator.abstract.money/contracts/interfaces.html#creating-an-interface

Kayanski:
> Minor request It would be great to use the `id = <default contract id>` syntax for the contract interfaces. https://orchestrator.abstract.money/contracts/interfaces.html#creating-an-interface

yes god idea ! 
However, when doing so, you won't be able to use the 2 same contracts inside a project. You wouldn't be able to test or script simply with 2 daos for instance, is that acceptable ?

ismellike:
> > Minor request It would be great to use the `id = <default contract id>` syntax for the contract interfaces. https://orchestrator.abstract.money/contracts/interfaces.html#creating-an-interface
> 
> yes god idea ! However, when doing so, you won't be able to use the 2 same contracts inside a project. You wouldn't be able to test or script simply with 2 daos for instance, is that acceptable ?

Ah ok! I wasn't aware of that, so it's probably not great from an external perspective. Thanks, happy with this as is.

JakeHartnell (approved):
LGTM. I love `cw-orchestrator`. It's so nice.

This is a good start. Refactoring the rest of the tests could be done piecemeal. I do think tests are our biggest piece of tech debt as when we started we didn't have as good of tools or patterns as we now have today.

JakeHartnell:
Closing in favor of #876

PR #835, opened 2024-06-01
staking rewards distribution followup

bekauz:
reworks the existing staking rewards distribution logic.

contract now functions as a factory for distributing staking rewards.
there is no more main voting power contract; each denom to be distributed needs to specify where to draw its voting power from.

reward rate is derived from the `RewardEmissionConfig` which describes how many udenoms (`Uint128`) are to be distributed per time duration (blocks/seconds).

introduces a shutdown option which claws back any undistributed rewards back to the treasury.

also implemented a suite test builder because the current unit test file was about to exceed 3000 lines of code (mostly because of copy pasting the setup in every test). however not all tests are present - this is a pending todo.


contracts/distribution/dao-rewards-distributor/src/msg.rs
```diff
@@ -0,0 +1,146 @@
...
+    Fund {},
+    /// Updates the reward emission config which controls the rate at which rewards are issued.
+    UpdateRewardEmissionConfig {
+        denom: String,
+        new_emission_rate: Uint128,
+        new_emission_time: Duration,
+    },
+    /// shuts down the rewards distributor. withdraws all future staking rewards
+    /// back to the treasury. members can claim whatever they earned until this point.
+    Shutdown { denom: String },
```
noahsaso, line 44: it may be worth splitting shutdown and withdraw into like `stop` and `withdraw` 🤔

and i also wonder if withdraw should have a destination address and optionally let you configure specifically what to withdraw

it's nonobvious to me that every time a DAO wants to pause rewards, they'll also want to withdraw funds back to the treasury. they may have some rewards funnel that sends funds to the distributor, and if they withdraw funds back to the treasury, it might get merged with all their other tokens and be hard to keep track of. also, the rewards manager and treasuries may be in different DAOs/SubDAOs, so just treating the owner as the treasury doesn't totally make sense to me either.

contracts/distribution/dao-rewards-distributor/src/msg.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// registers a new reward denom
+    RegisterRewardDenom(RewardDenomRegistrationMsg),
+}
+
+#[cw_serde]
+pub struct RewardDenomRegistrationMsg {
+    pub denom: UncheckedDenom,
+    pub reward_emission_config: RewardEmissionConfig,
+    pub vp_contract: String,
+    pub hook_caller: String,
```
noahsaso: maybe we make this optional and autofill with `vp_contract` if empty?

contracts/distribution/dao-rewards-distributor/src/msg.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Claims rewards for the sender.
+    Claim { denom: String },
+    /// Used to fund this contract with cw20 tokens.
+    Receive(Cw20ReceiveMsg),
+    /// Used to fund this contract with native tokens.
+    Fund {},
+    /// Updates the reward emission config which controls the rate at which rewards are issued.
+    UpdateRewardEmissionConfig {
+        denom: String,
+        new_emission_rate: Uint128,
```
noahsaso: `rate` means unit per time, so i think we should rename this. maybe `RewardEmissionConfig` becomes `RewardEmissionRate` with `amount` and `time` fields. then everywhere becomes `emission_rate` and `new_emission_rate` pointing to that struct

contracts/distribution/dao-rewards-distributor/src/msg.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Used to fund this contract with cw20 tokens.
+    Fund {},
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    /// Returns configuration information about this contract.
+    #[returns(InfoResponse)]
+    Info {},
```
noahsaso, line 118: `info` and `InfoResponse` in all our other contracts tends to be reserved for contract_version info (which we should add here too). i think we should change this to `config` and `ConfigResponse`, which makes more sense anyway.

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+
+/// registers a new denom for rewards distribution.
+/// only the owner can register a new denom.
+/// a denom can only be registered once; update if you need to change something.
+fn execute_register_reward_denom(
+    deps: DepsMut,
+    info: MessageInfo,
+    msg: RewardDenomRegistrationMsg,
+) -> Result<Response, ContractError> {
+    // only the owner can initiate a shutdown
```
noahsaso:
```suggestion
    // only the owner can register
```

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+        .checked_sub(Decimal::from_ratio(
+            passed_units_since_start,
+            reward_duration_units,
+        ))
+        .map_err(|e| ContractError::Std(StdError::overflow(e)))?;
+
+    // to get the clawback msg
+    let clawback_msg = get_transfer_msg(
+        info.sender.clone(),
+        reward_config.funded_amount * remaining_reward_duration_fraction,
```
noahsaso: does this need a checked_mul or anything?

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+
+fn execute_fund(
+    mut deps: DepsMut,
+    env: Env,
+    sender: Addr,
+    mut denom_reward_config: DenomRewardConfig,
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    // Ensure that the sender is the owner
+    cw_ownable::assert_owner(deps.storage, &sender)?;
```
noahsaso: i think anyone should be able to fund rewards. there may be many sources funneling rewards into one place. is this auth necessary for some reason? it makes sense to me that only the owner can register new denoms (so people can't spam some fake token and brick the contract), but funding i think can be permissionless

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+        Expiration::AtTime(t) => {
+            if t.seconds() <= env.block.time.seconds() {
+                // expiration is the funded duration after current block time
+                Expiration::AtTime(t.plus_seconds(funded_period_units))
+            } else {
+                // if the previous expiration had not yet expired, we extend
+                // the current rewards period by the newly funded duration
+                Expiration::AtTime(env.block.time.plus_seconds(funded_period_units))
+            }
+        }
```
noahsaso:
```suggestion
        Expiration::AtTime(t) => {
            if t.seconds() <= env.block.time.seconds() {
                // expiration is the funded duration after current block time
                Expiration::AtTime(env.block.time.plus_seconds(funded_period_units))
            } else {
                // if the previous expiration had not yet expired, we extend
                // the current rewards period by the newly funded duration
                Expiration::AtTime(t.plus_seconds(funded_period_units))
            }
        }
```

i think you flipped the cases. the first one should use the current block time, and the second one should extend the current period

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+    };
+
+    denom_reward_config.funded_amount += amount;
```
noahsaso: does this need to decrease at all by what's already been distributed? or is it just a total? and to whichever one it is, can you explain to me why 🙏

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+fn execute_claim(
+    mut deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    denom: String,
+) -> Result<Response, ContractError> {
+    // Update the rewards information for the sender.
+    update_rewards(&mut deps, &env, &info.sender, vec![denom.to_string()])?;
+
+    // Get the checked denom for the string based denom
```
noahsaso: cap

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+                Some(rc) => Ok(rc.bump_last_update(&env.block)),
+                None => Err(StdError::generic_err("Denom config not found")),
+            }
+        })?;
+    }
+
+    Ok(())
+}
+
+fn get_expiration_diff(a: Expiration, b: Expiration) -> StdResult<u64> {
```
noahsaso: can you add some comments or rename the variables (or both) to make it more clear why it returns 0 if `a < b`?

contracts/distribution/dao-rewards-distributor/src/state.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// - If the period finish expiration is `AtHeight(h)` or `AtTime(t)`,
+    /// we compare the current block height or time with `h` or `t` respectively.
+    /// If current block respective value is lesser than that of the
+    /// `period_finish_expiration`, means rewards are still being distributed.
+    /// We therefore return the current block `height` or `time`, as that was the
+    /// last date where rewards were distributed.
+    /// If current block respective value is greater than that of the
+    /// `period_finish_expiration`, means rewards are no longer being distributed.
+    /// We therefore return the `period_finish_expiration` value, as that was the
+    /// last date where rewards were distributed.
```
noahsaso:
```suggestion
    /// Returns the latest date where rewards were still being distributed.
    /// Works by comparing `current_block` with the distribution expiration:
    /// - If the period finish expiration is `Never`, then no rewards are being
    /// distributed, thus we return `Never`.
    /// - If the period finish expiration is `AtHeight(h)` or `AtTime(t)`,
    /// we compare the current block height or time with `h` or `t` respectively.
    /// If current block respective value is lesser than that of the
    /// `distribution_expiration`, means rewards are still being distributed.
    /// We therefore return the current block `height` or `time`, as that was the
    /// last date where rewards were distributed.
    /// If current block respective value is greater than that of the
    /// `distribution_expiration`, means rewards are no longer being distributed.
    /// We therefore return the `distribution_expiration` value, as that was the
    /// last date where rewards were distributed.
```

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+        }
+        (Expiration::Never {}, Expiration::Never {}) => Ok(0),
+        _ => Err(StdError::generic_err(format!(
+            "incompatible expirations: got a {:?}, b {:?}",
+            a, b
+        ))),
+    }
+}
+
+fn get_rewards_per_token(
```
noahsaso: can you add some explanations about what this does and how the math works? i kind of get the vibe of it but i'm having a little trouble fully comprehending the math because of the semi-vague variable names and lack of comments 😁

the one comment you did include is the simplest and most clearly-labeled line lollll

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+    mut denom_reward_config: DenomRewardConfig,
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    // Ensure that the sender is the owner
+    cw_ownable::assert_owner(deps.storage, &sender)?;
+
+    let denom_str = denom_reward_config.to_str_denom();
+
+    // first we update the existing rewards (if any)
+    update_rewards(&mut deps, &env, &sender, vec![denom_str.to_string()])?;
```
noahsaso: why update rewards for the owner that's funding the contract? isn't that the DAO itself?

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+            Ok(user_reward_config)
+        })?;
+
+        // Update the last update expiration in the DenomRewardConfig
+        REWARD_DENOM_CONFIGS.update(deps.storage, denom.clone(), |config| -> StdResult<_> {
+            match config {
+                Some(rc) => Ok(rc.bump_last_update(&env.block)),
+                None => Err(StdError::generic_err("Denom config not found")),
+            }
+        })?;
```
noahsaso: why bump the last update for an entire denom when only one user's rewards are being updated? if one user claims rewards, but another doesn't, then doesn't that last update only apply to the user that claimed?

or is this just tracking CUMULATIVE_REWARDS_PER_TOKEN updates? if so, maybe move this piece up to right below that update so it's clear that those are related, and explain why in the comment

contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -0,0 +1,683 @@
...
+        let reward_factor = reward_per_token.checked_sub(user_last_reward_per_token)?;
+
+        // Calculate the amount of rewards earned.
+        // voting_power * reward_factor / scale_factor
+        let earned_rewards_amount: Uint128 = voting_power
+            .checked_mul(reward_factor)?
+            .checked_div(scale_factor())?
+            .try_into()?;
+
+        entitled_rewards.insert(denom.to_string(), earned_rewards_amount);
```
noahsaso: i must be misunderstanding how this `get_accrued_rewards_since_last_user_action` works, but it seems like this is using the current voting power to determine all of the previously unclaimed rewards.

what i'm imagining:
- user has 10 voting power
- reward distribution starts
- 1 week goes by
- user stakes to get 20 voting power
- hook is triggered to update rewards for the user
- all of the unclaimed tokens (`reward_factor`) from that period get multiplied by the new voting power (20) and saved as pending

what am i missing?

noahsaso (commented):
ok wow. nice work! this is a first pass and i don't quite understand everything yet, but i'm getting there. take my questions to represent the lack of my understanding, not that i think you did something wrong.

i think there are a handful of places that could be commented/explained beyond the surface level instructions that are written out (and some variable names improved), but maybe i can take that upon myself to confirm understanding and clean things up once all of the feature changes are done.

memorializing things we discussed in the discord sync here so we don't forget to do them:
- split shutdown and withdraw into separate actions
- require specifying recipient when withdrawing, since owner may not be the desired recipient
- make funding permissionless but keep denom registration permissioned
- make it possible to update reward rates on-the-fly without affecting past unclaimed rewards (basically a stop and restart)

great fkn work tho, damn

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: merged the CUMULATIVE_REWARDS_PER_TOKEN value into the reward denom map since they were always tied together. it also makes it more clear that the `last_update` field in the denom reward struct is talking about the `total_earned_puvp` value

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: i added some comments and renamed stuff to make it more readable, please check that they're right!

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: i made the fix but pls verify that i'm right

contracts/distribution/dao-rewards-distributor/src/msg.rs
noahsaso: nah it's fine, explicit is good

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: i also noticed that the period start date is updated when new funding comes in, but `funded_amount` only ever increases. is this right? it feels like it's losing some information about the original timeframe and i'm worried some relationship between how much is funded and when the start and end are is breaking somewhere in here.

contracts/distribution/dao-rewards-distributor/src/msg.rs
bekauz: splitting shutdown and making destination address configurable makes sense to me!

contracts/distribution/dao-rewards-distributor/src/contract.rs
bekauz: great catch! i believe that's a leftover from previous design and we can safely remove it

noahsaso: ok for sure, but rewards should still be updated when funding, right? or no?

bekauz: as it stands right now, the only thing that really changes with funding (other than the contract balance) is the rewards end date. On fund, we look at the current emission rate configuration and extend the funding end date in one of these two ways:
1. if funding block height/time is before the current rewards duration end date, we extend the current rewards duration end date by the funded period duration. basically meaning if we fund before we run out of funds, we just move the deadline further away.
2. if funding block height/time is after the current rewards end date, or if there was no end date (first time funding), we apply the funded period duration on top of the current block height/time. in other words, if the existing deadline is in the past, we just start counting from the current block.

contracts/distribution/dao-rewards-distributor/src/contract.rs
bekauz: great question, i believe the way it works is:
1. user has 10 voting power and we start the rewards distribution
2. 1 week goes by, user accrued X rewards, but that is not logged anywhere yet
3. user doubles their stake to 20 voting power, `execute_stake` hook triggers `update_rewards`
4. `update_rewards` then:
    - gets `get_total_earned_puvp`. because it's the first time it's being called,
    it calculates the total rewards earned per unit voting power for duration from the last update
    until this block, which is that of 1 week.
    - stores that `total_earned_puvp` in the denom state. this is a global config that affects other users:
        - given that user now doubled their stake, other users will get relatively less rewards
        - updating the global config as a result of this action ensures that the next state change hook
        (from any user) will have the latest "snapshot" of the stakers' voting powers.
        i'll try to go deep on this point in the docs.
    - finds the user's accrued rewards since the last user action.
    because it's the first time that user is staking since rewards started,
    the user's denom_rewards_puvp for this denom is 0. users total rewards are
    therefore X - 0, because 0 is the last known users denom_rewards_puvp.
5. we update the user's reward state:
  - previous pending rewards: 0
  - earned rewards: X
  - new pending rewards: 0 + X = X

I have added a [unit test](https://github.com/DA0-DA0/dao-contracts/blob/0e28d29faa233d78239fb28e211d6bab4c63bcb7/contracts/distribution/dao-rewards-distributor/src/testing/tests.rs#L969) that goes through this in practice:

```rust

#[test]
fn test_native_dao_rewards_entry_edge_case() {
    let mut suite = SuiteBuilder::base(super::suite::DaoType::Native).build();

    // we start with the following staking power split:
    // [ADDR1: 100, ADDR2: 50, ADDR3: 50], or [ADDR1: 50%, ADDR2: 25%, ADDR3: 25%
    suite.assert_amount(1_000);
    suite.assert_ends_at(Expiration::AtHeight(1_000_000));
    suite.assert_duration(10);

    // skip 1/10th of the time
    suite.skip_blocks(100_000);

    // ADDR1 stakes additional 100 tokens, bringing the new staking power split to
    // [ADDR1: 200, ADDR2: 50, ADDR3: 50], or [ADDR1: 66.6%, ADDR2: 16.6%, ADDR3: 16.6%]
    // this means that per 100_000 blocks, ADDR1 should receive 6_666_666, while
    // ADDR2 and ADDR3 should receive 1_666_666 each.
    suite.mint_native_coin(coin(100, DENOM), ADDR1);
    suite.stake_native_tokens(ADDR1, 100);

    // rewards here should not be affected by the new stake,
    suite.assert_pending_rewards(ADDR1, DENOM, 5_000_000);
    suite.assert_pending_rewards(ADDR2, DENOM, 2_500_000);
    suite.assert_pending_rewards(ADDR3, DENOM, 2_500_000);

    // skip 1/10th of the time
    suite.skip_blocks(100_000);

    // here we should see the new stake affecting the rewards split.
    suite.assert_pending_rewards(ADDR1, DENOM, 5_000_000 + 6_666_666);
    suite.assert_pending_rewards(ADDR2, DENOM, 2_500_000 + 1_666_666);
    suite.assert_pending_rewards(ADDR3, DENOM, 2_500_000 + 1_666_666);
  
   <...>
}
```

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: ah right, ok great!

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: ok yes, sick.

memorializing what we talked about in discord: seems like this is working because `SnapshotMap`, which we use in staking modules to store historical voting power values, returns the *previous* block's value when queried at a given height, meaning that updates in hooks still access the previous voting power and total power values, even though they were just updated.

this is what is making this logic work, since really the rewards update needs to use the last voting power to update its pending rewards just before the new voting power takes effect.

however, we shouldn't rely on the staking contract (caller) to always give us the voting power/total power before the change in a hook, and instead we should be querying total power from the *previous* block, not the current one.

bekauz:
removing the reward emission config updates for now as this pr is big enough. added next steps in #839 .

noahsaso (approved):
hellllllll yeah. amazing work guys.

for the next PR, we want to refactor to support pausing/unpausing, separate shutdown with recipient message, and update-able emission configs that don't affect the unclaimed rewards

PR #837, opened 2024-06-05
moved .cargo/config to .cargo/config.toml

noahsaso:
fix deprecated cargo config files by renaming to `config.toml`, and fixed some linter warnings


JakeHartnell (approved):
LGTM

JakeHartnell:
Is there a way to use one of these across the entire workspace?

noahsaso:
> Is there a way to use one of these across the entire workspace?

great question! yes in fact: https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

> Cargo allows local configuration for a particular package as well as global configuration. It looks for configuration files in the current directory and all parent directories.

i'll delete all the ones not in the root :)

PR #840, opened 2024-06-17
upgrade CW packages to v2

noahsaso:
migration guide: https://github.com/CosmWasm/cosmwasm/blob/main/MIGRATING.md#15x---20x

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,12 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.75"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
+checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
+dependencies = [
+ "backtrace",
+]
 
 [[package]]
 name = "assert_matches"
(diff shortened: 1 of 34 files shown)


noahsaso:
blocked on https://github.com/CosmWasm/cw-nfts/pull/160

mintthemoon:
I'm looking into this, there's a bit of a dependency snag to resolve here -- cw-nfts actually depends on this repo, for the `cw-paginate-storage` package: https://github.com/public-awesome/cw-nfts/blob/v0.19.0/Cargo.toml#L28

I can't get the `cw721` package to compile without updating or removing this dependency. There's only one place the package is used and it only relies on the `paginate_map_values` query defined [here](https://github.com/DA0-DA0/dao-contracts/blob/development/packages/cw-paginate-storage/src/lib.rs#L99-L113).

Since it's only called once and the implementation is hardly any longer than the function call, maybe the best solution is to inline it and stop relying on the package?

noahsaso:
> I'm looking into this, there's a bit of a dependency snag to resolve here -- cw-nfts actually depends on this repo, for the `cw-paginate-storage` package: https://github.com/public-awesome/cw-nfts/blob/v0.19.0/Cargo.toml#L28
> 
> I can't get the `cw721` package to compile without updating or removing this dependency. There's only one place the package is used and it only relies on the `paginate_map_values` query defined [here](https://github.com/DA0-DA0/dao-contracts/blob/development/packages/cw-paginate-storage/src/lib.rs#L99-L113).
> 
> Since it's only called once and the implementation is hardly any longer than the function call, maybe the best solution is to inline it and stop relying on the package?

omg nooo, thank you for figuring this all out 🙏 

i'm very down for that solution. it sounds best to me to depend on fewer packages if the reliance is so trivial.

mintthemoon:
Actually I see they're not even using it for pagination now that I look closer at the implementation, passing `None` for both bounds. So I'd say that's definitely the way to go, code is much more clear without the indirection anyways.

noahsaso:
public-awesome/cw-nfts#160 was merged so this should be unblocked!

PR #843, opened 2024-06-25
Improve pre-propose proposal creation permission granularity (and bump versions to v2.5.0)

### LINKED ISSUE #841: Allowlist proposal creators
Right now, the pre-propose modules allow the DAO to decide between anyone can propose or only members can propose. We should add a third option that lets the DAO specify exactly who can make proposals, which enables many other use cases.

Maybe a DAO wants community members to vote but not create proposals, and a parent DAO or operations SubDAO is responsible for making proposals.

Or maybe a contract in a protocol makes proposals for the team to allow or reject automated actions.

noahsaso:
Resolves #841 

This makes it possible to specify exactly who can create proposals, by allowing the pre-propose modules to configure:
- whether or not DAO members can submit proposals
- an allowlist
- a denylist that overrides the other two settings

It also bumps the contract versions to v2.5.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,12 +271,12 @@ dependencies = [
  "cw-admin-factory",
  "cw-utils 1.0.3",
  "cw20 1.1.2",
- "cw20-stake 2.4.2",
+ "cw20-stake 2.5.0",
  "dao-dao-core",
  "dao-interface",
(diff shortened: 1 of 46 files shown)


JakeHartnell (approved):
Nice. : )

PR #844, opened 2024-07-04
Gauge Contracts

JakeHartnell:
Inspired by the [Curve gauge system](https://resources.curve.fi/reward-gauges/understanding-gauges/) and forked from the [Wynd DAO repo](https://github.com/wynddao/wynddao) with history and Apache 2.0 license maintained.

These contracts have been modified to be more generic and support any type of DAO.

However, consider this a straw-man implementation. There are many improvements to be made!
- [ ] More tests (consider refactoring with cw-orch)
- [ ] Make improvements to `GaugeAdapter` API
- [ ] Improve `GaugeAdapter` example (needs significant work, we may want more than one example)
- [ ] Currently oes not handle small voting power numbers elegantly (a problem with NFT DAOs, where people might stake 1 NFT)
- [ ] Better READMEs and documentation

Bonus:
- [ ] Implement decay, so that people eventually have to re-express their preferences on the gauge
- [ ] Think about how to hook in incentives from the `dao-rewards-distributor` contract so that DAOs can optional incentivize people to participate in gauges

If you are looking to contribute to this and want to understand how they work, the `gauge-orchestrator` [README](https://github.com/DA0-DA0/dao-contracts/tree/gauges-are-cool/contracts/gauges/gauge) is a good place to start.

[Twitter post and discussion.](https://x.com/JakeHartnell/status/1809247777567748187)

Gauges in actual decentralize applications:
- [Curve](https://dao.curve.fi/gaugeweight): the OG
- [Wynd](https://app.wynddao.com/gauges): what these are based on

Use cases:
- Budget Allocation
- Setting parameters
- Incentives Allocation
- Managing liquid staking protocols (choosing which validators to to delegate to)
- etc.


hard-nett:
> * [ ]  More tests (consider refactoring with cw-orch)
> * [ ]  Make improvements to `GaugeAdapter` API
> * [ ]  Improve `GaugeAdapter` example (needs significant work, we may want more than one example)

spearheading these three points this week! an additional gauge adapter template can be one that turns itself off after a specific # of times the gauges epoch occurs.

JakeHartnell:
> > * [ ]   More tests (consider refactoring with cw-orch)
> > * [ ]   Make improvements to `GaugeAdapter` API
> > * [ ]   Improve `GaugeAdapter` example (needs significant work, we may want more than one example)
> 
> spearheading these three points this week! an additional gauge adapter template can be one that turns itself off after a specific # of times the gauges epoch occurs.

AMAZING :heart:

contracts/gauges/gauge/src/msg.rs
```diff
@@ -0,0 +1,263 @@
...
+
+use crate::state::{Reset, Vote};
+
+type GaugeId = u64;
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    /// Address of contract to that contains all voting powers (where we query)
+    pub voting_powers: String,
+    /// Addres that will call voting power change hooks (often same as voting power contract)
```
JakeHartnell, line 14:
```suggestion
    /// Address that will call voting power change hooks (often same as voting power contract)
```

PR #847, opened 2024-07-13
Use cw-denom and payment utils in gauge adapter

ismellike:
Work on #844 

Improves deposit and reward validation by using cw-denom and payment utils.

Refactors tests w/ cw-orch @hard-nett


JakeHartnell (approved):
AMAZING ❤️

PR #848, opened 2024-07-15
dao rewards distribution emission rate updates

### LINKED ISSUE #839: enable dao-rewards-distributor to keep historic emission update logs
followup to #835 

currently the dao-rewards-distributor contract only tracks the latest emission schedule. this is limiting in cases where DAOs may want to update their emission schedules.

to implement that we should keep track of historic emission rate configs, along with their start dates. this way we can ensure that users can claim their earned rewards forever, and no emission rate updates will take that away from them.

bekauz:
closes #839 

contract now keeps all historic emission rates in storage which allows us to update the rewards and keep the pending rewards correct.


contracts/distribution/dao-rewards-distributor/src/state.rs
```diff
@@ -54,41 +78,119 @@ pub struct DenomRewardState {
...
+        // 2. push current epoch to historic configs
+        self.historic_epoch_configs.push(curr_epoch.clone());
+
+        // 3. deduct the distributed rewards amount from total funded amount,
+        // as those rewards are no longer available for distribution
+        let curr_epoch_earned_rewards = match curr_epoch.emission_rate.amount.is_zero() {
+            true => Uint128::zero(),
+            false => self.active_epoch_config.get_total_rewards()?,
+        };
+        self.funded_amount = self.funded_amount.checked_sub(curr_epoch_earned_rewards)?;
```
noahsaso:
```suggestion
        // 1. finish current epoch by changing the end to now
        let mut curr_epoch = self.active_epoch_config.clone();
        curr_epoch.ends_at = current_block_expiration;
        curr_epoch.finish_block = Some(current_block.to_owned());

        // TODO: remove println
        println!("transition_epoch: {:?}", curr_epoch);
        // 2. push current epoch to historic configs
        self.historic_epoch_configs.push(curr_epoch.clone());

        // 3. deduct the distributed rewards amount from total funded amount,
        // as those rewards are no longer available for distribution
        let curr_epoch_earned_rewards = match curr_epoch.emission_rate.amount.is_zero() {
            true => Uint128::zero(),
            false => curr_epoch.get_total_rewards()?,
        };
        self.funded_amount = self.funded_amount.checked_sub(curr_epoch_earned_rewards)?;
```

i'm pretty sure this should be using `curr_epoch`, right? since you're trying to remove the rewards already distributed here, and `curr_epoch` contains the updated `ends_at` from step 1.

bekauz: that makes sense to me!

noahsaso: sweet!

contracts/distribution/dao-rewards-distributor/src/state.rs
```diff
@@ -39,13 +43,15 @@ pub struct EpochConfig {
     /// total rewards earned per unit voting power from started_at to
     /// last_update
     pub total_earned_puvp: Uint256,
-    /// finish block height
-    pub finish_height: Option<BlockInfo>,
+    /// finish block set when epoch is over
+    pub finish_block: Option<BlockInfo>,
```
bekauz: nice one

bekauz:
the changes make sense to me @NoahSaso, thanks for the super thorough review!

noahsaso (approved):
incredible 🎉 

ready for the audit <3

PR #852, opened 2024-07-19
Add `info` query to pre-propose modules

noahsaso:
Add the `info` query that the rest of the contracts support to the pre-propose modules.


JakeHartnell (approved):
Nice. : )

PR #862, opened 2024-08-02
[audit] 1. Rewards are lost if the distribution is funded after the expiry

noahsaso:
From Oak:

> In contracts/distribution/dao-rewards-distributor/src/contract.rs:303, the execute_fund function restarts the distribution if it is not continuous and already expired at the current block. This is problematic because the rewards that accrued from the last updated timestamp (active_epoch.last_updated_total_earned_puvp) to the ending timestamp (active_epoch.ends_at) will be lost.
Consequently, stakers will not receive the rewards for the staking period, causing a loss of rewards.

This fix ensures rewards are calculated and stored before restarting the distribution so they are not lost.

### DIFF
--- a/contracts/distribution/dao-rewards-distributor/src/contract.rs
+++ b/contracts/distribution/dao-rewards-distributor/src/contract.rs
@@ -281,34 +281,104 @@ fn execute_fund_native(
 fn execute_fund(
     deps: DepsMut,
     env: Env,
+    distribution: DistributionState,
+    amount: Uint128,
+) -> Result<Response, ContractError> {
+    match distribution.active_epoch.emission_rate {
+        EmissionRate::Paused {} => execute_fund_paused(deps, distribution, amount),
+        EmissionRate::Immediate {} => execute_fund_immediate(deps, env, distribution, amount),
+        EmissionRate::Linear { .. } => execute_fund_linear(deps, env, distribution, amount),
+    }
+}
+
+/// funding a paused distribution simply increases the funded amount.
+fn execute_fund_paused(
+    deps: DepsMut,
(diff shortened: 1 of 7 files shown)


contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -279,36 +279,106 @@ fn execute_fund_native(
...
-    // will only be true if emission rate is linear and continuous is true
     let continuous =
         if let EmissionRate::Linear { continuous, .. } = distribution.active_epoch.emission_rate {
             continuous
         } else {
             false
         };
+    let previously_funded = !distribution.funded_amount.is_zero();
+    let was_expired = distribution.active_epoch.ends_at.is_expired(&env.block);
+    let discontinuous_expired = !continuous && was_expired;
```
bekauz, line 354: should we refer to previous expirations as discrete rather than discontinuous? latter kind of gives the impression that the rewards are discontinued and not to be restarted, but maybe that's just me 
```suggestion
    let discrete_expired = !continuous && was_expired;
```

bekauz (approved):
lgtm, just one nit about naming

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: hmm, i think continuous has a different meaning in this context. a continuous distribution is one that retroactively distributes rewards when when you fund it after its rewards have run out, making up for the missed time (backfill). a discontinuous distribution starts distributing new funds from the current block if it has previously expired. discrete isn't a meaningful term here, though i'm open to better naming if you have other suggestions!

PR #863, opened 2024-08-02
[audit] 2. Incorrect reward distribution due to inconsistent voting power queries

noahsaso:
From Oak:

> In contracts/distribution/dao-rewards-distributor/src/rewards.rs:113-136, the get_active_total_earned_puvp function computes the reward index by dividing the accrued rewards with total voting power queried from the previous block height (see contracts/distribution/dao-rewards-distributor/src/helpers.rs:19). This is incorrect because the user reward is computed with voting power queried from the current block height, as seen in contracts/distribution/dao-rewards-distributor/src/helpers.rs:33.
> 
> Consequently, the reward index will be computed using an outdated total voting power, resulting in incorrect reward distribution. An attacker can exploit this issue by staking many tokens one block before claiming the rewards to receive more at the expense of other users.

This fix ensures that the total voting power from the same height as the member's voting powers is used.

### DIFF
--- a/contracts/distribution/dao-rewards-distributor/src/helpers.rs
+++ b/contracts/distribution/dao-rewards-distributor/src/helpers.rs
@@ -10,13 +10,13 @@ use dao_interface::voting::{
 
 use crate::ContractError;
 
-pub fn get_prev_block_total_vp(
+pub fn get_total_voting_power_at_block(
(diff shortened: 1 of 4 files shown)


bekauz (approved):
nice!

PR #864, opened 2024-08-02
[audit] 3. Users can inflate their voting power by registering duplicate NFT token IDs

noahsaso:
From Oak:

> In contracts/voting/dao-voting-onft-staked/src/contract.rs:149, the execute_confirm_stake function does not validate that the supplied NFT token IDs are not duplicates. This is problematic because the register_staked_nfts function in contracts/voting/dao-voting-onft-staked/src/state.rs:65-77 increases the user’s voting power based on the total number of token IDs supplied, which may include duplicates.
> 
> An attacker can exploit this issue by calling the execute_confirm_stake function with duplicates of a token ID, granting them a high amount of voting power without requiring them to stake the necessary NFTs. These voting powers can be weaponized to manipulate the DAO into dispatching malicious messages, such as transferring funds to the attacker.

This fix deduplicates token IDs on confirm to prevent double-counting.

### DIFF
--- a/contracts/voting/dao-voting-onft-staked/src/contract.rs
+++ b/contracts/voting/dao-voting-onft-staked/src/contract.rs
@@ -146,9 +146,13 @@ pub fn execute_confirm_stake(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    token_ids: Vec<String>,
+    mut token_ids: Vec<String>,
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
+
+    // de-duplicate token IDs to prevent double-counting exploit
+    token_ids.sort();
+    token_ids.dedup();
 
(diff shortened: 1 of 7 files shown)


contracts/voting/dao-voting-onft-staked/src/contract.rs
```diff
@@ -146,10 +146,14 @@ pub fn execute_confirm_stake(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    token_ids: Vec<String>,
+    mut token_ids: Vec<String>,
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
 
+    // de-duplicate token IDs to prevent double-counting exploit
```
bekauz, line 153: @NoahSaso maybe we can switch this to a set eventually?

bekauz (approved):

PR #871, opened 2024-08-02
[audit] 5. Units of duration are not distinguished

noahsaso:
From Oak:

> Time durations are crucial to the logic of the dao-rewards-distributor contract. However, the data types implemented are error-prone.
>
> For example, the get_exp_diff function in contracts/distribution/dao-rewards-distributor/src/helpers.rs:78 subtracts one Expiration from another. Although the end and start parameters explicitly indicate the unit duration types (i.e., blocks or seconds), this information is not preserved in the output as the function returns u64 instead of the Expiration struct.
>
> Additionally, the get_exp_diff function returns 0 if the input parameters are both Expiration::Never, which is mathematically incorrect.
>
> Similarly, the get_duration_scalar function in line 42 removes the duration unit information and directly returns the inner value as u64, further elevating the potential for errors.
>
> Consequently, the code responsible for managing distribution schedules and timeframes becomes more challenging to interpret, particularly when it involves arithmetic operations with mixed time units. This complexity increases the likelihood of bugs in these segments.

This fix improves names and adds traits to make the code more readable and expands upon the comments to help explain what the helper functions do.

The contract's logic ensures that start and end Expiration units always match the configured emission rate Duration units, so the errors in the helper functions should never be encountered.

### DIFF
--- a/contracts/distribution/dao-rewards-distributor/src/contract.rs
+++ b/contracts/distribution/dao-rewards-distributor/src/contract.rs
@@ -2,7 +2,7 @@
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
     ensure, from_json, to_json_binary, Binary, Deps, DepsMut, Env, MessageInfo, Order, Response,
-    StdResult, Uint128, Uint256,
+    StdError, StdResult, Uint128, Uint256,
 };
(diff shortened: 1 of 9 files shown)


bekauz (approved):
lgtm!

PR #878, opened 2024-08-15
added support for querying the address from an NFT ID, closes #820

### LINKED ISSUE #820: voting-cw721-staked has no way to query address from NFT id
## Affected
- crates.io:dao-voting-cw721-staked 2.3.0
- Looks to be the same in main

## Summary
I am building out https://github.com/Reecepbcups/stargaze-nft-snapshots to snapshot NFT holders. This iterates all NFT id's and exports them to make airdrops easier.

## Issue
Staked DAODAO CW721 are no longer apart of the original collection for Racoon or Rekt bulls due to staking. In doing so, I need to query the crates.io:dao-voting-cw721-staked contract to get `staked id -> stars address`. Currently DAODAO only has `stars address -> [list of staked ids]`

## Solution
It would be nice to add a query to get `nft idx -> user address` that has it staked. Since there is no way to query who is staking without indexing?

theghostmac:
This implementation follows these steps:
1. Query the NFT contract to get the current owner of the token.
2. Check if the token is staked in this contract by iterating through `STAKED_NFTS_PER_OWNER` map.
3. If the token is staked, it returns the staker’s address.
4. If the token is not staked, it returns the owner from the NFT contract. So this way, whether the NFT is staked or not, the owner/staker can be queried.
5. Write tests for:
    1. Querying the owner of a staked NFT.
    2. Querying the owner of an unstaked NFT.
    3. Querying a non-existent nFT.
    4. Querying the owner of an NFT after unstacking
    5. Querying owners for multiple staked NFTs from different users.

### DIFF
--- a/contracts/voting/dao-voting-cw721-staked/src/contract.rs
+++ b/contracts/voting/dao-voting-cw721-staked/src/contract.rs
@@ -17,7 +17,7 @@ use dao_voting::threshold::{
     ActiveThresholdResponse,
 };
 
-use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, NftContract, QueryMsg};
+use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, NftContract, QueryMsg, AddressResponse};
 use crate::state::{
     register_staked_nft, register_unstaked_nfts, Config, ACTIVE_THRESHOLD, CONFIG, DAO, HOOKS,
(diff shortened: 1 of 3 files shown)


ismellike:
We might want to use an [IndexedMap](https://github.com/CosmWasm/cw-storage-plus/tree/main?tab=readme-ov-file#indexedmap) instead to avoid using O(n) lookup on staked tokens.

PR #879, opened 2024-09-06
Proposal Execute Callback Messages

ismellike:
This PR adds support for sending binary data through a proposal execute to improve message composability.

I was working on the [arena-token-gateway](https://github.com/Arena-DAO/arena-smart-contracts/tree/development/contracts/arena-token-gateway) and thought the data was currently being passed up, but it is only done during pre-propose module instantiation.

**Use case:**
DAO Executes Proposal -> Contract Generates Callback Messages -> DAO Receives Binary Data -> DAO Executes Callback Messages

In my case, I want the DAO to create a vesting contract on its `cw-payroll-contract` whenever the DAO accepts a token grant application.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1642,6 +1642,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "dao-callback-messages"
+version = "2.5.0"
+dependencies = [
+ "cosmwasm-schema",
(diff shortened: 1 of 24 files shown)


ismellike:
This can actually be implemented better with authz.

PR #881, opened 2024-09-11
Reward distributor improvements

noahsaso:
This does a few things:

1. Adds an undistributed rewards query to the distributor so that you can check how many rewards would be withdrawn.

2. Fixes a bug pausing an already expired linear distribution. This isn't a vulnerability, but it should be possible, and there was just a minor logic error.

3. Adds the ability to fund the most recent distribution, without knowing its ID. This is useful when creating CW20 distributions. Because we can't verify the sender of a CW20 token send contract execution, we can't allow creating new distributions with initial CW20 funds (since the contract owner must create new distributions). Instead, the DAO can create a distribution AND fund the latest distribution (without knowing its ID) in one proposal.

4. Adds an `open_funding` flag so the owner can decide if a distribution can be funded by accounts other than the owner. The owner can always fund a distribution.

5. Fixes a bug causing pending rewards and undistributed rewards queries to fail if a linear distribution has not yet been funded.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,12 +271,12 @@ dependencies = [
  "cw-admin-factory",
  "cw-utils 1.0.3",
  "cw20 1.1.2",
- "cw20-stake 2.5.0",
- "dao-dao-core 2.5.0",
- "dao-interface 2.5.0",
- "dao-pre-propose-single 2.5.0",
- "dao-proposal-single 2.5.0",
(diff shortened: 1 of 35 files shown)


JakeHartnell (approved):

bekauz (approved):
great findings, especially the one to do with pausing an already expired distribution. totally missed it before.

lgtm!

PR #882, opened 2024-09-12
change versions back to v2.5.0

noahsaso:
the DAO rewards distributor was never released, so i'm just going to re-release v2.5.0 instead of making a whole new version. #881 changed to v2.5.1, so this changes it back.


JakeHartnell (approved):
Sounds sensible to me. :joy:

PR #883, opened 2024-09-17
Add pre-propose-approval support for multiple choice proposals, and freeze approver

noahsaso:
this does two things:

1. it creates a dao-pre-propose-approval-multiple contract so that multiple choice proposal modules can have approvers, just like single choice proposal modules. it also ensures the approver contract works for both approval-single and approval-multiple

2. it freezes the `approver` in the proposal so that a previously created pending proposal does not change approvers if the approver gets updated while it's still pending. this maintains consistency with other DAO logic to keep things intuitive, where the members who can vote on a proposal are frozen upon proposal creation.

### DIFF
--- a/contracts/pre-propose/dao-pre-propose-approval-multiple/src/state.rs
+++ b/contracts/pre-propose/dao-pre-propose-approval-multiple/src/state.rs
@@ -1,0 +1,21 @@
+use cosmwasm_std::{Addr, StdResult, Storage};
+use cw_storage_plus::{Item, Map};
+
+use dao_voting::{approval::ApprovalProposal, proposal::MultipleChoiceProposeMsg};
+
+pub type Proposal = ApprovalProposal<MultipleChoiceProposeMsg>;
+
+pub const APPROVER: Item<Addr> = Item::new("approver");
+pub const PENDING_PROPOSALS: Map<u64, Proposal> = Map::new("pending_proposals");
+pub const COMPLETED_PROPOSALS: Map<u64, Proposal> = Map::new("completed_proposals");
+pub const CREATED_PROPOSAL_TO_COMPLETED_PROPOSAL: Map<u64, u64> =
+    Map::new("created_to_completed_proposal");
+
+/// Used internally to track the current approval_id.
+const CURRENT_ID: Item<u64> = Item::new("current_id");
+
+pub(crate) fn advance_approval_id(store: &mut dyn Storage) -> StdResult<u64> {
+    let id: u64 = CURRENT_ID.may_load(store)?.unwrap_or_default() + 1;
+    CURRENT_ID.save(store, &id)?;
+    Ok(id)
+}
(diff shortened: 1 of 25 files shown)


contracts/pre-propose/dao-pre-propose-approval-multiple/src/state.rs
```diff
@@ -0,0 +1,21 @@
...
+pub const APPROVER: Item<Addr> = Item::new("approver");
+pub const PENDING_PROPOSALS: Map<u64, Proposal> = Map::new("pending_proposals");
+pub const COMPLETED_PROPOSALS: Map<u64, Proposal> = Map::new("completed_proposals");
+pub const CREATED_PROPOSAL_TO_COMPLETED_PROPOSAL: Map<u64, u64> =
+    Map::new("created_to_completed_proposal");
+
+/// Used internally to track the current approval_id.
+const CURRENT_ID: Item<u64> = Item::new("current_id");
+
+pub(crate) fn advance_approval_id(store: &mut dyn Storage) -> StdResult<u64> {
```
bekauz, line 17: nice!

contracts/pre-propose/dao-pre-propose-approval-single/src/contract.rs
```diff
@@ -406,7 +414,129 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                                }
+                            },
+                        }),
+                    },
+                )?;
+            }
+
+            let completed_proposals =
+                dao_pre_propose_approval_single_v241::state::COMPLETED_PROPOSALS
+                    .range(deps.storage, None, None, Order::Ascending)
```
bekauz: wonder from how many props this could fail. maybe worth to do some test runs?

bekauz (approved):
this looks great, I don't have much to add.

maybe except that the `dao_pre_propose_base` design combined with message extensions is super clean and holding up really well!

contracts/pre-propose/dao-pre-propose-approval-single/src/contract.rs
noahsaso: yeah.. what's the easiest way to test this?

bekauz: both test-tube and interchaintest provide this. probably interchaintest would be easiest to test across multiple chains, but require a bit more upfront setup than test-tube

noahsaso: ok but do we really need this? we don't have a rigorous upgrade flow, and the migration will just fail anyways if it runs out of gas. imposing an arbitrary limit serves no real purpose

noahsaso: alternatively, we could make the approver field optional, and if it's empty, default to the currently configured approver! that way, it's backwards compatible and requires no migration 🙌

noahsaso: > alternatively, we could make the approver field optional, and if it's empty, default to the currently configured approver! that way, it's backwards compatible and requires no migration 🙌

actually nevermind. if any of the types change serialization in the future, not having a migration flow that force-serializes with new types can lead to unparseable state. dang.

PR #886, opened 2024-09-23
Fix CW20 funding bug in reward distributor

noahsaso:
This fixes a bug in the reward distributor that requires distributions to have `open_funding` enabled in order to be funded with a CW20 token.


bekauz (approved):
damn, crazy how this went unnoticed

PR #888, opened 2024-10-10
v2.7.0 - Vote Delegation, Initial Actions, Deep Instantiate2 Support

noahsaso:
# DAO Vote Delegation

The `dao-vote-delegation` contract allows members of a DAO to delegate their
voting power to other members of the DAO who have registered as delegates. It
works in conjunction with voting and proposal modules to offer a comprehensive
delegation system for DAOs that supports the following features:

- Fractional delegation of voting power on a per-proposal-module basis.
- Delegate votes that can be overridden on a per-proposal basis by each
  delegator.
- Configurable cap that restricts the maximum amount of voting power that a
  single delegate can wield when casting votes.
- Configurable expiration period for delegation.
- Configurable limit on the number of delegations a member can have.

## Instantiation and Setup

This contract must be instantiated by the DAO.

### Hooks

After instantiating the contract, it is VITAL to set up the required hooks for
it to work. To compute delegate voting power correctly, this contract needs to
know about both voting power changes and votes cast on proposals as soon as they
happen.

This can be achieved using the `add_hook` method on voting/staking contracts
that support voting power changes, such as:

- `cw4-group`
- `dao-voting-cw721-staked`
- `dao-voting-token-staked`
- `cw20-stake`

For proposal modules, the corresponding hook is `add_vote_hook`:

- `dao-proposal-single`
- `dao-proposal-multiple`

### Fractional Delegation via Percentages

In order to support fractional delegation, users assign a percentage of voting
power to each delegate. Percentages are used instead of choosing an absolute
amount of voting power (e.g. staked tokens) since voting power can change
independently of delegation. If an absolute amount were used, and a user who had
delegated all of their voting power to a few different delegates then unstaked
half of their tokens, there is no clear way to resolve what their new
delegations are. Using percentages instead allows voting power and delegation to
be decided independently.

## Implementation Notes

The trickiest piece of this implementation is navigating the snapshot maps,
which are the data structures used to store historical state.

Essentially, snapshot maps (and the other historical data structures based on
snapshot maps) take 1 block to reflect updates made, but only when querying
state at a specific height (typically in the past). When using the query
functions that do not accept a height, they read the updates immediately,
including those from the same block. For example, `snapshot_map.may_load`
returns the latest map values, including those changed in the same block by an
earlier transaction; on the other hand, `snapshot_map.may_load_at_height`
returns the map values as they were at the end of the previous block (due to an
implementation detail of snapshot maps that I'm not sure was intentional).

Ideally, we would just fix this discrepancy and move on. However, many other
modules have been built using SnapshotMaps, and it is important that all modules
behave consistently with respect to this issue. For example, voting power
queries in voting modules operate in this way, with updates delayed 1
block—because of this, it is crucial that we compute and store delegated voting
power in the same way. Otherwise we risk introducing off-by-one inconsistencies
in voting power calculations. Thus, for now, we will accept this behavior and
continue.

What this means for the implementation is that we must be very careful whenever
we do pretty much anything. When performing updates at the latest block, such as
when delegating or undelegating voting power, or when handling a change in
someone's voting power (in order to propagate that change to their delegates),
we need to be sure to interact with the latest delegation and voting power
state. However, when querying information from the past, we need to match the
delayed update behavior of voting power queries.

More concretely:

- when registering/unregistering a delegate, delegating/undelegating, or
  handling voting power change hooks, we need to access the account's latest
  voting power (by querying `latest_height + 1`), even if it was updated in the
  same block. this ensures that changes to voting power right before a
  registration/delegation occurs, or voting power changes right after a
  delegation occurs, are taken into account. e.g. an account should not be able
  to get rid of all their voting power (i.e. stop being a member) and then
  become a delegate within the same block.
- when delegating/undelegating or handling voting power change hooks, in order
  to update a delegate's total delegated VP, we need to query the latest
  delegated VP, even if it was updated earlier in the same block, and then
  effectively "re-prepare" the total that will be reflected in historical
  queries starting from the next block. `snapshot_map.update` takes care of this
  automatically by loading the latest value from the same block.
- when querying information from the past, such as when querying a delegate's
  total unvoted delegated VP when they cast a vote, or when a vote cast hook is
  triggered for a delegator, we need to use historical queries that match the
  behavior of the voting module's voting power queries, i.e. delayed by 1 block.

### Voting Module Compatibility

The delegation module expects the voting power module to realize voting power
changes on the block following a change. In DAO DAO's voting power contracts,
this is accomplished by using CosmWasm's `SnapshotMap` storage type.

More concretely: staking tokens on block `n` will not reflect the change in
voting power until block `n+1`. Querying voting power at height `n` will return
the voting power as it was at the beginning of block `n`, before any
transactions occurred, after any changes from `n-1`. Querying `n+1` is the first
block where the voting power changes from `n` will be reflected.

If custom voting power modules are used, the voting module must ensure that the
1-block delay is respected, or else proposal vote tallies will be incorrect.

### Voting Power Granularity

Because voting power is floored when multiplying by delegation percentages, the
granularity of delegation is restricted to the unit size used by the voting
power module.

For example, if a DAO is using the `cw4-group` voting power module—a
multisig-like structure with static membership weights—and a member has a weight
of `1`, then they can only delegate all their voting power to a single delegate.
Delegating any less than 100% will be floored to 0 and effectively nullify the
delegation. The DAO can remedy this by increasing the order of magnitude of
voting power weights, giving members a weight of 1,000 (or more) instead of 1,
for example.

This is particularly problematic for NFT-based DAOs, since 1 NFT corresponds
with 1 voting power unit: the number of NFTs a member has staked will determine
how many delegations they can have and the percentages that can be used.

For token-based DAOs, this isn't really an issue, since tokens usually have
divisible units with a precision of at least 6 decimal places, meaning 1 token
in a user's wallet is actually 1,000,000 token units.

### Delegation Limits

Because all math is done on-chain, the block gas limit configured by the chain
determines the maximum number of delegations a user can have. When a delegator's
voting power changes, or a delegator's vote overrides their delegates' votes on
a proposal, the computation needed to update all the relevant state depends on
the number of delegations. This is unavoidable, unless the computation were to
be moved off-chain.

### Delegates Cannot Delegate

Delegates cannot delegate their voting power to other delegates. This is
technically possible to implement, though it would be more complex and approach
computation limits must faster, so it was not included in this first version.

### Delegation Expiration Defined in Blocks

Automatic delegation expiry must be configured in blocks because proposals
freeze members' voting power at the block when the proposal was created, and
historical queries don't have access to timestamps associated with past blocks
(only the current one).

# Initial Actions

The creator of the DAO can specify actions to be executed on behalf of the DAO during instantiation. A query is added so that the initial actions can be queried, which the UI should clearly display in order to address security concerns by allowing the DAO creator to arbitrarily execute messages on behalf of the DAO.

# Other refactors

`ModuleInstantiateInfo`, which is used to add modules to the DAO, now contains an optional `salt` field, which will use `instantiate2` to instantiate contracts instead of the normal instantiate so that the caller can predict the addresses of any module it is adding. This also replaces many hardcoded `WasmMsg::Instantiate` definitions with `ModuleInstantiateInfo` to allow other modules to take advantage of this mechanism and let the caller decide whether or not to use instantiate2. For example, the dao-voting-cw20-staked contract can optionally instantiate the cw20-stake contract with instantiate2 or not—same for dao-voting-cw4's cw4-group contract, among other contracts.

This change is necessary to make the initial actions useful, as the caller can create initial actions that operate on the voting and proposal modules (and they'd need to know the addresses before they're instantiated to do so).

Also, a while ago, the `funds` field was added to `ModuleInstantiateInfo`, making some versions of modules incompatible with other versions. This refactor makes that optional, so that old versions should work fine with newer modules, provided they still use the same interfaces (which we have never changed).


contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,757 @@
...
+        ExecuteMsg::MemberChangedHook(msg) => execute_membership_changed(deps, env, info, msg),
+        ExecuteMsg::VoteHook(vote_hook) => execute_vote_hook(deps, env, info, vote_hook),
+    }
+}
+
+fn execute_register(deps: DepsMut, env: Env, info: MessageInfo) -> Result<Response, ContractError> {
+    nonpayable(&info)?;
+    ensure_setup(deps.as_ref())?;
+
+    let delegate = info.sender;
```
bekauz: wyt if we move the `nonpayable` call into the execute entry point handler before the match statement?
that way we could avoid passing `MessageInfo` just in order to do `let delegate/delegator = info.sender` inside the functions. on the other hand, that would mean that the hooks incur a redundant check.

noahsaso: ah yes great idea, since none of the contract takes funds. i'll push that here

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,745 @@
...
+        ExecuteMsg::VoteHook(vote_hook) => execute_vote_hook(deps, env, info.sender, vote_hook),
+    }
+}
+
+fn execute_register(deps: DepsMut, env: Env, delegate: Addr) -> Result<Response, ContractError> {
+    ensure_setup(deps.as_ref())?;
+
+    if is_delegate_registered(deps.as_ref(), &delegate, None)? {
+        return Err(ContractError::DelegateAlreadyRegistered {});
+    }
```
bekauz, line 159: nice check

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,745 @@
...
+    ensure_setup(deps.as_ref())?;
+
+    if !is_delegate_registered(deps.as_ref(), &delegate, None)? {
+        return Err(ContractError::DelegateNotRegistered {});
+    }
+
+    unregister_delegate(deps, &delegate, env.block.height)?;
+
+    Ok(Response::new())
+}
```
bekauz, line 195: clean!

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,745 @@
...
+
+    // delegates cannot delegate to others
+    if is_delegate_registered(deps.as_ref(), &delegator, None)? {
+        return Err(ContractError::DelegatesCannotDelegate {});
+    }
+
+    // ensure delegate is registered
+    if !is_delegate_registered(deps.as_ref(), &delegate, None)? {
+        return Err(ContractError::DelegateNotRegistered {});
+    }
```
bekauz: this is great. I love how it implicitly asserts that this is not a self delegation because if `delegate == delegator`, both `is_delegate_registered` and `!is_delegate_registered` could not evaluate to true.

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    // check latest state instead of historical height, since we need access to
+    // immediate updates made earlier in the same block
+    if is_delegate_registered(deps.as_ref(), &addr, None)? {
+        let delegate = addr;
+
+        // unregister if no more voting power
+        if new_vp.is_zero() {
+            unregister_delegate(deps, &delegate, env.block.height)?;
+        }
```
bekauz: nice check

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -0,0 +1,222 @@
...
+            voter,
+            power,
+            height,
+            is_first_vote,
+            ..
+        } => {
+            // if first vote, update the unvoted delegated VP for their
+            // delegates by subtracting this member's delegated VP. if not first
+            // vote, this has already been done.
+            if is_first_vote {
```
bekauz: great

contracts/proposal/dao-proposal-multiple/src/contract.rs
bekauz, line 1: I really like how contained the delegation logic is from the proposals. Feels like if anything was to go wrong with delegations, the proposals shouldn't be affected.

contracts/delegation/dao-vote-delegation/README.md
```diff
@@ -0,0 +1,170 @@
...
+- `cw4-group`
+- `dao-voting-cw721-staked`
+- `dao-voting-token-staked`
+- `cw20-stake`
+
+For proposal modules, the corresponding hook is `add_vote_hook`:
+
+- `dao-proposal-single`
+- `dao-proposal-multiple`
+- `dao-proposal-condorcet`
```
bekauz: We don't support condorcet for delegations yet, do we? Maybe worth removing this for now to avoid confusion

contracts/delegation/dao-vote-delegation/README.md
```diff
@@ -0,0 +1,170 @@
...
+voting power changes, or a delegator's vote overrides their delegates' votes on
+a proposal, the computation needed to update all the relevant state depends on
+the number of delegations. This is unavoidable, unless the computation were to
+be moved off-chain.
+
+### Delegates Cannot Delegate
+
+Delegates cannot delegate their voting power to other delegates. This is
+technically possible to implement, though it would be more complex and approach
+computation limits must faster, so it was not included in this first version.
```
bekauz:
```suggestion
computation limits much faster, so it was not included in this first version.
```

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,807 @@
...
+            deps.storage,
+            &delegator,
+            &delegation,
+            env.block.height,
+            config.delegation_validity_blocks,
+        )?;
+
+        // don't let them update if they are over the max, instead requiring
+        // them to remove existing delegations before updating any
+        ensure_max_delegations_not_reached(config.max_delegations, total_count, total_count)?;
```
bekauz: took me a minute to understand why is this check needed given that we are doing `DELEGATIONS.remove` and `DELEGATIONS.push` above. but it ensures that if `max_delegations` in the config get updated to a smaller value,   any users who may have been delegating to the max will need to undelegate to get back within limits before being able to delegate again. nice.

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,807 @@
...
+            let cap = calculate_delegated_vp(total_power, vp_cap_percent);
+
+            effective = total.min(cap);
+        }
+    }
+
+    Ok(UnvotedDelegatedVotingPowerResponse { total, effective })
+}
+
+fn query_effective_unvoted_delegated_vote_power_reduction(
```
bekauz, line 629: love the comments here as it is one loaded query

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -0,0 +1,342 @@
...
+                // remove the delegator's delegated VP from the delegate's
+                // unvoted delegated VP for this proposal since this
+                // delegator just voted.
+                let new_udvp = udvp.checked_sub(delegated_vp)?;
+
+                UNVOTED_DELEGATED_VP.save(
+                    deps.storage,
+                    (&delegate, &proposal_module, proposal_id),
+                    &new_udvp,
+                )?;
```
bekauz, line 304: This seems to be the only thing we need to ensure that each unit of vp is only counted once, right?

contracts/proposal/dao-proposal-single/src/contract.rs
```diff
@@ -529,15 +559,36 @@ pub fn execute_vote(
...
+            proposal_id,
+            prop.start_height,
+            &vote_power.individual,
+            BALLOTS,
+            &mut |vote, power| {
+                prop.votes.remove_vote(*vote, power);
+                Ok(())
+            },
+        )?;
+    }
```
bekauz, line 586: I like this design. can't think of a lower lift to separate the concerns between the delegations module and proposals. i think even if something went ultra wrong with delegations, every member voting this way would be all that's needed to assert the final say in that VP, right?

contracts/delegation/dao-vote-delegation/src/testing/tests.rs
```diff
@@ -0,0 +1,2286 @@
...
+
+    // ensure new proposal module is synced
+    assert_eq!(
+        suite.proposal_modules(None, None).len(),
+        dao.proposal_modules.len() + 1
+    );
+}
+
+#[test]
+fn test_vote_with_override() {
```
bekauz, line 980: nice test!

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,807 @@
...
+
+pub(crate) const CONTRACT_NAME: &str = "crates.io:dao-vote-delegation";
+pub(crate) const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub const DEFAULT_LIMIT: u32 = 10;
+
+/// in tests on Neutron, with a block max gas of 30M (which is one of the lowest
+/// gas limits on any chain), we found that 50 delegations is a safe upper
+/// bound, so this defaults to 50.
+pub const DEFAULT_MAX_DELEGATIONS: u64 = 50;
```
bekauz, line 47: honestly, while technically this is a limitation, I like the way it works.

this places the gas burden throughout the delegation (and proposal) lifecycle almost entirely on the delegators. that feels right - the delegates should not have to worry about stuff like monitoring their delegators which may be misbehaving or something.

contracts/proposal/dao-proposal-multiple/src/proposal.rs
```diff
@@ -135,26 +166,32 @@ impl MultipleChoiceProposal {
             }
         }
 
+        let votes_to_consider = if is_expired {
+            &self.votes
+        } else {
+            &self.individual_votes
+        };
```
bekauz, line 173: simple enough!

bekauz (approved):

contracts/delegation/dao-vote-delegation/src/contract.rs
noahsaso: omg brilliant. i hadn't fully thought to this conclusion but you're totally right—a delegator can only ever affect their own gas requirements. they can't cause anyone else voting or voting power change difficulties, only themselves. 50 should be conservatively low enough that they never run into any gas issues, and if they somehow max out their storage with expired delegations, max out the delegation limit, and do find themselves in that position, it can't hurt anyone else.

contracts/proposal/dao-proposal-single/src/contract.rs
noahsaso: yep i think so

contracts/proposal/dao-proposal-multiple/src/contract.rs
noahsaso: yeah, it's as maximally separated as possible i believe. though the proposal module does have to trust that the delegation module tells it the right amount to subtract during a vote override, so there's no surefire way to safeguard proposals

contracts/delegation/dao-vote-delegation/README.md
noahsaso: ah right, good idea

contracts/delegation/dao-vote-delegation/src/contract.rs
noahsaso: hell yeah

contracts/delegation/dao-vote-delegation/src/hooks.rs
noahsaso: yep, UDVP is the source of truth for how much voting power a delegate should have tallied in a proposal. at the end of every TX, a proposal ballot for a delegate should be their personal VP + their UDVP value (with the cap applied if any)

packages/dao-interface/src/state.rs
```diff
@@ -157,4 +182,78 @@ mod tests {
...
+        let no_admin = ModuleInstantiateInfo {
+            code_id: 42,
+            msg: to_json_binary("foo").unwrap(),
+            admin: None,
+            label: "bar".to_string(),
+            funds: Some(coins(100, "uatom")),
+            salt: Some(to_json_binary("test_salt").unwrap()),
+        };
+        assert_eq!(
+            no_admin.into_wasm_msg(Addr::unchecked("ekez")),
```
bekauz, line 197: <3

packages/cw-snapshot-vector-map/src/lib.rs
bekauz, line 1: great work here by the way! it's already confirmed to be working via delegations so not much feedback here.

do you think in the future this could support updating values in a more efficient way than doing `remove() + push()`? one thing that made me think of this is a situation where some user x wants to update the expiration (but keep the value), while user y wants to update the value (but keep the expiration). I guess I wondered if it was possible to separate the storage logic from expiration?

packages/dao-hooks/src/vote.rs
```diff
@@ -1,32 +1,54 @@
...
         proposal_id: u64,
+        /// The voter that cast the vote.
         voter: String,
+        /// The vote that was cast.
         vote: String,
+        /// The total voting power of the voter.
+        power: Uint128,
+        /// The individual voting power of the voter (excluding any delegated
+        /// voting power).
+        individual_power: Uint128,
```
bekauz, line 20: nice

bekauz (commented):
lets get this in 🫡

packages/dao-interface/src/state.rs
noahsaso: 😭

packages/cw-snapshot-vector-map/src/lib.rs
noahsaso: yeahh i've thought about that a little but haven't explored it too deeply

because this has to exist over time, the only case in which an update can just change the value stored in the main list is if they update it on the same block as the original item was created on. changing a value after it was added can only take effect from that point on, so in effect it is really two different values.

but changing expiration can definitely be done more efficiently, yeah. that's a great idea.

wdym by separate storage logic from expiration?

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -0,0 +1,807 @@
...
+        let mut delegation = DELEGATIONS.load_item(deps.storage, &delegator, id)?;
+
+        // remove existing percent and replace with new percent
+        let new_total = current_percent_delegated
+            .checked_sub(delegation.percent)?
+            .checked_add(percent)?;
+
+        // remove current delegated VP based on existing percent
+        let old_vp = calculate_delegated_vp(vp, delegation.percent);
+        remove_delegated_vp(deps.storage, &env, &delegate, old_vp, expiration)?;
```
noahsaso: i found another bug! and this one is a doozy 😵‍💫

basically, `remove_delegated_vp` is the inverse of `add_delegated_vp`, so we've been using it to undo a previous delegation before adding a new one, EXCEPT for one key distinction: `add_delegated_vp` always operates on the new delegations (created at the current block and expiring in the future), but `remove_delegated_vp` operates on existing delegations which were created at previous blocks and expire sometime after their creation. this means we have to be very careful how we use `remove`...

the problem is that we were implicitly using `remove_delegated_vp` as if it were operating on delegations that were created in the past but expire in the future, BUT its callers (`execute_delegate` and `execute_undelegate`) were never verifying that the delegation was still active. `add_delegated_vp` configures `DELEGATED_VP` to auto-decrement in the future—that's why we're using Wormhole for it, to automatically enforce expirations—so we don't actually need to call `remove_delegated_vp` if a delegation is already expired. and `DELEGATIONS` also automatically filters out expired items, so when loading from it (like in `handle_delegator_voting_power_changed_hook`), we can safely know that delegations it loads are not yet expired.

where delegations are NOT automatically expired is in `DELEGATION_ENTRIES`. `execute_delegate` stores the ID and expiration of the created delegation in `DELEGATION_ENTRIES` for constant time delegation loading in future delegates/undelegates (`execute_delegate` and `execute_undelegate`). `DELEGATION_ENTRIES` thus will return the delegation ID and expiration even after the delegation is expired, and thus the callers should be only removing delegated VP if not already expired, since if it's already expired, the `DELEGATED_VP` has already been decremented thanks to Wormhole. `execute_undelegate` was removing the entry from `DELEGATION_ENTRIES`, and the rest of the code was incorrectly assuming that `DELEGATION_ENTRIES` would be updated on an expiration without any explicit undelegation, but that's not the case.

funnily enough, this bug went unnoticed because `remove_delegated_vp`, in assuming that the expiration is in the future, first undoes the auto-expiration decrement (by incrementing at the expiration), and then removes the delegated VP by decrementing at the current block. i initially thought we should be seeing underflow errors trying to decrement a value that should already have been decremented, but in fact the bug is subtler than that: when the expiration was in the past, it was "undoing" the expiration by incrementing in the past, before decrementing at the current block—in effect, it was rewriting the past to pretend like the delegation did not expire until the current block, but it would be out of sync with the `DELEGATIONS` state that says there is no delegation during that range.

this causes issues because all past state must be frozen. say a delegation expires, then a proposal is created, and then the delegator redelegates to the same delegate that expired before the proposal was created. now when the delegate goes to vote on behalf of all their delegators, their `DELEGATED_VP` would be too high because the expiration that should've expired was accidentally extended, BUT `DELEGATIONS` still accurately auto-expired the delegation. so when the delegator goes to override their delegates' votes, the vote hook does not think that a delegation exists for that original delegate, and thus does not untally any of the delegator's VP from the delegate's vote, thus giving the delegate too much power and making the total voting power turnout more than available in the DAO.

i think the cleanest fix is to simply make `remove_delegated_vp` check if `original_expiration` is less than or equal to the current block, and simply do nothing if so :)

noahsaso: fixed in 5c92e743975afdc4eb50db2ee108455b3aa2de6e

i discovered this bug while adding support for expiration-only updates

i made `remove_delegated_vp` check that expiration is in the past, and i also made the callers check if the delegation is expired and act accordingly

~the delegate function is now pretty ugly, but it's all in the name of efficiency...~ ok i created a bunch of helpers that encapsulate less logic so it's more readable :)

packages/cw-snapshot-vector-map/src/lib.rs
noahsaso: ok i added support for expiration-only updates to `SnapshotVectorMap` and delegations in `execute_delegate` in 5c92e743975afdc4eb50db2ee108455b3aa2de6e. i also found a critical bug in doing so, which is great.

for now, you pretty much have to remove and and push, but i did make a new `update` function that makes the API a bit cleaner than remove+push and reduces the overhead

packages/cw-snapshot-vector-map/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// the total number of items remaining. The block height should be greater
+    /// than or equal to the blocks all previous items were added/removed at.
+    /// Updating in the past will lead to incorrect behavior.
+    ///
+    /// This is a convenient way to update items, and reduces the overhead of a
+    /// remove and then push, but it does still create a new item since we need
+    /// to preserve the old item for historical queries. This works even if the
+    /// item is expired since IDs are preserved indefinitely for historical
+    /// queries.
+    pub fn update(
```
bekauz, line 176: yeah this definitely feels like the right approach. even beyond added convenience, it will make it harder to mess something up when doing updates manually.

packages/cw-snapshot-vector-map/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+        // ensure remove operations are performed at or after the last update
+        self.validate_order_and_update(store, k, curr_height)?;
+
+        // get active list for the key
+        let mut active = self.active.may_load(store, k.clone())?.unwrap_or_default();
+
+        // remove existing item and any expired items
+        active.retain(|(active_id, expiration)| {
+            active_id != &id && expiration.map_or(true, |expiration: u64| expiration > curr_height)
+        });
```
bekauz, line 194: I feel like by just doing `.retain` on the active list we are opening ourselves up for situations where updating an expired delegation effectively functions as creating a new delegation. Is that the desired outcome here? imho returning an error here would be more clear


```suggestion
        // get active list for the key
        let mut active = self.active.may_load(store, k.clone())?.unwrap_or_default();

        // prior to update, verify that the item exists and is not expired
        let target_id_exists = active.iter().any(|(active_id, expiration)| {
            *active_id == id && expiration.map_or(true, |exp| exp > curr_height)
        });

        if !target_id_exists {
            return Err(StdError::generic_err("item not found or expired"));
        }

        // remove existing item and any expired items
        active.retain(|(active_id, expiration)| {
            active_id != &id && expiration.map_or(true, |expiration: u64| expiration > curr_height)
        });
```

bekauz: I've pushed a test that asserts this: https://github.com/DA0-DA0/dao-contracts/blob/a647cb3f2e8385dfb62a501d45c5e42196b03ddd/packages/cw-snapshot-vector-map/src/tests.rs#L622-L647 

If you feel like this is the expected behavior then please feel free to remove it! I noticed that a few delegation tests fail as a result of this suggested change and I didn't want to alter the core logic before getting your take on it

bekauz (commented):
snapshot vector map changes make sense to me except for that one point which i highlighted above.

I pushed some more unit tests there to confirm my own understanding. During that I discovered that updating an expired item creates a new entry starting at the next block which I found unexpected, lmk what you think regarding that!

contracts/delegation/dao-vote-delegation/src/helpers.rs
```diff
@@ -0,0 +1,391 @@
...
+    // expiration based on the current config.
+    if !expired && existing_delegation_percent == new_percent {
+        // if both expirations are none, do nothing.
+        if existing_delegation_expiration.is_none() && config.delegation_validity_blocks.is_none() {
+            return Ok((
+                current_percent_delegated,
+                (existing_delegation_id, existing_delegation_expiration),
+                false,
+            ));
+        }
```
bekauz, line 262: nice early return

contracts/delegation/dao-vote-delegation/src/helpers.rs
```diff
@@ -0,0 +1,391 @@
...
+        )?;
+
+        update_delegated_vp_expiration(
+            deps.storage,
+            env,
+            delegate,
+            vp,
+            existing_delegation_expiration,
+            new_expiration,
+        )?;
```
bekauz, line 279: passing the `update_expiration` return value as the input for further calls is great!

contracts/delegation/dao-vote-delegation/src/helpers.rs
```diff
@@ -0,0 +1,391 @@
...
+    let (entry, total_count) = DELEGATIONS.update(
+        deps.storage,
+        delegator,
+        existing_delegation_id,
+        env.block.height,
+        |d| {
+            d.percent = new_percent;
+        },
+        config.delegation_validity_blocks,
+    )?;
```
bekauz, line 317: clean update api! love passing `Fn` for the update

bekauz:
I like how the core delegation module logic is now kept more clean and concise in the `execute_msg` handlers.

Having the different helpers abstract over the underlying logic (especially when it comes to passing the block heights for snapshot map-based structures) really makes it easier to follow!

packages/cw-snapshot-vector-map/src/lib.rs
noahsaso: memorializing our conversation here:

yeah, it is the expected behavior that updating an expired item creates a new entry. in fact, updating any item creates a new entry, since we need to preserve old data for historical queries. `update` is just a convenience method for loading a known item by its ID, modifying it, and saving a new item at the current block height.

IMO we don't need to error if the item is expired here because it's not violating any principles of the data structure—the caller should know what they're doing if they're saving an ID and updating it later. if we made this error, and the caller had an ID they wanted to modify, they would have to load it manually, explicitly check it's expiration, conditionally remove it, and then push the modified/new item. this `update` just encapsulates all of that and handles all the cases cleanly. either way, they want to modify some data that they know existed at some point and push a new item, so that's what this `update` is for.

noahsaso:
great work @bekauz, this is a huge accomplishment and i couldn't have done it without you 🥳

i feel very good about what it has become and think this is close to the best we could ask for given the current limitations of on-chain compute. some gnarly data structures in there 🤯

just one (huge) step closer to building transparent and scalable digital democracies

and thanks @ismellike for the initial actions initiative! that is going to enable a lot of cool stuff

PR #890, opened 2024-10-15
organize tests

noahsaso:
this organizes tests a bit more and prepares a suite to help create DAOs. it could be improved, but it is good enough for now to use for #888, and it will be improved further as it is used in creating a new set of tests.

- moved all contract wrappers to dao-testing
- created a dao-testing suite that helps quickly create DAOs with any of the various voting modules
- rewrote dao-rewards-distributor tests to use dao-testing suite as a proof of concept


JakeHartnell (approved):
❤️

PR #891, opened 2024-10-23
fix reward distributor interval delay bug

noahsaso:
the reward distribution contract currently suffers a bug where reward distribution intervals are delayed each time a voting power change occurs.

this is due to the fact that integer division is used to make it so that only increments of the linear emission rate duration are distributed. when a voting power change occurs before one interval has passed, it calculates 0 new rewards distributed and updates the last updated timestamp, which effectively restarts the interval timer.

for example, if you distribute 100 tokens per hour:
- at minute 0, timer starts counting
- at minute 30, member stakes more tokens, and timer restarts
- at minute 60, nothing happens
- at minute 90, first distribution occurs

this PR fixes it so that it uses decimal math when counting the intervals that have passed since the last update in order to update the rewards distributed.

this also adds a force withdraw execution message to allow the contract owner to recover locked funds, since the delayed funds are likely stuck in the contract.

## recover delayed/locked rewards

i think the ideal way to remedy this is:

- pause all distributions
- withdraw all undistributed funds from the distributions
- upgrade the contract to fix the partial-interval distribution and add the force withdraw message
- calculate the missed rewards by looking at what's left over in the contract and then looping through everyone's pending rewards and subtracting those from the left over amounts
- force withdraw that unaccounted for amount, leaving all pending rewards that were successfully distributed intact
- restart the distributions and fund them again

this way the state should be reset to account for the pending rewards that were allegedly distributed, and you can manually withdraw any funds beyond that amount before restarting and refunding.

### DIFF
--- a/contracts/distribution/dao-rewards-distributor/src/contract.rs
+++ b/contracts/distribution/dao-rewards-distributor/src/contract.rs
@@ -1,8 +1,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    ensure, from_json, to_json_binary, Addr, Binary, Deps, DepsMut, Env, MessageInfo, Order,
-    Response, StdError, StdResult, Uint128, Uint256,
+    ensure, from_json, to_json_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut, Env,
+    MessageInfo, Order, Response, StdError, StdResult, Uint128, Uint256,
 };
 use cw2::{get_contract_version, set_contract_version};
 use cw20::{Cw20ReceiveMsg, Denom};
@@ -91,6 +91,9 @@ pub fn execute(
         ExecuteMsg::FundLatest {} => execute_fund_latest_native(deps, env, info),
         ExecuteMsg::Claim { id } => execute_claim(deps, env, info, id),
         ExecuteMsg::Withdraw { id } => execute_withdraw(deps, info, env, id),
+        ExecuteMsg::UnsafeForceWithdraw { amount } => {
+            execute_unsafe_force_withdraw(deps, info, amount)
+        }
     }
 }
 
@@ -594,6 +597,27 @@ fn execute_update_owner(
     Ok(Response::new().add_attributes(ownership.into_attributes()))
 }
 
+fn execute_unsafe_force_withdraw(
+    deps: DepsMut,
+    info: MessageInfo,
+    amount: Coin,
+) -> Result<Response, ContractError> {
+    nonpayable(&info)?;
+
+    // only the owner can initiate a force withdraw
+    cw_ownable::assert_owner(deps.storage, &info.sender)?;
+
+    let send = CosmosMsg::Bank(BankMsg::Send {
+        to_address: info.sender.to_string(),
+        amount: vec![amount.clone()],
+    });
+
+    Ok(Response::new()
+        .add_message(send)
+        .add_attribute("action", "unsafe_force_withdraw")
+        .add_attribute("amount", amount.to_string()))
+}
+
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
     match msg {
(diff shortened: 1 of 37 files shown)


contracts/distribution/dao-rewards-distributor/src/contract.rs
```diff
@@ -594,6 +597,27 @@ fn execute_update_owner(
     Ok(Response::new().add_attributes(ownership.into_attributes()))
 }
 
+fn execute_unsafe_force_withdraw(
```
bekauz, line 600: regarding this force withdraw, given that the funds are in the contract already, couldn't we call `execute_fund` to re-account for them?

contracts/distribution/dao-rewards-distributor/src/testing/tests.rs
```diff
@@ -2659,6 +2659,43 @@ fn test_large_stake_before_claim() {
     suite.claim_rewards(MEMBER3, 1);
 }
 
+#[test]
+fn test_stake_during_interval() {
```
bekauz, line 2663: this feels like it covers it but maybe we should add some test cases for when the vp change event happens on the "edges" of interval? e.g. on the final block and such?

contracts/distribution/dao-rewards-distributor/src/helpers.rs
```diff
@@ -130,13 +130,13 @@ impl DurationExt for Duration {
         }
     }
 
-    fn checked_div(&self, denominator: &Self) -> Result<Uint128, ContractError> {
+    fn ratio(&self, denominator: &Self) -> Result<Decimal, ContractError> {
```
bekauz, line 133: i like the idea of using decimals here. took a sec to understand what the return value is but lgtm

bekauz (commented):
overall makes sense to me.

perhaps one thing that feels a bit extreme is the force withdraw. have you thought of possible alternatives to withdrawing? maybe some kind of re-accounting method would suffice?
main reason why I feel that way is just how the distributor contract is in a bit of a detached state after the force withdrawal. unless I'm missing something, wouldn't the queries return stale data?

contracts/distribution/dao-rewards-distributor/src/testing/tests.rs
noahsaso: yeah good idea

contracts/distribution/dao-rewards-distributor/src/contract.rs
noahsaso: yeah so unfortunately, because we don't store claimed and pending rewards anywhere, it gets quite complicated to piece back together the missed rewards. we store rewards in the 'per-unit-voting-power' units, meaning that we'd need to take into account all total voting power changes since the stream began in order to compute what the `total_earned_puvp` *should* be AND what it translates to in terms of absolute reward tokens. this is further made more complex by the fact that distributions may have transitioned between different emission rates. there's no automatic way to just "fix" the existing distributions.

to your point in the review comment, the distributor is now already in a detached state because the state does not correspond with the amount of rewards in the contract. the distributor thinks rewards have been distributed that have not been, leaving excess funds in the contract that should have been distributed but were not. so right now, `funded_amount` is correct, but `ends_at` is incorrect, since it was computed assuming `funded_amount` is fully distributed. however, a distribution will distribute less than `funded_amount`, with some left over.

i think the ideal way to remedy this is:
- pause all distributions, which resets the dates
- withdraw all undistributed funds from the distributions, which resets `funded_amount` to 0
- upgrade the contract with the decimal fix that *does* distribute partial-interval amounts correctly
- calculate the missed rewards by looking at what's left over in the contract and then looping through everyone's pending rewards and subtracting those from the left over amounts. force withdraw that unaccounted for amount, leaving all pending rewards that were successfully distributed intact
- restart the distributions and fund them again

this way the state should be reset to just account for the pending rewards that were distributed, and you can manually withdraw any funds beyond that amount before restarting and refunding.

contracts/distribution/dao-rewards-distributor/src/testing/tests.rs
noahsaso: done!

contracts/distribution/dao-rewards-distributor/src/contract.rs
bekauz: yeah, that's fair. the hard reset logic you describe with pause -> withdraw -> restart & fund is pretty much what I had in mind as well!

bekauz (approved):

PR #893, opened 2024-10-31
remove cw721 claim limit, and upgrade to v2.6.0

noahsaso:
this rewrites the NFT claims logic to make it paginate-able, removing the limit on how many outstanding claims can exist simultaneously

it is backwards compatible by introducing a new state map, seamlessly integrating with the legacy NFT claims so that DAOs can migrate without hassle

specific changes:
- all new claims go through the new claims logic, which stores claims in a `Map<(address, token_id) -> release_at>` that can be infinitely paginated, thereby removing any limit on outstanding claims 
- when claiming, you can either pass no argument, an empty `token_ids` list, or a specific set of `token_ids`. no argument will attempt to claim all outstanding legacy claims. an empty `token_ids` list attempts to claim all non-legacy claims, and it will simply fail if there are too many for one transaction. and a specific set of `token_ids` will attempt to claim just those outstanding non-legacy claims. ideally, we are always specifying the exact tokens to claim, and the frontend can automatically batch as necessary
- querying claims is now paginated, and all legacy claims are included at the beginning every time. the query's pagination (`start_after` and `limit`) only apply to the non-legacy claims

### DIFF
--- a/contracts/voting/dao-voting-cw721-staked/src/contract.rs
+++ b/contracts/voting/dao-voting-cw721-staked/src/contract.rs
@@ -17,7 +17,10 @@ use dao_voting::threshold::{
     ActiveThresholdResponse,
 };
 
-use crate::msg::{ClaimType, ExecuteMsg, InstantiateMsg, MigrateMsg, NftContract, QueryMsg};
+use crate::msg::{
+    ClaimType, ExecuteMsg, InstantiateMsg, MigrateMsg, NftClaim, NftClaimsResponse, NftContract,
+    QueryMsg,
+};
 use crate::state::{
     register_staked_nft, register_unstaked_nfts, Config, ACTIVE_THRESHOLD, CONFIG, DAO, HOOKS,
     INITIAL_NFTS, LEGACY_NFT_CLAIMS, NFT_BALANCES, NFT_CLAIMS, STAKED_NFTS_PER_OWNER,
@@ -348,7 +351,6 @@ pub fn execute_claim_nfts(
         ClaimType::All => {
             let token_ids = NFT_CLAIMS
                 .query_claims(deps.as_ref(), &info.sender, None, None)?
-                .nft_claims
                 .into_iter()
                 .filter(|nft| nft.release_at.is_expired(&env.block))
                 .map(|nft| nft.token_id)
@@ -648,18 +650,28 @@ pub fn query_nft_claims(
         .query_claims(deps, &addr)?
         .nft_claims
         .into_iter()
-        .map(|c| cw721_controllers::NftClaim::new(c.token_id, c.release_at))
+        .map(|c| NftClaim {
+            token_id: c.token_id,
+            release_at: c.release_at,
+            legacy: true,
+        })
         .collect::<Vec<_>>();
 
     // paginate all new claims
     let claims = NFT_CLAIMS
         .query_claims(deps, &addr, start_after.as_ref(), limit)?
-        .nft_claims;
+        .into_iter()
+        .map(|c| NftClaim {
+            token_id: c.token_id,
+            release_at: c.release_at,
+            legacy: false,
+        })
+        .collect::<Vec<_>>();
 
     // combine legacy and new claims
     let nft_claims = legacy_claims.into_iter().chain(claims).collect();
 
-    to_json_binary(&cw721_controllers::NftClaimsResponse { nft_claims })
+    to_json_binary(&NftClaimsResponse { nft_claims })
 }
 
 pub fn query_hooks(deps: Deps) -> StdResult<Binary> {
--- a/contracts/voting/dao-voting-cw721-staked/src/contract.rs
+++ b/contracts/voting/dao-voting-cw721-staked/src/contract.rs
@@ -17,7 +17,7 @@ use dao_voting::threshold::{
     ActiveThresholdResponse,
 };
 
-use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, NftContract, QueryMsg};
+use crate::msg::{ClaimType, ExecuteMsg, InstantiateMsg, MigrateMsg, NftContract, QueryMsg};
 use crate::state::{
     register_staked_nft, register_unstaked_nfts, Config, ACTIVE_THRESHOLD, CONFIG, DAO, HOOKS,
     INITIAL_NFTS, LEGACY_NFT_CLAIMS, NFT_BALANCES, NFT_CLAIMS, STAKED_NFTS_PER_OWNER,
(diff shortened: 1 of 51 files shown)


packages/cw721-controllers/src/nft_claim.rs
```diff
@@ -1,7 +1,20 @@
...
+#[derive(Error, Debug, PartialEq)]
+pub enum NftClaimError {
+    #[error(transparent)]
+    Std(#[from] StdError),
+
+    #[error("NFT claim not found for {token_id}")]
+    NotFound { token_id: String },
+
+    #[error("One or more NFTs is not ready to be claimed")]
+    NotReady {},
```
ismellike: It'd be useful to include which NFTs are not ready to be claimed in the error.

packages/cw721-controllers/src/nft_claim.rs
```diff
@@ -46,51 +59,70 @@ impl<'a> NftClaims<'a> {
...
-        let nft_claims = self.0.may_load(deps.storage, address)?.unwrap_or_default();
+        let limit = limit.map(|l| l as usize).unwrap_or(usize::MAX);
+        let start = start_after.map(Bound::<&String>::exclusive);
+
+        let nft_claims = self
+            .0
+            .prefix(address)
+            .range(deps.storage, start, None, Order::Ascending)
+            .take(limit)
+            .map(|item| item.map(|(token_id, v)| NftClaim::new(token_id, v)))
```
ismellike: Would rename v to released for clarity here.

packages/cw721-controllers/src/nft_claim.rs
```diff
@@ -46,51 +59,70 @@ impl<'a> NftClaims<'a> {
...
+                    }),
+                    Err(e) => Err(e.into()),
+                }
+            })
+            .collect::<Result<Vec<_>, NftClaimError>>()?;
+
+        // remove all now that we've confirmed they're mature
+        token_ids
+            .iter()
+            .for_each(|token_id| self.0.remove(storage, (addr, token_id)));
```
ismellike: Can probably just move this call into the map and have 1 iteration. State should get rolled back on errors

contracts/voting/dao-voting-cw721-staked/src/contract.rs
```diff
@@ -343,22 +337,57 @@ pub fn execute_claim_nfts(
...
+
+            if legacy_nfts.is_empty() {
+                return Err(ContractError::NothingToClaim {});
+            }
+
+            legacy_nfts
+        }
+        // attempt to claim non-legacy NFTs
+        Some(token_ids) => {
+            let token_ids = if token_ids.is_empty() {
```
ismellike: An enum would make more sense here tbh

```rust
pub enum ClaimType {
    All {},
    Select { token_ids: Vec<String> }
}
```

contracts/voting/dao-voting-cw721-staked/src/contract.rs
```diff
@@ -343,22 +337,57 @@ pub fn execute_claim_nfts(
...
+    token_ids: Option<Vec<String>>,
 ) -> Result<Response, ContractError> {
-    let nfts = NFT_CLAIMS.claim_nfts(deps.storage, &info.sender, &env.block)?;
-    if nfts.is_empty() {
-        return Err(ContractError::NothingToClaim {});
-    }
+    let token_ids = match token_ids {
+        // attempt to claim all legacy NFTs
+        None => {
+            // claim all legacy NFTs
```
ismellike: We should probably have a check here that ensures we don't have a claim in the new system.

contracts/voting/dao-voting-cw721-staked/src/contract.rs
```diff
@@ -343,22 +337,57 @@ pub fn execute_claim_nfts(
...
+            if legacy_nfts.is_empty() {
+                return Err(ContractError::NothingToClaim {});
+            }
+
+            legacy_nfts
+        }
+        // attempt to claim non-legacy NFTs
+        Some(token_ids) => {
+            let token_ids = if token_ids.is_empty() {
+                // query all NFT claims if none are provided
```
ismellike: Also have a check in here to ensure that we don't have a claim in the legacy system.

ismellike (changes requested):
Looks good!

I think it's just missing checks to ensure we don't have claims in both systems. Other suggestions are mainly for clarity and better API.

contracts/voting/dao-voting-cw721-staked/src/contract.rs
noahsaso: agreed. i made this decision so that it would be somewhat backwards compatible, but that doesn't serve much purpose here in the message. this really should be:

```rust
pub enum ClaimType {
    Legacy {},
    All {},
    Select { token_ids: Vec<String> }
}
```

contracts/voting/dao-voting-cw721-staked/src/contract.rs
noahsaso: ah, the goal of this fix is to be backwards compatible so DAOs can seamlessly migrate. both the legacy and non-legacy claim systems will exist simultaneously, until all legacy claims expire. at that point, we could remove all the legacy logic, but for now we need to support claiming both, even when others may exist.

contracts/voting/dao-voting-cw721-staked/src/contract.rs
noahsaso: same as above

packages/cw721-controllers/src/nft_claim.rs
noahsaso: ideally, yeah. but we error on the first one that is not ready, so we don't necessarily find all of them. i'll make it include the one it errors on first.

contracts/voting/dao-voting-cw721-staked/src/contract.rs
ismellike: Right but the actual message type could be `Option<ClaimType>` to keep that backward compatibility

contracts/voting/dao-voting-cw721-staked/src/contract.rs
ismellike: This logic I described should actually be in unstake, but it's only using the new system there now. Looks good here then and makes sense

contracts/voting/dao-voting-cw721-staked/src/contract.rs
noahsaso: ohhh i see, nice. well now i'm convinced that we should not make this message backwards compatible, since the behavior of the empty argument is now changing. we should force the caller to know about the behavior, as the UI needs to detect the contract version to make sure it's claiming the right stuff.

ismellike (approved):
🔥

contracts/voting/dao-voting-cw721-staked/src/contract.rs
ismellike: In this case, maybe we can attempt to claim from both legacy and current system? Then we can just have All and Select variants.
Behavior for All: 
If the error of both is EmptyClaims, then return empty claims error. If only one, then continue with claims for populated version. If other error, then error. If both not error, then both claimed.

noahsaso: i believe legacy claiming still needs to be separate. in the case that the combination of all legacy and non-legacy claims is too large to load at once, then legacy claims will still be blocked.

even if the user selectively claims all non-legacy claims that are outstanding, there may still be non-legacy claims going through the unstaking period whose expirations have to be loaded and checked.

PR #894, opened 2024-10-31
fix v2.5.0 publishing

noahsaso:
this inserts a commit right after the current v2.5.0 tag in the past that removes the v2.4.1 dependencies to prevent name collisions when publishing.

dao-pre-propose-base v2.5.0 currently can't be published because it depends on cw-denom v2.4.1 (which is correctly aliased) and dao-pre-propose-base v2.4.1, which depends on cw-denom v2.4.1 (which is *not* aliased since at the time it was the latest version of the package).

it seems that we cannot have packages depend on a past version of themself in normal dependencies—dev dependencies during testing are fine because we use `cargo hack publish` to strip those before publishing. this is because an old version of the same contract will necessarily use some of the same dependencies as the current version, so there will likely be a name overlap with mismatched versions.

this is caused by state migrations where we want to conveniently access old types in order to transform them. unfortunately, it seems like that's not a viable pattern, and instead we should just copy the few structs, enums, and state keys that are relevant for the migration into the migration function itself. this commit that i inserted (d453222e159d537531e5aa703a0b705afe4422d6) does just that.


noahsaso:
this PR serves to document what happened. since this is rewriting history (inserting a commit in the past), i have to force push the development branch and can't simply merge this PR.

ideally, we never do this. unfortunately, i published tag v2.5.0 before attempting to publish all the packages, and then kept on developing new features and merging them, so this is my fault. this should fix it!

PR #903, opened 2025-02-17
Initial dao actions on instantiate

### LINKED ISSUE #902: Execute messages on instantiation
It would be nice for a DAO to be able to execute arbitrary messages on creation to assist with setup. Things like creating cross-chain accounts, setting up vote delegation, etc. Initial items can be set on instantiation, which is currently how we enable a lot of modules via the frontend, but some modules require even more setup and require messages to be executed by the DAO's address.

Crucially, these messages would not be subject to governance as they would be set by the creator themself. This is potentially dangerous, as someone could create a DAO and immediately authz grant themselves the authority to spend on behalf of the DAO, giving them access to future funds and bypassing governance. However, the DAO creator is already trusted since they can add only themself as a member, execute some messages, and then add others. Thus, to support this functionality, we should just ensure that the UI makes it clear what messages were executed on instantiation, potentially showing them just like any other proposal in the list.

ismellike:
closes #902 

Adds an initial_dao_actions param to allow DAO's to execute actions on instantiate. These are stored in state and can be queried through `QueryMsg::InitialDaoActions`.

+ clippy fixes


noahsaso (approved):
thank you!!

my delegations branch is going to become the v2.7.0 branch and has all of the CI fixes that you also implemented, so i merged it

PR #910, opened 2025-03-30
vote delegations review

bekauz:
# DAO Vote delegation followups

This pr is a part of an extensive review of #888 and mostly contains cosmetic changes and small optimizations.

The summary of proposed changes is as follows:
 
### Instantiation

Instantiate message `no_sync_proposal_modules` field was changed into `sync_proposal_modules` in order to avoid the double-negative condition (`if !msg.no_sync_proposal_modules.unwrap_or(false)`). This is the only proposed api change.

Instantiation now also asserts that no funds were sent along with the message to be consistent with the execution handlers - otherwise, `execute_sync_proposal_modules` would be checked to be nonpayable via executemsg, but not via instantiation.

### `execute_delegate`

Makes use of a new helper `ensure_max_delegations_not_reached` to check for max_delegations limit.
Moves some calls that apply to both new/existing delegations to be outside the if/else check.

### `execute_undelegate`

Updates the percentage via `.update` instead of loading and saving in separate calls.

### `execute_stake_changed` + `execute_nft_stake_changed`

Cleans up the match statements.

### `handle_voting_power_changed_hook`

Removes the `get_voting_power` helper which internally loads the `DAO` address and uses the `dao_voting::voting::get_voting_power` directly to avoid loading `DAO` twice.
It was quite an involved hook handler so it now branches into either `handle_delegator_voting_power_changed_hook` or `handle_delegate_voting_power_changed_hook` depending on `is_delegate_registered` response.

### `handle_delegator_voting_power_changed_hook`

Probably the only significant logical change in this pr so deserves some attention.
Previously delegator voting power changed handler did two things: remove original delegated VP if nonzero and add new delegated VP if nonzero. Internally, this logic could result in up to four storage updates to the `DELEGATED_VP` wormhole kv-store.

The proposal here is that now, instead of clearing the existing state and writing a new one, handler does the following (for the next block wormhole entry):
1. calculate the delegated vp delta
2. if new vp is less than old vp, decrement the vp by the delta
3. if new vp is greater than old vp, increment the vp by the delta
4. if vp is unchanged, noop

After that:
1. if original delegation had voting power and an expiration date, that entry is undone
2. if the new delegation has voting power & global config specifies a delegation validity duration, the expiration entry is created

This results in up to three storage updates in the worst case while achieving the same functionality.

### Testing

Adds`/suite/cw721.rs` testing suite to enable cw721 staking dao testing.
Adds more unit tests for unauthorized hook callers and cw721 hook handlers.
Under dao-prop-single, adds vote-delegation related unit tests.

### DIFF
--- a/contracts/delegation/dao-vote-delegation/src/contract.rs
+++ b/contracts/delegation/dao-vote-delegation/src/contract.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{ensure, Addr, Order};
+use cosmwasm_std::{ensure, Addr, Order, Uint128};
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::{
     entry_point, to_json_binary, Binary, Decimal, Deps, DepsMut, Env, MessageInfo, Response,
@@ -520,13 +520,27 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
             delegate,
             proposal_module,
             proposal_id,
-            height,
+            proposal_height,
         } => Ok(to_json_binary(&query_unvoted_delegated_vp(
             deps,
             delegate,
             proposal_module,
             proposal_id,
-            height,
+            proposal_height,
+        )?)?),
+        QueryMsg::LessVotePowerWithoutDelegation {
+            proposal_module,
+            proposal_id,
+            proposal_height,
+            delegate,
+            delegated_vp,
+        } => Ok(to_json_binary(&query_less_vote_power_without_delegation(
+            deps,
+            proposal_module,
+            proposal_id,
+            proposal_height,
+            delegate,
+            delegated_vp,
         )?)?),
         QueryMsg::ProposalModules { start_after, limit } => Ok(to_json_binary(
             &query_proposal_modules(deps, start_after, limit)?,
@@ -655,6 +669,61 @@ fn query_unvoted_delegated_vp(
     }
 
     Ok(UnvotedDelegatedVotingPowerResponse { total, effective })
+}
+
+fn query_less_vote_power_without_delegation(
+    deps: Deps,
+    proposal_module: String,
+    proposal_id: u64,
+    proposal_height: u64,
+    delegate: String,
+    delegated_vp: Uint128,
+) -> StdResult<Uint128> {
+    let udvp = query_unvoted_delegated_vp(
+        deps,
+        delegate,
+        proposal_module,
+        proposal_id,
+        proposal_height,
+    )?;
+
+    // compute the new effective UDVP after this voter's delegated VP is
+    // removed, respecting the configured cap. subtract the voter's delegated VP
+    // from the delegate's total UDVP, and cap the result at the delegate's
+    // effective UDVP, to ensure we properly take into account the configured VP
+    // cap (the effective UDVP is the total UDVP with the cap applied, so the
+    // effective UDVP can be used in place of the cap in this computation).
+    let new_effective_udvp = udvp.total.checked_sub(delegated_vp)?.min(udvp.effective);
+
+    // compute the amount of UDVP the delegate will lose due to this voter's
+    // delegated VP being removed (likely due to a vote override).
+    // new_effective_udvp is capped at udvp.effective, so this will never
+    // underflow, but use saturating_sub for vibes anyway.
+    //
+    // the delegate will lose none, part, or all of this voter's delegated VP
+    // based on how the delegate's total UDVP and voter's delegated VP compare
+    // to the configured cap:
+    //
+    // 1. if the delegate's total UDVP is less than or equal to the cap, the
+    //    delegate will lose all of this voter's delegated VP since the cap is
+    //    not exceeded.
+    //
+    //       IF total_udvp <= cap, THEN loss = voter_delegated
+    //
+    // 2. if the delegate's total UDVP is greater than the cap by a margin less
+    //    than this voter's delegated VP, the delegate will lose part of this
+    //    voter's delegated VP. specifically: the difference between the cap and
+    //    the delegate's total UDVP without the voter's delegated VP.
+    //
+    //       IF total_udvp - voter_delegated < cap AND total_udvp > cap, THEN
+    //       loss = cap - (total_udvp - voter_delegated)
+    //
+    // 3. if the delegate's total UDVP is greater than the cap by a margin
+    //    greater than or equal to this voter's delegated VP, the delegate will
+    //    not lose any VP since the cap is already low enough.
+    //
+    //       IF total_udvp - voter_delegated >= cap, THEN loss = 0
+    Ok(udvp.effective.saturating_sub(new_effective_udvp))
 }
 
 fn query_proposal_modules(
--- a/contracts/delegation/dao-vote-delegation/src/contract.rs
+++ b/contracts/delegation/dao-vote-delegation/src/contract.rs
@@ -147,7 +147,7 @@ pub fn execute(
         ExecuteMsg::MemberChangedHook(msg) => {
             execute_membership_changed(deps, env, info.sender, msg)
         }
-        ExecuteMsg::VoteHook(vote_hook) => execute_vote_hook(deps, env, info.sender, vote_hook),
+        ExecuteMsg::VoteHook(vote_hook) => execute_vote_hook(deps, info.sender, vote_hook),
     }
 }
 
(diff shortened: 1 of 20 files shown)


contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -304,29 +288,26 @@ fn execute_delegate(
...
+            attempt: new_total_percent
+                .checked_mul(Decimal::percent(10_000))?
                 .to_string(),
         });
     }
 
-    PERCENT_DELEGATED.save(deps.storage, &delegator, &new_total_percent_delegated)?;
+    // final state updates applicable to both new and existing delegations
+    DELEGATION_ENTRIES.save(deps.storage, (&delegator, &delegate), &new_entry)?;
+    PERCENT_DELEGATED.save(deps.storage, &delegator, &new_total_percent)?;
```
NoahSaso, line 310: all the changes to execute_delegate look great 🙏

noahsaso, line 314: all the changes to execute_delegate look great 🙏

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -28,10 +26,7 @@ pub(crate) fn execute_stake_changed(
     }
 
     match msg {
-        StakeChangedHookMsg::Stake { addr, .. } => {
-            handle_voting_power_changed_hook(deps, &env, addr)
-        }
-        StakeChangedHookMsg::Unstake { addr, .. } => {
+        StakeChangedHookMsg::Stake { addr, .. } | StakeChangedHookMsg::Unstake { addr, .. } => {
```
noahsaso: slay

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -69,97 +64,153 @@ pub(crate) fn execute_nft_stake_changed(
     }
 
     match msg {
-        NftStakeChangedHookMsg::Stake { addr, .. } => {
-            handle_voting_power_changed_hook(deps, &env, addr)
-        }
-        NftStakeChangedHookMsg::Unstake { addr, .. } => {
+        NftStakeChangedHookMsg::Stake { addr, .. }
+        | NftStakeChangedHookMsg::Unstake { addr, .. } => {
```
noahsaso: slay

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -178,45 +229,64 @@ pub fn execute_vote_hook(
...
+
+        UNVOTED_DELEGATED_VP.save(
+            deps.storage,
+            (&delegate, proposal_module, proposal_id),
+            &new_udvp,
+        )?;
+    }
+
+    Ok(())
+}
```
noahsaso: i'm gonna move this code back into the other function — i don't think it's long enough to need to make a new function

also, the `height` from the `NewVote` struct is actually the proposal start height, and that's the height that needs to be used for both the `DELEGATIONS.load_all` and `get_udvp` calls. you changed it to `env.block.height`, which is wrong. but you inadvertently found a bug because i was using `env.block.height for `DELEGATIONS.load_all` when i should've been using the proposal height, so nice find!! 😁

contracts/delegation/dao-vote-delegation/src/hooks.rs
```diff
@@ -69,97 +64,153 @@ pub(crate) fn execute_nft_stake_changed(
...
+            if let Some(expire_in) = expiration {
+                DELEGATED_VP.increment(
+                    deps.storage,
+                    delegate.clone(),
+                    expire_in,
                     current_delegated_vp,
-                    expiration,
                 )?;
             }
+        }
```
noahsaso: i really like the direction of the all changes made to this hook, only incrementing/decrementing the delta instead of fully removing and re-adding. however i _think_ there's a bug here. i'm going to keep looking at it, but with Wormhole, increments and decrements change at the height specified _and_ all future changes, so i'm pretty confident the `increment` here that is undoing the previous expiration needs to take into account the delta as well

noahsaso (changes requested):
thank you so much for the detailed review 🥵 i'm making some small tweaks / clean up of my own, but i think it's almost ready

contracts/delegation/dao-vote-delegation/src/hooks.rs
noahsaso: ok yeah, there's an underflow risk. i fixed it by changing the order from:

- update delta
- undo original expiration
- set new expiration

to:
- undo original expiration
- update delta
- set new expiration

this effectively preserves the order of operations from the original implementation with full remove and then full add, and then combines the updates in the middle into a single delta update

contracts/delegation/dao-vote-delegation/src/hooks.rs
bekauz: I totally missed that `NewVote` was returning the prop start height! great catch, and your fix makes sense to me

contracts/delegation/dao-vote-delegation/src/hooks.rs
bekauz: yeah great call. even beyond the underflow risk the new order you propose (undo original expiration -> update delta -> set new expiration) is a lot cleaner and easier to follow.

bekauz:
@NoahSaso I pushed a few changes (contained to the testing module) after playing around with the config updates and their effects on existing delegations. Other than that I think this looks good!

noahsaso:
@bekauz everything looks good to me! the last thing is i pushed d53740dfe9d5407e2b5b60977bc655d25c6e84a3 and 45390cd82ee4f68539bab72309f3643611a3ccaf which moves math (that calculates how much unvoted delegated voting power (UDVP) to remove from a delegate for a given delegator vote on a given proposal) from the proposal module vote execution to a query on the delegation module. i did this to preserve separation of concerns, because the math has to know about and apply the configured delegate voting power cap, and the proposal module shouldn't really have to interact with that logic. this shouldn't increase gas costs at all because it just replaces the `UnvotedDelegatedVotingPower` query the proposal module was performing with a `DelegatedVotePowerReduction` query, which internally accesses the UDVP and performs the capping math as well. this should be cleaner and preserve all abstraction boundaries!

effectively, when a delegator votes, the proposal module loops over everyone they've delegated to, checks if each delegate has voted, and removes the delegator's delegated VP from each delegate's vote tally in order to override their vote. the amount to remove for each delegate depends on the total delegated VP they've voted with, the cap that is applied, and the VP they've been delegated by the delegator. if the delegate has surpassed the cap by more than the delegator's VP, there is no reduction in vote tally, since reducing the delegate's total delegated VP still surpasses the cap. and if the delegate is below the cap or just above the cap by a small enough margin, some or all of the delegated VP needs to be removed from the delegate's vote tally. now, the proposal module code is very clean because it simply asks the delegation module to compute the reduction to apply for the current vote and delegated VP, so all the cap logic lives in the delegation module.

packages/dao-voting/src/delegation.rs
```diff
@@ -232,59 +249,30 @@ pub fn handle_delegate_vote_override<Vote: Serialize + DeserializeOwned>(
...
+                let reduction: Uint128 = deps.querier.query_wasm_smart(
+                    delegation_module,
+                    &QueryMsg::DelegatedVotePowerReduction {
+                        proposal_module: proposal_module.to_string(),
+                        proposal_id,
+                        proposal_height,
+                        delegate: delegate.to_string(),
+                        delegated_vp,
+                    },
+                )?;
```
bekauz: nice!

packages/dao-voting/src/delegation.rs
```diff
@@ -41,13 +41,30 @@ pub enum QueryMsg {
...
+    /// Returns the VP that should be removed from a delegate's effective UDVP
+    /// (and vote tally if already voted) on a specific proposal when a
+    /// delegator casts a vote (potentially overriding the delegate's vote). The
+    /// `proposal_height` field is the height at which the proposal was created.
+    /// The `delegated_vp` field is the amount of VP delegated by the delegator
+    /// to the delegate for this proposal. This query takes into account the
+    /// configured VP cap and should be used by proposal modules when a
+    /// delegator overrides a delegate's vote to compute ballot VP updates.
+    #[returns(Uint128)]
+    DelegatedVotePowerReduction {
```
bekauz: definitely feels like the right place for this, sweet

bekauz (commented):
lgtm! great catch

contracts/delegation/dao-vote-delegation/src/contract.rs
```diff
@@ -528,20 +528,22 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
             proposal_id,
             proposal_height,
         )?)?),
-        QueryMsg::DelegatedVotePowerReduction {
+        QueryMsg::EffectiveUnvotedDelegatedVotingPowerReduction {
```
bekauz, line 531: fire, actually love how explicit it is

noahsaso: me too

PR #913, opened 2025-06-19
Role-based Authorization Module (RBAM)

noahsaso:
# dao-rbam

[dao-rbam on crates.io](https://crates.io/crates/dao-rbam)
[docs.rs](https://docs.rs/dao-rbam/latest/dao_rbam/)

Role-based Authorization Module (RBAM) that lets DAOs configure
organization-wide roles and authorization policies.

## Filtering

Authorizations for a role are defined by a
[`cw-jsonfilter`](https://github.com/DA0-DA0/dao-contracts/tree/development/packages/cw-jsonfilter) filter. The JSON filter
format, operators, and transformations are described in the [filter
documentation](https://github.com/DA0-DA0/dao-contracts/blob/development/packages/cw-jsonfilter/docs/filter.md).

You can make very complex filters by combining the various operators and
transformations.

For example:

```json
{
  "bank": {
    "send": {
      "to_address": "cosmos1...",
      "amount": [
        {
          "denom": "uatom",
          "amount": { "#to_number": { "$between": [1000, 2000] } }
        }
      ]
    }
  }
}
```

This filter would allow the role-holder to send between 1,000 and 2,000 uatom to
the `cosmos1...` address.

Note the `#to_number` transformation, which converts the string `"1000"` to the
number `1000` before the comparison operator is applied, since bank messages use
strings for amounts. Since strings can be compared lexicographically, strings
and numbers are both valid inputs to comparison operators, and thus strings are
not coerced to numbers automatically and require manual conversion like this.

```json
{
  "bank": {
    "send": {
      "to_address": {
        "$or": ["cosmos123...", "cosmos145..."]
      },
      "amount": [
        { "denom": "uatom", "amount": { "#to_number": { "$lte": 10000 } } }
      ]
    }
  }
}
```

This filter would allow the role-holder to send <= 10,000 uatom to either
`cosmos123...` or `cosmos145...`.

### Base64

CosmWasm messages encode binary data as base64-encoded strings, which can be
filtered with the `#base64` transformation. To authorize contract instantiations
or executions, for example, you first decode the base64-encoded message into a
JSON object, and then filter on the JSON object, like so:

```json
{
  "wasm": {
    "execute": {
      "contract_addr": "exact_contract",
      "msg": {
        "#base64": {
          "update_config": {
            "config": {
              "name": { "$contains": "new_name" }
            }
          }
        }
      },
      "funds": []
    }
  }
}
```

This filter would allow the role-holder to execute the `exact_contract` contract
with the `update_config` message, ONLY if the `name` field contains `new_name`.
Other fields can be anything, since no filter is specified for them. This also
ensures that no funds are sent with the message, since the `funds` field must be
an empty array.

### Stargate/Protobuf

You can use the `#stargate` transformation to filter on custom protobuf
messages. See the [protobuf
documentation](../../../packages/cw-jsonfilter/docs/protobuf.md) for more
details.

To enable protobuf decoding support, you must first register the protobuf types
with the contract, and then create authorizations that use the transformation.

1. Register the protobuf types with the contract by executing the
   `RegisterProtobufs` message. The `file_descriptor_sets` field is a list of
   file descriptor sets, each of which is encoded as bytes. The protobuf
   documentation linked above describes how to generate the file descriptor
   sets. The generated `.pb` file is the file descriptor set encoded as bytes.

   ```json
   {
     "register_protobufs": {
       "file_descriptor_sets": [
         [1, 2, 3] // file descriptor set 1
         [4, 5, 6] // file descriptor set 2
         [7, 8, 9] // file descriptor set 3
       ]
     }
   }
   ```

2. Create an authorization that uses the `#stargate` transformation. The
   `type_url` field is the full type URL of the protobuf message (prefixed with
   a `/` like normal), and the `value` field is the value to filter on. This
   decodes the base64-encoded protobuf message into a JSON object that can be
   filtered on.

   ```json
   {
     "create_authorization": {
       "role_id": 1,
       "name": "My Authorization",
       "filter": {
         "#stargate": {
           "type_url": "/some.custom.protobuf.message",
           "value": {
             "field1": "value1",
             "field2": "value2",
             "field3": {
               "$between": [0, 100]
             },
             "field4": {
               "#to_number": {
                 "$gte": 50
               }
             }
           }
         }
       }
     }
   }
   ```


bekauz:
First of all, this is super impressive. The capabilities of json filter alone are crazy, and adding the protobuf-registry into the mix is really cool.

It's quite a pr, so most of my feedback can be seen in #915 .

I'd say the most (probably only) significant change there was a slight refactor of the way `CwJsonFilter` is configured with a protobuf decoder. My rationale there was twofold:

1. to try to get rid of a boxed `dyn Fn` with lifetimes (for gas and code simplicity reasons)
2. add some type safety

The latter point is probably going to be easy to see across most of my suggested changes. Considering the capabilities of the json filter matching and how generic/abstract the matching engine is (by design), my thinking was that it's probably a good idea to "shield" the matching layer with as much type safety as possible.

Other than that, I tried to focus on making the code easier to review and audit in the future, added some comments/docs in places where I spent some time figuring out what's happening, and tried to save some gas wherever possible given the gas-intensive nature of the matching/decoding engine. 

I have a few more thoughts on some areas my pr didn't touch; I'll leave those here in a bit.

Again, awesome work! :heart:

contracts/external/cw-filter/src/decoder.rs
bekauz, line 1: wasn't sure if this should be in `cw_protobuf_registry` actually, wyt?

contracts/external/cw-protobuf-registry/src/contract.rs
```diff
@@ -0,0 +1,518 @@
...
+            // If file is already registered, merge the files.
+            Some(existing_fd_data) => {
+                merge_file_descriptors(store, existing_fd_data, file_package, file_name, new_fd)?
+            }
+            // otherwise, write a new file descriptor under the given file name
+            None => {
+                let file_descriptor_data = new_fd.encode_to_vec();
+                FILES.save(store, file_name, &file_descriptor_data)?;
+            }
+        }
```
bekauz, line 182: i was wondering if file descriptor merging is something we really need to deal with. i'm not really familiar with protoc artifact format, but can we potentially end up with a semantically invalid `FileDescriptorProto`?

contracts/external/cw-protobuf-registry/src/contract.rs
```diff
@@ -0,0 +1,518 @@
...
+            new_file_package: file_package,
+        });
+    }
+
+    // Add new or overwrite existing dependencies/messages/enums.
+    for dep in new_fd.dependency {
+        if !existing_fd.dependency.contains(&dep) {
+            existing_fd.dependency.push(dep);
+        }
+    }
```
bekauz, line 213: could `new_fd` be missing any of the `existing_fd` dependencies here? e.g. if during an update some enums/messages/other file descriptor primitives were removed. wondering if we can end up with some dead references as a result of such update and can that cause any issues.

contracts/external/cw-protobuf-registry/src/contract.rs
```diff
@@ -0,0 +1,518 @@
...
+        if let Some(i) = existing_fd
+            .enum_type
+            .iter()
+            .position(|e| e.name == new_enum.name)
+        {
+            existing_fd.enum_type[i] = new_enum; // replace
+        } else {
+            existing_fd.enum_type.push(new_enum); // insert
+        }
+    }
```
bekauz, line 237: what would happen if in this case?

`existing_fd` contains:

  - `EnumX` (v1)
  - `MessageA` that has a field of type `EnumX`

`new_fd` is being merged and contains:

- `EnumX` (v2)
- `MessageB` that has a field of type `EnumX` (v2)
- It does not contain `MessageA`


As far as I understand, during the enum merge we would encounter `EnumX` (name unchanged) and replace the v1 with v2. What's the state of `MessageA` now, is it referring to `EnumX` v2?

bekauz, line 237: also was looking at the `FileDescriptorProto` struct and there was a lot to it. maybe there's something we should be validating from the other fields? e.g. there's:

```rust
/// The syntax of the proto file.
/// The supported values are "proto2" and "proto3".
#[prost(string, optional, tag = "12")]
pub syntax: ::core::option::Option<::prost::alloc::string::String>,
```

could fds generated with different proto syntax being merged break anything?

contracts/external/cw-protobuf-registry/src/protobuf.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+/// Base64 decoding engine
+#[cfg(any(feature = "library", test))]
+pub const BASE64_ENGINE: GeneralPurpose = GeneralPurpose::new(
+    &alphabet::STANDARD,
+    GeneralPurposeConfig::new()
+        .with_decode_allow_trailing_bits(true)
+        .with_decode_padding_mode(DecodePaddingMode::Indifferent),
+);
```
bekauz, line 322: we have this defined in cw-jsonfilter as well, maybe we can just import from there?

packages/cw-jsonfilter/src/json_ops/op_comparison.rs
```diff
@@ -0,0 +1,441 @@
...
+                    // If the types are incompatible, fail.
+                    None => {
+                        return FilterResult::operator_failed(
+                            op_ctx.operator,
+                            "filter bounds and value are not all numbers or all strings",
+                            op_ctx.filter_path,
+                            op_ctx.obj_path,
+                        )
+                    }
+                };
```
bekauz: was thinking about rewriting it like this but felt too much like a nit haha

```rust
let inclusive = !op_ctx.operator.ends_with("_exclusive");

// if range check is inclusive of the edges, check them first
let in_range = if inclusive && (value == min || value == max) {
    true
} else {
    // otherwise validate the range
    let (is_gt_min, is_lt_max) = match (gt_json(value, min), lt_json(value, max)) {
        (Some(gt), Some(lt)) => (gt, lt),
        _ =>  return FilterResult::operator_failed(
            op_ctx.operator,
            "filter bounds and value are not all numbers or all strings",
            op_ctx.filter_path,
            op_ctx.obj_path,
        )
    };
    is_gt_min && is_lt_max
};

FilterResult::from_bool(
    in_range,
    op_ctx.operator,
    format!(
        "value ({value}) not between {} min ({min}) and max ({max})",
        match inclusive {
            true => "(inclusive)",
            false => "(exclusive)",
        },
    ),
    op_ctx.filter_path,
    op_ctx.obj_path,
)
```

bekauz (commented):
some more thoughts

contracts/external/cw-filter/src/decoder.rs
noahsaso: yeah it makes more sense there, and `cw-filter` depends on it as a library, so that should be fine

contracts/external/cw-protobuf-registry/src/contract.rs
noahsaso: a file descriptor is just a set of enums and messages. when we create the minimal file descriptor set for a particular message, we pull out all the necessary ones with their dependencies, so these are just used for storage and are not actually used directly. though i do not think they end up semantically invalid at all since the protoc/buf commands make sure to generate complete sets with all dependencies resolved.

i think there will be a point where all messages in a single file descriptor and all their dependencies exceed the max TX size limit, so they need to be able to submit parts of files that get added together in the contract. tho maybe a better way to do this is storing individual messages/enums in a tuple `(file_name, message_name)`? but then that's a storage read for every message instead of just once for the whole file—do you know which of those is more expensive or if it's counted the same?

mainly i think that if we don't support merging, people will accidentally clear past messages added that they still need, and i don't want them to have to figure out which messages are already registered in order to register more.

contracts/external/cw-protobuf-registry/src/contract.rs
noahsaso: yeah that's definitely possible. though it would be quite tricky (potentially prohibitively gas expensive) to backtrack all dependencies of a specific file. enums/messages/file descriptors can only be removed through `Unregister`, which there's really no need to do unless you're replacing it with a correct one, and if you're unregistering you probably know what you're doing.

the main time i see unregister being useful is if a file is too large in general, so you might need to do a `Register`, `Prepare`, and `Unregister` in a loop for various messages you care about.

the alternative is we store reverse dependencies and clear them when a file is removed, but then we also have to force them to unregister dependencies in order and remove any dependents first. which is a nightmare. so basically i don't think unregister will be used much at all, and if so, you'll be doing it as part of more operations 🤷

contracts/external/cw-protobuf-registry/src/contract.rs
noahsaso: a chain only has one valid decoding of a given enum or message type, so `MessageA` and `MessageB` can't reference different versions of enums of the same name—there is no such thing

noahsaso: maybe 🤷 but i'm quite confident that chains internally use a single proto syntax, since they're all managed in the SDK repo with the `buf` protobuf manager. no one will be manually specifying this stuff so i don't think we have to worry about syntax issues

contracts/external/cw-protobuf-registry/src/protobuf.rs
noahsaso: the `cw-protobuf-registry` doesn't currently depend on `cw-jsonfilter` and in theory they are entirely unrelated, so i don't see the point just for that. separation of concerns and all that

packages/cw-jsonfilter/src/json_ops/op_comparison.rs
noahsaso: no that's great, love it

noahsaso: done :)

contracts/external/cw-protobuf-registry/src/contract.rs
bekauz: right, the last paragraph makes sense. sounds good.

contracts/external/cw-protobuf-registry/src/contract.rs
bekauz: right, so if `EnumX` is extended with additional variants and that new enum is written to state, the `MessageA` reference to it will updated and valid, right?

contracts/external/cw-protobuf-registry/src/protobuf.rs
bekauz: fair enough, lets keep it as is

bekauz (approved):
lgtm! left one more question regarding the enum updates but approving in advance as those edge cases could be covered in readme/docs

contracts/external/cw-protobuf-registry/src/contract.rs
noahsaso: yes! though `MessageA` will need to be prepared again

PR #929, opened 2026-05-13
feat: dao-voting-juno-staked — historical-snapshot voting via x/votin…

JakeHartnell:
Replaces the live-query approach attempted in DA0-DA0/dao-contracts#832 with a thin consumer of Juno v30's x/voting-snapshot chain module. Voting power is now read at the exact requested height (at-or-before semantics enforced by the chain), with LST exclusion handled at the chain layer.

Architecture:
- Custom-query binding (JunoQuery) mirroring juno/wasmbindings/types/query.go exactly. Contract is Deps<JunoQuery> internally; external WasmQuery::Smart consumers don't need to know about the chain type.
- Standard dao-interface voting surface: VotingPowerAtHeight, TotalPowerAtHeight, Dao, Info.
- Sudo intake for x/cw-hooks staking events. AfterDelegationModified + BeforeDelegationRemoved drive a stake-delta computation (new power read from the chain snapshot, prev power read at current_height-1). Other variants are swallowed silently to keep cw-hooks registration alive.
- Hook fan-out via dao_hooks::stake::StakeChangedHookMsg::{Stake,Unstake} to a DAO-gated subscriber list (the standard cw-hooks::Hooks pattern), so the gauge orchestrator and rewards distributor wire up unchanged.

Out of scope for v1:
- auto_register_staking_hooks (would need a prost dep just for one proto message; registration happens out-of-band via the CLI). Field is kept for future use; passing Some(true) returns AutoRegisterNotYetSupported.
- VotingPowerOverRange consumer — surfaced in bindings.rs for downstream contracts that want plural-voting-style integration, but not used by the voting module itself.

Verification:
- 9/9 unit tests pass under nightly-2024-01-08 (the dao-contracts CI pin). Tests cover instantiate / DAO storage / hook add+remove auth / power proxy under at-or-before / stake-delta math / fan-out / silent swallowing of validator and slash events.
- cargo clippy --all-targets -- -D warnings clean.
- cargo fmt -- --check clean.
- wasm32-unknown-unknown release build clean (1.7 MB).

Refs: memory/juno-voting-design.md (option C settled), memory/v30-upgrade-plan.md (x/voting-snapshot in PR #1202), memory/hack-juno-plan-2026-05-12.md (this is the "dao-voting-juno-staked" item in the program's sequencing diagram).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2702,6 +2702,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "dao-voting-juno-staked"
+version = "2.8.0-alpha.2"
+dependencies = [
+ "anyhow",
+ "cosmwasm-schema",
+ "cosmwasm-std",
+ "cw-hooks 2.8.0-alpha.2",
+ "cw-multi-test",
+ "cw-storage-plus 1.2.0",
+ "cw2 1.1.2",
+ "dao-dao-macros 2.8.0-alpha.2",
+ "dao-hooks 2.8.0-alpha.2",
+ "dao-interface 2.8.0-alpha.2",
+ "serde",
+ "thiserror",
+]
+
+[[package]]
 name = "dao-voting-onft-staked"
 version = "2.8.0-alpha.2"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,6 +142,7 @@ dao-voting-cw4 = { path = "./contracts/voting/dao-voting-cw4", version = "2.8.0-
 dao-voting-cw721-roles = { path = "./contracts/voting/dao-voting-cw721-roles", version = "2.8.0-alpha.2" }
 dao-voting-cw721-staked = { path = "./contracts/voting/dao-voting-cw721-staked", version = "2.8.0-alpha.2" }
 dao-voting-onft-staked = { path = "./contracts/voting/dao-voting-onft-staked", version = "2.8.0-alpha.2" }
+dao-voting-juno-staked = { path = "./contracts/voting/dao-voting-juno-staked", version = "2.8.0-alpha.2" }
 dao-voting-token-staked = { path = "./contracts/voting/dao-voting-token-staked", version = "2.8.0-alpha.2" }
 nft-controllers = { path = "./packages/nft-controllers", version = "2.8.0-alpha.2" }
 
--- a/contracts/voting/dao-voting-juno-staked/.cargo/config
+++ b/contracts/voting/dao-voting-juno-staked/.cargo/config
@@ -1,0 +1,4 @@
+[alias]
+wasm = "build --release --target wasm32-unknown-unknown"
+unit-test = "test --lib"
+schema = "run --example schema"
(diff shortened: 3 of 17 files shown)


Dragonmonk111:
Substantive review pass on this — applying the same deterministic-scrutiny benchmark we use on the JunoClaw contracts. The full review is at [MOULTBOOK_REVIEW_OF_DAO_VOTING_JUNO_STAKED_PR_929.md](https://github.com/Dragonmonk111/junoclaw/blob/main/docs/MOULTBOOK_REVIEW_OF_DAO_VOTING_JUNO_STAKED_PR_929.md) for the long form (≈19K, 9 sections). Headline: **the architecture is correct.** The thin-consumer-of-`x/voting-snapshot` shape genuinely fixes the sparse-delegator bug from #832 at the architectural level — the chain owns the snapshot, the cw contract owns the routing.

Two findings I'd want to discuss before merge. Both have concrete worked examples in the long form.

---

### F1 (HIGH if confirmed) — Same-block multi-event accumulation

When two staking events for the same delegator land in the same block, the `prev_power = chain[h-1]` formula is too stale, and the emitted `StakeChangedHookMsg::{Stake, Unstake}` deltas are cumulative-against-block-start rather than incremental.

**Worked example.** Voter A enters block H with `chain[H-1] = 100`. Two `MsgDelegate` in one tx land sequentially:

| Event | Chain after | Sudo computes | Emits |
|---|---|---|---|
| 1: AfterDelegationModified (now 150) | `chain[H] = 150` | new=150, prev=100 → +50 | `Stake { 50 }` |
| 2: AfterDelegationModified (now 175) | `chain[H] = 175` | new=175, prev=100 → +75 | `Stake { 75 }` |

Subscribers see `+125` for voter A in block H; the actual stake change is `+75`. The duplicate is `+50`.

This is invisible to **cumulative-power-tracking** subscribers (they re-read voting power at hook-fire time and treat the emitted delta as a wake-up signal). It's a real correctness bug for **flow-tracking** subscribers (a stake-velocity indexer, or a rewards system that proportions to deltas instead of balances).

The 9 unit tests don't fire two events for the same delegator at the same height; the bug isn't in the test set.

**One fix shape:** track the contract's own emitted-power per delegator per block in a `Map<&Addr, EmittedPower { height, power }>`. On each event: if `last_emitted.height == current_height`, use `last_emitted.power` as `prev_power` instead of `chain[h-1]`. ~25 LoC + 1 test. Costs one read + one write per sudo.

---

### F2 (HIGH) — `reply` handler is empty but README promises auto-unregistration on hook failure

The README says:

> Subscribers register via `AddHook { addr }` (gated to the DAO) and are auto-unregistered if their execute call errors (standard `reply_on_error` pattern from the rest of the dao-contracts hook surface).

But `contract.rs:reply` is a no-op:

```rust
pub fn reply(_deps: DepsMut<JunoQuery>, _env: Env, _reply: Reply) -> Result<Response, ContractError> {
    // Reserved for future auto-unregistration on hook failure. The
    // current dao_hooks call sites use plain SubMsg::new (no reply
    // requested), so this entry-point is unreachable in practice.
    Ok(Response::new())
}

PR #932, opened 2026-09-01
fix(stake hooks): make hook receiver failures non-fatal to staking (audit Critical 3)

noahsaso:
Fixes audit Finding 3 (Critical): "One owner action can make staking and unstaking impossible for an entire DAO".

## Root cause

Stake hooks were dispatched by `dao-hooks` as plain submessages (`SubMsg::new`) with no reply, and none of the producing contracts had a reply handler for them. Any error returned by a hook receiver therefore aborted the producer's whole transaction. The rewards distributor correctly rejects hook calls from an address it no longer recognises (`InvalidHookSender`), so after a routine owner `Update` of a distribution's `hook_caller`, or in the ordinary gap between `AddHook` and `Create`, every stake and unstake in the DAO failed, staked principal could not be withdrawn, and voting power froze at the moment of the failure. Any registered stake hook receiver held the same veto.

The audit lists only `dao_hooks::stake` in its source, but the NFT stake hook builders in `dao_hooks::nft_stake` had the identical no-reply dispatch, and the distributor rejects unrecognised NFT stake hook callers the same way, so NFT-staked DAOs were exposed to exactly the same freeze. This PR fixes both.

## Fix

- `dao-hooks`: `stake_hook_msgs`, `unstake_hook_msgs`, `stake_nft_hook_msgs` and `unstake_nft_hook_msgs` keep their signatures but now dispatch with `SubMsg::reply_always`, tagging each submessage with `STAKE_HOOK_REPLY_ID_BASE`/`UNSTAKE_HOOK_REPLY_ID_BASE` plus the hook's index in the producer's registry. A shared `handle_stake_hook_reply` builds the reply response: success yields an empty response; failure yields attributes `action=stake_hook_failed`, `hook=stake|unstake`, `addr=<receiver>`, `error=<error>` so the failure is observable on-chain without failing the staking transaction and without removing the hook (the audit's recommended policy: neither fail-closed nor silently remove). Producers must handle that reply range and must not use reply IDs inside it.

  The `addr` attribute matters because CosmWasm chains redact submessage errors before they reach a reply handler (`redactError` in wasmd), so on chain `error` is a codespace and code rather than the receiver's message. Without the address a DAO with several hooks registered could not tell which receiver failed.
- `cw20-stake`: new `reply` entry point handling the stake/unstake reply IDs.
- `dao-voting-token-staked` and `dao-voting-cw721-staked`: new reply IDs handled in the existing `reply` entry point.
- `dao-voting-onft-staked`: new `reply` entry point.
- `dao-testing`: `cw20-stake` and `dao-voting-onft-staked` test wrappers now register their reply handlers.
- Docs updated in `dao-hooks` and `dao-rewards-distributor` READMEs.

Adding a `reply` entry point is a no-op for existing state; migrating deployed `cw20-stake` / `dao-voting-onft-staked` instances to the new code is sufficient.

## Tests

- `dao-rewards-distributor`: `stale_rewards_distributor_cannot_block_cw20_stake_or_unstake` reproduces the audit scenario (distribution moved to a new `hook_caller` while the old staking contract still lists the distributor as a hook). Stake and unstake succeed, the failure attributes are emitted, a second (healthy) hook still runs, and both hooks remain registered.
- `dao-voting-token-staked`: `test_staking_hooks_continue_after_failure` with one failing and one healthy hook.
- `dao-voting-cw721-staked` / `dao-voting-onft-staked`: `test_add_remove_hooks` updated to assert that staking and unstaking succeed with a non-contract hook registered, voting power updates, the failure is recorded, and the hook stays registered. Unit tests for the reply handlers.
- `cw20-stake`: unit tests for the reply handler.
- `dao-hooks`: unit tests asserting `reply_always` dispatch with index-tagged IDs and unchanged payloads, that the reply names the receiver that failed, that a receiver no longer in the registry reports `unknown`, and that a producer's own reply IDs fall through untouched.

The multitest suites assert that with a failing and a healthy hook registered together, the failure event names the failing one and not the healthy one.

Verified locally: `cargo test --workspace`, `cargo +nightly-2024-01-08 fmt --all -- --check`, `cargo +nightly-2024-01-08 clippy --all-targets -- -D warnings`.

## Deliberately not included

The audit independently recommends validating at `Create`/`Update` that the named `hook_caller` lists the distributor among its hooks. An earlier draft of this branch implemented that, but the DAO DAO UI's "Create reward distribution" proposal sends `Create` (and `FundLatest`) **before** `AddHook` on the voting module, so strict validation at `Create` would break every proposal built by the current UI and any third-party integration using the same order. With this PR a missing or stale hook registration can no longer block staking, so the validation is a hardening step rather than a security fix. It should be revisited together with a UI change (register the hook first, or validate on a later message). The README now documents the recommended ordering.

A per-hook `with_gas_limit` was considered and rejected. The dispatch it would replace, `SubMsg::new`, also set `gas_limit: None`, so gas exposure is unchanged by this PR; a gas-exhausting receiver requires a passed governance proposal to register, unlike the routine owner action in the finding; legitimate hook gas is unbounded by design (audit Findings 26 and 30 both conclude the threshold "cannot be established from the code alone"), so a limit set too low would silently stop reward accrual and delegated voting power updates; and wasmd only applies the limit while `*msg.GasLimit < gasRemaining`, so it would not close the hole anyway. The audit's own remedy for that residual is Finding 30's: bound the fan-out.

Related audit findings 4, 6 and 7 (the vote/proposal hook registries that silently remove a hook on error, and the index-based removal) are a separate framework-wide policy decision and are not addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YBSNZfoCq2XqymbVV9r5rE

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2387,6 +2387,7 @@ dependencies = [
  "cw721-base 0.18.0",
  "dao-hooks 2.8.0-alpha.2",
  "dao-interface 2.8.0-alpha.2",
+ "dao-proposal-hook-counter",
  "dao-rewards-distributor",
  "dao-testing",
  "dao-voting 2.8.0-alpha.2",
--- a/contracts/distribution/dao-rewards-distributor/Cargo.toml
+++ b/contracts/distribution/dao-rewards-distributor/Cargo.toml
@@ -34,6 +34,7 @@ thiserror = { workspace = true }
 
 [dev-dependencies]
 dao-rewards-distributor = { workspace = true }
+dao-proposal-hook-counter = { workspace = true }
 cw-multi-test = { workspace = true }
 anyhow = { workspace = true }
 cw20-stake = { workspace = true, features = ["library"] }
--- a/contracts/distribution/dao-rewards-distributor/README.md
+++ b/contracts/distribution/dao-rewards-distributor/README.md
@@ -36,12 +36,14 @@ voting power changes, such as:
 - `dao-voting-token-staked`
 - `cw20-stake`
 
-Creating a distribution, or changing an existing distribution's `hook_caller`,
-requires the caller to expose either `GetHooks {}` or `Hooks {}` and list this
-rewards distributor. Custom callers without either list query are rejected.
-Order setup atomically: `AddHook` then `Create`; or `AddHook(new)`, `Update`,
-then `RemoveHook(old)`. A runtime failure does not cause a fungible staking
-producer to remove the registered receiver.
+This contract rejects hook calls from any address that is not the
+`hook_caller` of at least one distribution. Staking contracts treat a failed
+hook call as non-fatal: the stake or unstake still succeeds, the hook stays
+registered, and the failure is recorded as attributes on the staking
+transaction (`action: stake_hook_failed`). This means a stale or premature hook
+registration can never block staking, but it also means rewards are not being
+tracked while the hook is misconfigured, so set up hooks and distributions
+together (ideally in the same proposal).
 
 ### Creating a new distribution
 
@@ -156,9 +158,9 @@ You can also update the `vp_contract`, `hook_caller`, and
 `withdraw_destination`.
 
 > **WARNING:** You probably always want to update `vp_contract` and
-> `hook_caller` together. Make sure you know what you're doing. Register the
-> distributor on the new caller before updating, then remove it from the old
-> caller after the update succeeds.
+> `hook_caller` together. Make sure you know what you're doing. And be sure to
+> add/remove hooks on the old and new `hook_caller`s accordingly, ideally in the
+> same proposal as the update.
 
 ### Withdrawing
 
--- a/contracts/distribution/dao-rewards-distributor/README.md
+++ b/contracts/distribution/dao-rewards-distributor/README.md
@@ -36,6 +36,13 @@ voting power changes, such as:
 - `dao-voting-token-staked`
 - `cw20-stake`
 
+Creating a distribution, or changing an existing distribution's `hook_caller`,
+requires the caller to expose either `GetHooks {}` or `Hooks {}` and list this
+rewards distributor. Custom callers without either list query are rejected.
+Order setup atomically: `AddHook` then `Create`; or `AddHook(new)`, `Update`,
+then `RemoveHook(old)`. A runtime failure does not cause a fungible staking
+producer to remove the registered receiver.
+
 ### Creating a new distribution
 
 Only the `owner` can create new distributions.
@@ -149,8 +156,9 @@ You can also update the `vp_contract`, `hook_caller`, and
 `withdraw_destination`.
 
 > **WARNING:** You probably always want to update `vp_contract` and
-> `hook_caller` together. Make sure you know what you're doing. And be sure to
-> add/remove hooks on the old and new `hook_caller`s accordingly.
+> `hook_caller` together. Make sure you know what you're doing. Register the
+> distributor on the new caller before updating, then remove it from the old
+> caller after the update succeeds.
 
 ### Withdrawing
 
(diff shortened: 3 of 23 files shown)


noahsaso:
Thanks for this. I've acted on the third point, and I want to push back on the first with evidence.

## Failure attribution: agreed, and it was worse than described

Fixed in the latest commit. The failure event now names the receiver.

It matters more than the review suggests, because the error string is not the receiver's. wasmd redacts submessage errors before they reach a reply handler:

```go
// x/wasm/keeper/msg_dispatcher.go, redactError
return fmt.Errorf("codespace: %s, code: %d", codespace, code)
```

So on chain the `error` attribute was a bare codespace and code, never "Voting power changed hook sender incorrect". With more than one hook registered a DAO could not tell which receiver failed, or why. My unit tests passed raw strings through cw-multi-test, which does not redact, so they overstated what operators would see. That is fixed too.

`cosmwasm-std` 1.5.5 has no `SubMsg::payload`, so the address has to come from the reply ID. Stake and unstake hooks now occupy two reserved ID ranges, each submessage is tagged with the hook's index in the producer's registry, and `handle_stake_hook_reply` resolves the address from that index. The failure event is now `action: stake_hook_failed`, `hook`, `addr`, `error`. Tests assert that with a failing hook and a healthy hook registered together, the event names the failing one and not the healthy one.

A side benefit: `dao-hooks` owns the ID range, so the four message builders keep their original signatures and the diff against `development` shrank.

## Gas limit: mechanism is right, remedy is wrong

Two premises are correct. `reply_always` does set `gas_limit: None`, and out-of-gas is only catchable inside `dispatchMsgWithGasLimit`, so without a limit it aborts the transaction. But I don't think `with_gas_limit` should be added here.

**It isn't introduced by this PR.** The dispatch it replaces, `SubMsg::new`, also sets `gas_limit: None`. Gas exposure is identical before and after. What changed is that a receiver used to be able to block staking by *returning an error*, which is the audit's Critical and needed no gas at all. That path is now closed.

**It needs a passed proposal.** `AddHook` is gated on the DAO or owner in all four producers. Installing a gas bomb in a DAO's own staking registry takes a governance vote that could just as easily migrate the staking contract or drain the treasury. The audit's Critical needed no malice: a routine owner update of a distribution, or plain message ordering, was enough.

**No safe constant exists, and the audit says so twice.** Finding 30 notes distributor work grows with an owner-set, unbounded number of distributions. Finding 26 notes wormhole rewrites grow with a delegate's pending expiries. Both conclude the threshold "cannot be established from the code alone: it depends on the host chain's gas schedule and limits, which vary between the chains DAO DAO is deployed on." A gas limit is exactly that number.

**A limit set too low is worse than the bug.** Under the fail-open policy a legitimate hook that exceeds it fails silently: staking succeeds, reward accrual stops, delegated voting power stops updating. That is the "silently discards the receiver" class the audit condemns in Findings 4 and 6, moved onto the funds path.

**And it would not close the hole.** wasmd applies the limit conditionally:

```go
limitGas := msg.GasLimit != nil && (*msg.GasLimit < gasRemaining)
```

Once remaining gas falls below the limit it is skipped entirely and out-of-gas propagates uncaught. Later hooks in a registry, and any transaction near its gas budget, are exactly where a bomb would sit. On out-of-gas the parent is charged the full limit anyway, so the griefing cost survives.

**The suggested test cannot be written.** cw-multi-test 0.20.1 has no gas metering; it only forwards the `gas_limit` field. A real gas-exhaustion test needs test-tube against a chain binary, and that workflow is currently red on `development` for unrelated reasons.

The residual is real but it is a bounded-work problem, and the audit's own remedy for it is Finding 30's: bound the fan-out. I'd rather do that than pick a magic number. Happy to open a follow-up.

## hook_caller validation

Deliberately omitted, and called out in the PR description. The UI builds distribution proposals as Create, then FundLatest, then AddHook, so validating at Create rejects every proposal it produces today. Now that hook failures are non-fatal it is hardening rather than a security fix, so it should land with the UI ordering change rather than ahead of it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YBSNZfoCq2XqymbVV9r5rE

PR #935, opened 2026-09-21
fix(ci): restore compatible optimizer and test fixtures

tauagent:
## Summary

I restored the optimizer/test fixtures and replaced the obsolete native integration client, without changing contract runtime code, Test Tube versions, gas limits, or the original application assertions:

- Roll workspace/release optimizer images back to **0.16.1**, preserving sign-extension lowering.
- Pin **CW721 0.18.0** and enforce its SHA256; select the intended default/Osmosis token-staked test artifact.
- Pin the integration fixture to **Juno 19.0.0 / wasmvm 1.5.2**, with the corrected gRPC binding; retain exact CI Wasm/checksums for three days.
- Replace cosm-orc/cosm-tome in **both integration-tests and bootstrap-env** with a native-only facade using **cosmrs 0.22.0 + tendermint-rpc 0.40.4**, explicitly using CometBFT 0.37.
- Submit once, require matching hash, positive committed height and both CheckTx/DeliverTx success, and preserve committed data/events/gas. Disable redirects, bound the whole transaction operation to 30 seconds, and cache failed setup to prevent implicit resubmission.
- Preserve configuration/deployment maps and stdout variables; require an explicit RPC endpoint for previously grpc-only custom configurations. See the [native-client compatibility and maintenance notes](https://github.com/tauagent/dao-contracts/blob/e0df5ef1042995f0274292d462599507d752dcb7/ci/chain-client/README.md).

**This now changes Cargo manifests/lockfile and native harness code.** There is no vendored client. Native dependency additions/removals include a required **cc 1.1.6 → 1.2.8** update for patched ring 0.17.14; this is not a zero-dependency-change claim. Compiler/workflow versions are unchanged. The contract Wasm normal/build/feature trees remain identical across all 36 contract packages; final optimized-byte identity still needs CI verification.

## Why

Optimizer 0.17.0 removed `wasm-opt --signext-lowering` and bumped Rust to 1.86. Its [changelog](https://github.com/CosmWasm/optimizer/blob/v0.17.0/CHANGELOG.md#0170---2025-06-26) warns that output requires CosmWasm 3.0+. The rollback builds successfully under Rust 1.81, but exposed external CW721 drift and RBAM/filter float instructions that the old integration VM rejects.

The full float-bearing artifact set is **not claimed compatible with wasmvm 1.1**. [CosmWasm 1.5 introduced float support](https://github.com/CosmWasm/cosmwasm/blob/v1.5.0/CHANGELOG.md#150---2023-10-31). The pinned historical node is a fixture, not production upgrade advice, and does **not** fix the existing `serde-json-wasm` floating-JSON serialization panic. [Initial investigation](https://github.com/DA0-DA0/dao-contracts/pull/935#issuecomment-5754656661) and retained CI analysis document that limitation.

After the fixture corrections, SDK 0.47 rejected the old client's removed blocking gRPC broadcast mode. SYNC alone would only prove CheckTx admission. The maintained RPC transport supports the 0.37 dialect; the facade checks actual committed execution instead. It decodes the RPC execution-data base64 separately from already-plain-string event attributes and requires canonical uppercase transaction hashes.

## Current-head verification

Head: **`e0df5ef1042995f0274292d462599507d752dcb7`**.

All three new workflows currently require maintainer approval to run:

- [Basic](https://github.com/DA0-DA0/dao-contracts/actions/runs/35689458203)
- [Integration Tests](https://github.com/DA0-DA0/dao-contracts/actions/runs/35689458173)
- [Test Tube Tests](https://github.com/DA0-DA0/dao-contracts/actions/runs/35689458202)

**Final-head acceptance is not yet met.** Full Integration/Basic/Test Tube success, all 43 stored artifacts, the 40 workspace checksum comparisons, and execution of the bootstrap compatibility test remain required.

Completed checks:

- Existing `nightly-2024-01-08` / Rust 1.77: both native consumers pass test-target compiler checks and scoped clippy with warnings denied; repository rustfmt and whitespace checks pass.
- **16 native-client tests pass**, covering fixture mnemonic/address derivation and signature verification, unchanged simulation/signing/fee envelopes, malformed/rejected responses, deadlines, one-submission behavior, redirects/disconnects, event/data handling, registry/gas compatibility and cached setup failure. One opt-in chain pilot is intentionally ignored during ordinary unit runs and was run separately.
- **Pinned-node pilot passed in 11.55 seconds** using retained, checksum-verified CI CW4 bytes: real store/instantiate/execute/query, all five account lookups, bank/bonded-validator queries, block-height/time polling, simulation rejection and a test-only injected committed rejection (**CheckTx 0, DeliverTx 5, positive height**). The disposable node/volume were removed. This pilot is **not** full-suite or all-artifact storage proof.
- The original eight integration tests, their assertions and inputs are preserved; only client imports/access and setup handling change. All 36 contract-target normal/build/feature trees match `9df5cd8e`; no contract/package runtime source changed in the native migration.
- The native-only crate compiles as an empty wasm32 library and adds no contract `cdylib`. Dependency versions are constrained to the existing compiler; upstream license metadata and lockfile checksums were reviewed.
- Bootstrap stdout/config compatibility coverage was added and typechecked; execution awaits Basic CI. Local combined native-test builds ended without confirmed completion while compiling `osmosis-std`, so I do not count them as passing and did not continue heavy local retries.

Earlier evidence remains distinct:

- At `9df5cd8e`, [Basic passed](https://github.com/DA0-DA0/dao-contracts/actions/runs/35645014208) and [full Test Tube passed](https://github.com/DA0-DA0/dao-contracts/actions/runs/35645014246); [Integration failed](https://github.com/DA0-DA0/dao-contracts/actions/runs/35645014260) on the removed broadcast mode. Those are **not** results for this new head.
- All 43 retained CI checksums were verified; RBAM panic metadata and filter float paths were examined. The earlier checksum corruption-rejection test and eight actual-loader filename-selection tests passed. They verify downloads/file selection, not full-chain execution.
- Full optimizer build evidence comes from CI. Earlier local optimizer attempts remain incomplete; no MSRV fallback or new broad local optimizer rebuild is claimed.

## Follow-up

After this lands, [PR #932](https://github.com/DA0-DA0/dao-contracts/pull/932), [PR #933](https://github.com/DA0-DA0/dao-contracts/pull/933), and [PR #934](https://github.com/DA0-DA0/dao-contracts/pull/934) need fresh merge-ref CI/update/rebase to pick up the fix.

Broader runtime modernization, repository-wide Codecov failures, deprecated workflow actions, and optional optimizer digest pinning remain separate work. Review and merge are not implied by the local checks or pilot.


noahsaso (approved):

tauagent:
### Investigation update — proposed fixes, not implemented

I found a concrete CW721 pin and traced a float-producing RBAM serialization path. The PR remains unchanged at `c77fc0f7`; both integration workflows are still red.

**CW721 candidate:** use the [0.18.0 release artifact](https://github.com/public-awesome/cw-nfts/releases/download/v0.18.0/cw721_base.wasm), matching the existing Rust 0.18 API, with a hard-coded SHA-256 check:

`ba81e10d053814f1dfbb92f20f77ce1cccf64b27b639db9e2afa8ab5d6ea3cf7`

I matched this to the publisher's release checksum and decoded no sign-extension, bulk-memory copy/fill, or float instructions. The instantiate/minter/mint/send interfaces match the test source. This is supporting static evidence, not a claimed chain-test pass. The release tag is unsigned; the reviewed hash supplies immutability, not independent publisher authentication.

**RBAM attribution:** the failing CI run did not retain its Wasm artifact, so exact binary attribution remains limited. However, the published v2.8.0-alpha.2 RBAM binary uses the same RBAM/filter/protobuf sources and lockfile. Its sole float instruction is a discarded `f64.load offset=16` immediately before a panic. Decoding the panic metadata identifies `serde-json-wasm-0.5.2/src/ser/mod.rs:238:9`: `serialize_f64`'s `unreachable!()`. Static call paths reach this from instantiate, execute and query. This corroborates the source path through serialization of RBAM's `serde_json::Value` filters; it is stronger than a generic dependency suspicion, but the published binary has a different hash from CI.

An RBAM-only opcode workaround would not establish compatibility: the same-source published `cw-filter` binary also contains real float comparisons, conversions and arithmetic, consistent with its intentional mixed-number comparison code.

**Recommendation for approval:** pin/checksum CW721 0.18, keep optimizer 0.16.1, and—if the supported-chain policy permits it—move the integration Juno fixture to `v19.0.0`, pinned at manifest digest `sha256:1ec8ecb4b2f5aa967de526526edcefeb8cc27bd77e292f7d4c36daf474e9ad5e`. I verified this image reports wasmvm 1.5.2 / wasmd 0.45.0 and contains the existing bootstrap interface; no node/tests were run on it. [CosmWasm 1.5 explicitly introduced float support with NaN canonicalization](https://github.com/CosmWasm/cosmwasm/blob/v1.5.0/CHANGELOG.md#150---2023-10-31).

This would preserve contract code but change the tested runtime and compatibility guarantee: float-bearing RBAM/filter functionality requires a float-capable VM. If supporting every current artifact on the old float-rejecting VM is mandatory, a larger no-float JSON/filter/protobuf redesign is needed instead. Updating the VM does not fix the separate inability of `serde-json-wasm` to serialize actual floating JSON values.

I also recommend retaining Wasm/checksums from the next CI run to close the exact-artifact evidence gap. Any approved implementation must pass both full integration workflows and Basic checks, preserving gas limits and assertions. No contract, download, runtime, or workflow changes beyond the existing rollback have been made for this investigation.

tauagent:
### Current-head results and the remaining setup defects

I checked all runs against `7c63bc840d32161ddef72bc53ce0e198ba29a1b5`:

| Workflow | Result |
|---|---|
| [Basic](https://github.com/DA0-DA0/dao-contracts/actions/runs/35592004720) | Test Suite and Lints passed |
| [Integration Tests](https://github.com/DA0-DA0/dao-contracts/actions/runs/35592004679) | All eight cases time out waiting for the first block, **before artifact storage** |
| [Test Tube Tests](https://github.com/DA0-DA0/dao-contracts/actions/runs/35592004695) | CW721 Wasm tests now pass; eleven later token-staked cases cannot read the expected artifact filename |

I have not marked the fix verified: both full integration workflows must still pass without changing gas limits or application assertions.

#### Exact CI Wasm attribution

The retained `integration-contracts-35592004679-2` artifact contains 43 checksummed Wasm files; **all 43 checksums verify**. Decoding these actual CI binaries closes the earlier release-binary evidence gap:

- RBAM SHA256 `429ac0538470e6691162bdb99b5f1f3cd7bc15dc983bebe5a24ef52f4892299d` (also the earlier rollback-only CI checksum): exactly one `f64.load offset=16`, discarded before a panic. The actual panic metadata resolves to **`serde-json-wasm-0.5.2/src/ser/mod.rs:238:9`**, `serialize_f64`'s `unreachable!()`. Static direct-call paths reach this function from instantiate, execute, and query; that is potential reachability, not a dynamic transaction trace.
- Filter SHA256 `8b3d06127c8d5d0ed9707324bb3ddf5e466eb664074ba0eb7a65f54cce6c6b6f`: 114 float-related instructions, including genuine comparisons/conversions/arithmetic reachable from query, consistent with the intentional mixed-number comparison fallback. This is not just RBAM's discarded load.
- The CI CW721 bytes match the pinned 0.18.0 SHA256. None of these three decoded binaries contains sign-extension instructions.

These results support the documented float-runtime floor and unchanged floating-JSON serialization limitation. They do **not** prove complete artifact storage; this integration run stopped earlier.

#### Smallest corrective changes I propose

1. **One Docker environment argument:** `JUNOD_GRPC_ADDRESS=0.0.0.0:9090` in `deploy-local`. Juno19's generated gRPC config defaults to container-local `localhost:9090`, and the existing image scripts do not change it. In two isolated, bounded probes using the exact pinned image and existing bootstrap, both nodes produced blocks, but the published gRPC port returned HTTP/2 SETTINGS only with this environment override. This is a connectivity smoke—not a complete integration/storage pass. The [SDK's environment-config handling](https://github.com/cosmos/cosmos-sdk/blob/v0.47.6/server/util.go) supports the override without replacing bootstrap scripts.
2. **Only test-fixture filename selection:** the token-staked test loader expects `dao_voting_token_staked.wasm`, but its unchanged optimizer metadata builds `dao_voting_token_staked-default.wasm` and `dao_voting_token_staked-thorchain.wasm`. The suite uses `OsmosisTestApp`, so it must explicitly load the **default / osmosis_tokenfactory** variant. I propose fixing the primary filename and retaining the same explicitly default-prefixed architecture-suffix fallback pattern. Current optimizer0.16.1 normally preserves the variant basename on both architectures; this fallback must not select Thorchain or stale unsuffixed bytes.

I have not implemented these further changes. Optimizer/runtime versions, contract behavior, gas limits, and application assertions remain unchanged. The separate pre-existing malformed ARM shell conditional is not included in this proposal.

tauagent:
### Two workflows green; remaining SDK broadcast incompatibility

On `9df5cd8e702f2362265bec0a100666dba7ba87b3`, [Basic](https://github.com/DA0-DA0/dao-contracts/actions/runs/35645014208) and [the full Test Tube workflow](https://github.com/DA0-DA0/dao-contracts/actions/runs/35645014246) passed.

[Integration Tests](https://github.com/DA0-DA0/dao-contracts/actions/runs/35645014260) now passes first-block polling and attempts its first store transaction, but fails with `unsupported return type unspecified; supported types: sync, async`. I traced this exactly: [cosm-tome 0.2.2](https://github.com/de-husk/cosm-tome/blob/v0.2.2/src/clients/cosmos_grpc.rs) sends the removed `Block` broadcast enum; [SDK 0.47.6](https://github.com/cosmos/cosmos-sdk/blob/v0.47.6/client/broadcast.go) normalizes unsupported enums to `unspecified` and raises that error. This is not complete storage or VM-acceptance proof.

The newly retained artifacts have all 43 checksums verified and an identical manifest to the binaries already decoded, so the exact RBAM/filter attribution remains applicable.

I propose a narrowly modified, provenance-verified local copy of the existing native client: broadcast once with **Sync**, then poll **GetTx** for committed execution under one bounded deadline. Merely switching to Sync would be wrong: CheckTx admission is not execution success. The fix must verify transaction identity and committed height, propagate both CheckTx and execution failures, preserve committed events/gas, retry only NotFound, time out safely, and never automatically rebroadcast. Existing nonblocking APIs would remain unchanged.

There is no newer published drop-in release of these clients. Vendoring adds a maintenance/review burden, despite the narrow behavior change. I have not implemented that further change; current source, dependency versions, gas limits and application assertions remain unchanged. Both-green acceptance is still unmet.

